PR #1037, opened 2010-12-16
CPColorList

stevegeek:
Hey guys, here is an implementation of NSColorList for cappuccino (sans file write/delete). 

Demo here: http://www.stephenierodiaconou.com/colorlistdemo/
(based on an example in Cocoa Programming for Mac OS X , Hillegass)

I wanted to get some feedback on what should be the format of the ColorLists themselves. In Cocoa ColorLists are stored as bundles in a number of possible formats, so far i have seen either a simple text format or as binary archived objects. Cocoa color lists can be localised and can represent colors in numerous color spaces from what I can tell. Cocoa determines available color lists from the system and user by searching in certain folders (eg /System/Library/Colors/).

For Capp I decided to implement a slightly different plist based format for the lists that encapsulated the colorspace, colors and localised strings, (the localised strings at the moment dont do much as there obv isnt such support in Capp anyway). 

The available lists are noted in either the <appkit bundle>/Colors/lists.plist for framework lists or <app bundle>Colors/lists.plist for app lists. Lists can be either included right in this plist or can simply reference another file to be loaded lazily as required. For each list a dictionary contains the keys "Mode", "Names", "Keys" and "Colors" where Mode is a string representing the color space, "Names" is a dictionary containing the localised key/names pairs, "Keys" is an ordered list of the color keys, and "Colors" is a dictionary of key/array pairs where the array contains the color components for the particular color (as many as needed depending on the Mode).  

1) What do people think about lists.plist file? 

2) At the moment im using XML plist format but I suppose it should detect if the plist is XML or 280N? Does cappuccino have a way to do this automatically?

3) Archived versions of the lists could be created in advanced as 280N Plists and used, as in Cocoa, but that would need some simple modification of the plist parsing code to detect whether the loaded file is an archived object or a plist in the format described above, and then unarchive the object as appropriate. 

4) The localised strings at the moment dont matter too much, but in general for the future I suppose including them in the plist is not a good idea as the plist size will grow unnecessarily , hence I think that for each language code in "Names" maybe they should also support referencing another file to be loaded if needed. 

5) At the moment I have implemented only 1 color list, the "Apple" colour list. I will implement the rest when the above has been sorted.

Feedback/comments/etc please


daboe01:
@aparajita should we close this one now? IMHO it is not going anywhere.

aparajita:
@daboe01 Don't close it, maybe it will be useful someday.

daboe01:
this one is jamming our inbox.
i vote for closing as nobody seems interested after 7 years waiting.
any objections?

daboe01:
+#wont-fix

PR #1697, opened 2012-11-30
Added support for Full Keyboard Access for checkboxes, radios and popups

mrcarlberg:
Full keyboard access can be turned on by setting the key CPFullKeyboardAccess to YES
in the Info.plist file or set it programmatic on CPApplication.
A theme state "focused" is added to allow controls to show when it is in focus.
A CPMatrix embryo is added to support full keyboard access for radios.
It is only supporting radio buttons and is very simple.
TODO: There are controls that don't support keyboard access.
TODO: Aristo theme need graphics to show the new "focused" theme state.

I have written a test application using a special theme for demonstration: http://mini.carlberg.org/dev/FullKeyboardAccess/index-debug.html

### DIFF
--- a/AppKit/AppKit.j
+++ b/AppKit/AppKit.j
@@ -68,6 +68,7 @@
 @import "CPImageView.j"
 @import "CPKeyBinding.j"
 @import "CPLevelIndicator.j"
+@import "CPMatrix.j"
 @import "CPMenu.j"
 @import "CPMenuItem.j"
 @import "CPOpenPanel.j"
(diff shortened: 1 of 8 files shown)


daboe01:
@mrcarlberg times have changed. i would love to have CPFullKeyboardAccess. could we please revisit this cool stuff?

mrcarlberg:
I think it is up to the community to decide if Full Keyboard Access should be merged or not. The current Radio Button implementation in Cappuccino removes the NSMatrix view and replaces it with a RadioGroup object that is not a view. My implementation will keep the NSMatrix view and still create the RadioGroup object. The NSMatrix view handles full keyboard access and the RadioGroup still handles that only one radio button is selected. You can make the case that this is not optimal as all this can be handled in one place along with all the stuff discussed above. I feel that it is better to have the functionality than to wait for the perfect implementation. And it has worked for almost 6 years without any major changes.

theredhead:
+1: Jeff Raskin demands this

PR #1892, opened 2013-04-02
CPTableView formatting and fixes [+1]

cacaodev:
There might be a "breaking change" here, in commit  be7f6ae. I didn't mention it in the commit message because i wanted your input first.
Previously, -reloadData was reloading everything: the views, the data and performing various tasks needed to setup the view. Now it just fetches the data from the model and applies it to the view. Except when the number of rows changed: in this case we reload the views like we did before meaning that when calling -reloadData from the controller after a row insertion/removal, there should be no change. To be sure nothing breaks, i would like to know how you use -reloadData in your controllers and if you need the views to be reloaded even when the number of rows in the content did not change.

There is one situation I can think of where reloading the views might be useful but i don't think anyone ever tested it. It's when using view-based tables with the delegate for creating views. You may want to initiate a remote request to fetch some data (lazy loading) and then create a view in the delegate based on the content. In this case, you need to tell the table to reload one or more specific views when the data is received.
If we do this, we will need a new implementation, not the previous reloadData (renamed _reloadDataViews here) . This method was not fetching the views from the delegate, instead, we were just enqueuing the views in the cache and redistributing them with new values and new frames, which is not enough.


primalmotion:
@cacaodev can we get rid of this one once for all :)

Can you make it mergeable and then tell me? I'll merge it right away

Thanks

+#needs-improvement
-#ready-to-commit

daboe01:
hi @cacaodev, i need this one badly ;-)
+1

cacaodev:
@primalmotion  @daboe01 : hey guys, sorry for the delay but I don't have much time currently to work on this. I'll have a look when I can.
If someone want to finish the work, you can merge it, fix the conflicts, and then fix issues with highlighting. From what i remember, in some situations the highlighted state was not correctly set on the main cell view or subviews.

cacaodev:
@Dogild : can you re-apply e56cc85e963a (i'm not sure what to check) and then i think that's it.

Dogild:
Let me try that asap ;)

primalmotion:
@cacaodev awesome

@Dogild I have no doubt you'll tell me when you 're done :)

Dogild:
@primalmotion I'm kinda in a middle of a huge big thing ... :p

Dogild:
@primalmotion @cacaodev Archipel and Nuage works and not crash with this PR. 
@cacaodev #e56cc85 is still good in the PR

primalmotion:
@cacaodev can you sum up what this PR actually does? Thanks!

-- 
Antoine Mercadal

cacaodev:
@primalmotion New features are: new methods -enumerateAvailableViewsUsingblock: and -preparedViewAtColumn:row:
Then there are a lot of bug fixes related to column d&d, minor fixes for content bindings setDataView:
Internally, the code is rearranged and simplified thanks to the new view enumeration order (row -> column) and editing for both view-based and cell-based is now unified.

primalmotion:
finally merged!

Thanks!

+#fixed

cacaodev:
There is a style issue with this commit. A lot of brackets are not correctly aligned. It may be because of a setting in my merging tool.
Sorry for that. I believe the only solution is to manually fix this (capp_lint does not detect that). I'll do it.

PR #1951, opened 2013-06-24
When determining whether we're a local file request, see if we have a valid host before using the window's location.

GlennAustin:
Before checking whether the request is local, we should first check to see if the original request had a host before assuming that we need to use the window's host.


GlennAustin:
Old, things are probably a lot different now so I'm just going to close this.

Issue #1432, opened 2012-01-15
CPTextView class missing


daboe01:
Hi,
Will this be fixed?
Would you accept a port of the respective cocotron obj-c classes (MIT-license)?
Best greetings,
Daniel

daboe01:
updated the online demo so you guys can see the new regex based granularity engine. make sure to clear the browser cache

aparajita:
Absolutely awesome! Two issues I see:

- When the Font panel is opened, the color is not set to the color of the text if all of the selected text is the same color.

- When clicking on a letter, it always places the caret to the left of the letter. The caret should be placed to the right of the letter if the click point is >50% width of the letter.

daboe01:
thank you @aparajita.
i fixed both issues and updated the demo.
best greetings,
daniel

aparajita:
How are you calculating the character width? It seems the calculation is biased towards the left by a certain percentage, because at smaller font sizes I have to click farther to the right to get the caret to appear to the right.

daboe01:
i do not see it (testing in chrome)
the code is
```objj
 partialFraction[0] = (point.x - frames[j].origin.x) / frames[j].size.width;
```
it has to be a rounding issue.

aparajita:
I checked again zooming in on the screen, and it is working correctly. Great job!

robert-plasq:
What's the current status of this? Will it get merged in to the main Cappuccino code base soon?

daboe01:
hi all,
implementing the native deadkeys caused a rat-tail of issues. e.g. this broke system-wide copy-paste and all menu keyboard shortcuts.
by now i fixed everything that i spotted.
i am sure i missed some issues, though.
i.e. i did not test on Safari and IE.
could you please help me with testing?
i updated the online demo.
many thanks and best greetings,
daniel

mrcarlberg:
Awesome!

I did a quick test in Safari and I found a couple of things:
1. Double click on a word and do 'Copy' from the 'Edit' meny (or CMD-c). When I select another place with the mouse and do paste I get an extra space at the end of the word. I do not get it if I paste it into an external editor. (Sometimes I can get it to work but I can't figure out why)

2. If I hit 'Arrow-down' key somewhere on the last line the cursor does not go to the end of the file.

3. If I type some text (easier to see if it is many lines) and press 'Shift-Arrow-Up'. The selection is now from the last clicked cursor position to the current one. Not from where I typed to the current cursor position.

[...]

Issue #1400, opened 2011-11-17
CPTextField sizeToFit may clip text


walisser:
You can see in the screenshot the right edge is clipped by one pixel

http://img24.imageshack.us/img24/1391/psimagef.jpg

``` objective-j
var slash1 = [[CPTextField alloc] initWithFrame:CGRectMake(23, 7, 15, 18)];
 [slash1 setStringValue:@"/"];
 [slash1 sizeToFit];
 [slash1 setBackgroundColor:[CPColor greenColor]];
```

Cappuccino 0.9.5
Chrome 16.0, Ubuntu 11.10

And here I've expanded the frame (after sizeToFit) by one pixel
http://img20.imageshack.us/img20/4802/psimageq.jpg

PR #2068, opened 2014-02-10
New: CPTextView, CPTextStorage, CPTextContainer, CPLayoutManager and CPSimpleTypesetter + CPFontPanel + RTF parser/producer [+2]

daboe01:
We are highly excited to announce that CPTextView and all the other classes of the Cocoa text system have today been merged into Cappuccino. This work is based on the orphaned repository from http://github.com/emaillard/cappuccino. After three years of hard work and more than additional 600 commits, we proudly support a strong set of features:

* High quality and fast text rendering with ligatures and kerning
* Highly optimized to support large documents (tested with 2000 lines and more)
* Consistent cross-browser / cross platform look and feel that is nearly feature par with the Mac
* Font panel
* Unlimited undo / redo
* Native copy and paste
* Support for rich text
* RTF Parser and RTF Writer
* Support for input of diacritics (aka 'dead keys' or native input manager)
* Support for Xcode (nib2cib)
* Tested on Chrome, Safari and Firefox

CPTextView can be used for a simple plain multiline text input  and also to for a full blown online word processor application.

If you miss a feature or discover a bug, please open an issue on https://github.com/cappuccino/cappuccino.

Best whishes,

Daniel and the core committers team

---

Cappuccino still needs a patch to make it work fully:

* Merge in PR #2462 (rich text copy/paste)
* Fixes issues: #1432 #2357


Dogild:
@daboe01 did you have a chance to work on nib2cib or not ? I'll try to take some times (next week) to help you on this one !

daboe01:
@Dogild i have never used nib2cib in my projects. i am a happy user of cappusance :-)
IMHO it should be sufficient to only extract frame and bordertype in the beginning. Future users can improve on this basis as needed.

aparajita:
@daboe01 If we're going to add nib2cib support, all attributes supported in Cappuccino have to be supported in nib2cib. I'm sure @Dogild can handle it.

aparajita:
@daboe01 I just looked at the source, and you should be moving a bunch of ivars and methods to CPText. Take a look at the Cocoa docs, a lot of functionality in CPTextView inherits from CPText.

daboe01:
The apple docs say that you should gernerally use NSTextView instead of NSText
@aparajita do you expect instances of CPText to actually work for the end-user?
 This would require a lot of new code and result in broad code duplication from CPLayoutManager and friends.

Dogild:
@daboe01 @aparajita is right, you can take a look to NSText.h and NSTextView.h to see where ivars are declared.

daboe01:
@aparajita @Dogild did i get it right with f39a1e7?

Dogild:
Yep it's good like this ! I have started to take a look in the class, I'll try to regularly tell you what we need to improve :) (I'll work on the nib2cib and unittest in the same time)

AppKit/CPTextView/CPTextView.j
```diff
@@ -0,0 +1,1890 @@
...
+    @ignore
+*/
+- (void)resignKeyWindow
+{
+    [self setNeedsDisplay:YES];
+}
+
+- (BOOL)_isFocused
+{
+   return [[self window] isKeyWindow] && _isFirstResponder;
```
Dogild: You should replace _isFirstResponder by `[[self window] firstResponder] === self ` and then remove this ivars

AppKit/CPTextView/CPTextView.j
```diff
@@ -0,0 +1,1890 @@
...
+
+#pragma mark -
+#pragma mark Responders method
+
+- (BOOL)acceptsFirstResponder
+{
+    if ([self isSelectable])
+        return YES;
+
+    return NO;
```
Dogild: Just return [self isSelectable] right now.

I'm surprised that this method does not have the same behavior as the one of the CPTextField. You you check

AppKit/CPTextView/CPTextView.j
```diff
@@ -0,0 +1,1890 @@
...
+    if (_delegate)
+    {
+        if ([_delegate respondsToSelector:@selector(textDidChange:)])
+            [notificationCenter addObserver:_delegate selector:@selector(textDidChange:) name:CPTextDidChangeNotification object:self];
+
+        if ([_delegate respondsToSelector:@selector(textViewDidChangeSelection:)])
+            [notificationCenter addObserver:_delegate selector:@selector(textViewDidChangeSelection:) name:CPTextViewDidChangeSelectionNotification object:self];
+
+        if ([_delegate respondsToSelector:@selector(textViewDidChangeTypingAttributes:)])
+            [notificationCenter addObserver:_delegate selector:@selector(textViewDidChangeTypingAttributes:) name:CPTextViewDidChangeTypingAttributesNotification object:self];
```
Dogild: This is going to add memory leaks in the application. You should take a look to the commit 392a6cfb to see how to deal with this kind of delegate

AppKit/CPTextView/CPTextView.j
```diff
@@ -0,0 +1,1890 @@
...
+        _caretDOM = document.createElement("span");
+        style = _caretDOM.style;
+        style.position = "absolute";
+        style.visibility = "visible";
+        style.padding = "0px";
+        style.margin = "0px";
+        style.whiteSpace = "pre";
+        style.backgroundColor = "black";
+        _caretDOM.style.width = "1px";
+        _DOMElement.appendChild(_caretDOM);
```
Dogild: In my opinion we should create a private class _CPCaret. So we can separate the creation of the caret and the textView.

AppKit/CPTextView/CPTextView.j
```diff
@@ -0,0 +1,1890 @@
...
+    ret.style.whiteSpace = "pre";
+    ret.style.backgroundColor = [aColor cssString];
+
+    ret.style.width = (aRect.size.width)+"px";
+    ret.style.left = (aRect.origin.x) + "px";
+    ret.style.top = (aRect.origin.y) + "px";
+    ret.style.height = (aRect.size.height) + "px";
+    ret.style.zIndex = -1000;
+    ret.oncontextmenu = ret.onmousedown = ret.onselectstart = function () { return false; };
+    return ret;
```
Dogild: Same remark here

AppKit/CPTextView/CPTextView.j
```diff
@@ -0,0 +1,1890 @@
...
+
+    if ([self shouldDrawInsertionPoint])
+    {
+        [self updateInsertionPointStateAndRestartTimer:NO];
+        [self drawInsertionPointInRect:_caretRect color:_insertionPointColor turnedOn:_drawCaret];
+    }
+    else // <!> FIXME: breaks DOM abstraction, but i did get it working otherwise
+    {
+        if (_caretDOM)
+            _caretDOM.style.visibility = "hidden";
```
Dogild: With a private class we could avoid to have this kind of line of code, but something more objective-j

AppKit/CPTextView/CPTextView.j
daboe01: that is probably because a textview has no enabled property

AppKit/CPTextView/CPTextView.j
daboe01: thank you, fixed.

AppKit/CPTextView/CPTextView.j
daboe01: done

AppKit/CPTextView/CPTextView.j
daboe01: done, but untested

aparajita:
> The apple docs say that you should gernerally use NSTextView instead of NSText
@aparajita do you expect instances of CPText to actually work for the end-user?
This would require a lot of new code and result in broad code duplication from CPLayoutManager and friends.

CPTextView is a subclass of CPText, thus a CPTextView **is** a CPText. So any other classes that reference CPTextView don't know or care that the methods have been moved to CPText.

daboe01:
@Dogild i tried writing unit tests myself (c3e6f4f). However, these need the DOM to make any sense. Is this doable with travis?

Dogild:
Not yet

Dogild:
@daboe01 I have added nib2cib for the CPTextView. (just some options). Can you tell me which options are supported by nib2cib please ?

The unit tests are currently failing because the version of ojtest is not up-to-date on travis. @aljungberg anyway to take the last version of ojtest for the tests (using the new version of OJMoq)?

daboe01:
do you mean which options are supported by the current CPTextView implementation?
it looks like you cover it completely.
we currently do not render most of the paragraph style properties.
best greetings,
daniel

Dogild:
Ok perfect !

AppKit/CPTextView/CPTypesetter.j
```diff
@@ -0,0 +1,467 @@
...
+
+var _measuringContext,
+    _measuringContextFont,
+    _isCanvasSizingInvalid,
+    _didTestCanvasSizingValid,
+    _sharedSimpleTypesetter,
+    _sizingCache;
+
+function _widthOfStringForFont(aString, aFont)
+{
```
Dogild: @daboe01 what os the diff with sizeWithFont: ? It seems this method works with a cache as well ?

daboe01: @Dogild sizeWithFont: turned out to slow for CPTextView. The function _widthOfStringForFont is is ~10x faster on Chrome and FF (where it can be canvas-based). In Safari, it falls back to sizeWithFont: since canvas text sizing is broken there. Please also see my comment in issue #1400

Dogild: Whats are the drawbacks of this method ? It looks like we should the implementation of sizeWithFont: with this method.

daboe01: yes, absolutely.

Dogild: Do you take care of it ?

daboe01: i am not aware of a canvas-based measuring of text height.
does the height return of sizeWithFont: depend on the string?
or is the height solely defined by the font?

aparajita: It's solely defined by the font.

See [here](http://stackoverflow.com/questions/1134586/how-can-you-find-the-height-of-text-on-an-html-canvas) for a different method measuring text height that may be more efficient than the technique we are using now.

daboe01:
@Dogild i did the refactoring of font sizing as you suggested. untested for the moment.

AppKit/CPStringDrawing.j
```diff
@@ -53,20 +57,50 @@ CPStringSizeCachingEnabled = YES;
...
+        CPStringSizeMeasuringContext = CGBitmapGraphicsContextCreate();
+
+    if (!CPStringSizeDidTestCanvasSizingValid && CPFeatureIsCompatible(CPHTMLCanvasFeature))
     {
+        var teststring = "0123456879abcdefghiklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ,.-()";
+        CPStringSizeDidTestCanvasSizingValid = YES;
+        CPStringSizeMeasuringContext.font = cssString;
+        CPStringSizeIsCanvasSizingInvalid = ABS([teststring sizeWithFont:aFont].width - CPStringSizeMeasuringContext.measureText(teststring).width) > 2;
+    }
+
```
Dogild: We should put all of these test/setup things in the `+initialize` method of this object

daboe01: im not sure whether this is a good idea, since we are in a category to CPString. If we implement +initialize here, we might introduce a conflict to a +initialize in the CPString main file (even if this initializer it does not exist yet, it may be introduced in the future and then break the string sizing stuff).

daboe01:
any ideas how to support entering diacritics?
should we try to emulate it in code?
or should we use a (transient) native text field?

aparajita:
How is text entry happening now? Diacritics (or any world character set for that matter) should be done in the normal way for the host OS. You may need a transient input element for that.

daboe01:
i use CPResponder's interpretKeyEvents:

aparajita:
What happens with diacritics now?

daboe01:
the trigger keys (e.g. ^ and ~) are dead.

AppKit/CPStringDrawing.j
Dogild: Yep, I didn't pay attention that we are in a Category. We should at least put that in an private init method.

aparajita:
Take a look at google docs. They use a hidden iframe with a contentEditable body and an event listener on keyup/keydown that goes to that iframe. I think this is how they capture key events and allow native OS input. You can't rely on the Cappuccino key events, that won't work across all input keyboards and languages.

daboe01:
in my testing, google docs does not seem to use the native diacritis mechanism: native: yellow background, google docs: underline (see screenshot mac snowleo+chrome latest). the same is true for github (try it while typing here :-)

bildschirmfoto 2015-05-21 um 10 00 31

aparajita:
Actually, that's just how Chrome implements diacriticals in all input and contentEditable elements. Check in Safari, you'll see it uses a yellow background which is not quite the same color as the system standard.

I looked closely at what happens when entering text in google docs. From what I see, there is an empty span whose left edge is aligned with the left edge of the cursor. Probably the span is not contentEditable until it receives a key event, in which case the event handler sets it to be contentEditable and forwards the event. After the keyup event, contentEditable is set false again so there is no cursor.

The proof that contentEditable is being used is that when you type a diacritical mark and wait, the thinner system cursor appears at the right of the span, not the thick google docs cursor.

Does this make sense?

daboe01:
yes, that makes sense. best greetings, daniel

aparajita:
Do you think you can implement that?

daboe01:
yes, i will give it a try.
it might be convenient to use the DOMPasteboardElement for this purpose (_platformWindow._platformPasteboard._DOMPasteboardElement) instead of a contenteditable div.
what do you think?

aparajita:
Whatever works, as long as you have native input.

daboe01:
i was successful in detecting when a 'dead key' is pressed in the keyUp: method.
However, i have not been successful in propagating the dead key DOM event to the native DOM element (i currently use a CPTextField).

this is my experimental code:
```ObjJ
- (void)keyUp:(CPEvent)event
{
    modifierFlags = [event modifierFlags],
    character = [event charactersIgnoringModifiers],
    selectorNames = [CPKeyBinding selectorsForKey:character modifierFlags:modifierFlags];

    if (character.charCodeAt(0) === 0 && !selectorNames) // dead key detected
    {
        _nativeInputField =[[CPTextField alloc] initWithFrame:CGRectMake(0,0,50,27)];
        [self addSubview:_nativeInputField];
        [_nativeInputField setEditable:YES]
        [_window makeFirstResponder:_nativeInputField]
        [[_window platformWindow] _propagateCurrentDOMEvent:YES];  // does not work
    }
    [super keyUp:event];
}
```

any help is appreciated!

aparajita:
1. I think it's a bad idea to use CPTextField for this, it's way too heavy for this and there are too many interactions with the rest of Cappuccino. Best to use a contentEditable span or div directly if possible.
2. Dead keys are natively implemented on key down, not key up.

daboe01:
If i use contentEditable divs or input dead keys and a lot of other keystrokes do not go through. It looks like e.g. Key equivalents are filtered out


Von meinem iPhone gesendet

aparajita:
Okay, in any case #2 was the important thing, you have to trap dead keys on key down, not key up.

daboe01:
In my testing you cannot detect a dead key on keydown. You need keyup for this. 

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

Von meinem iPhone gesendet

aparajita:
Well, I hate to say it, but I think you'll have to implement it the other way around, with a contentEditable and trapping key equivalents. It is impossible to deal with dead keys using the keyboard scan codes from Javascript events. See [this](http://www.openjs.com/scripts/events/keyboard_shortcuts/) for sample code on how to allow native input and pass through key equivalents.

daboe01:
good news!
i found a solution that works on chrome and FF (clumsy due to timer)
cannot test on current safari/IE though, because i am on snowleo
requires to bubble the DOM event in CPTextView's keyDown:
what do you think?

```objj

var _nativeInputField;
var _nativeInputFieldKeyUpCalled;
var _nativeInputFieldActive;

// FIXME: also handle native copy/paste here
@implementation _CPNativeInputManager : CPObject

+ (void)initialize
{
	_nativeInputField = document.createElement("div");
	_nativeInputField.contentEditable=YES
	_nativeInputField.onkeyup = function(e)
	{
		_nativeInputFieldKeyUpCalled = YES;
		var currentFirstResponder = [[CPApp mainWindow] firstResponder]

		if (![currentFirstResponder respondsToSelector:@selector(_activateNativeInputElement:)])
			return;

		if (!_nativeInputFieldActive && e.keyIdentifier === 'Unidentified') // chrome signal for deadkey
		{
            _nativeInputFieldActive = YES;
            [currentFirstResponder _activateNativeInputElement:_nativeInputField];
		} else
		{
             _nativeInputField.innerHTML = ''
		}
	}
	_nativeInputField.onkeydown=function(e)
	{
		_nativeInputFieldKeyUpCalled = NO;
		var currentFirstResponder = [[CPApp mainWindow] firstResponder]

		if (![currentFirstResponder respondsToSelector:@selector(_activateNativeInputElement:)])
			return;

		// on FF (no keyIdentifier) the only way to detect a dead key is the missing keyup event
		setTimeout(function(){
				if (e.keyIdentifier === undefined && !_nativeInputFieldActive && _nativeInputFieldKeyUpCalled == NO && !e.repeat)
				{
                    _nativeInputFieldActive = YES;
                    [currentFirstResponder _activateNativeInputElement:_nativeInputField];
                }
                else if (!_nativeInputFieldActive)
                    [self hideInputElement];
			}, 400);  // CLUMSY!!
    }
    _nativeInputField.style.width="64px"
    _nativeInputField.style.zIndex = 10000;
    _nativeInputField.style.position = "absolute";
    _nativeInputField.style.visibility = "visible";
    _nativeInputField.style.padding = "0px";
    _nativeInputField.style.margin = "0px";
    _nativeInputField.style.whiteSpace = "pre";
    _nativeInputField.style.outline = "0px solid transparent";
}
+ (void)focus
{
    var currentFirstResponder = [[CPApp mainWindow] firstResponder]

    if (![currentFirstResponder respondsToSelector:@selector(_activateNativeInputElement:)])
        return;

    [self hideInputElement];
	currentFirstResponder._DOMElement.appendChild(_nativeInputField);
    _nativeInputField.focus()
}

+ (void)hideInputElement
{
	_nativeInputField.style.top="-1000px"
	_nativeInputField.style.left="-1000px"
}
@end
```

aparajita:
Cool! Can I see it in action in a demo?

daboe01:
i pushed it here: <https://github.com/daboe01/CPTextView>

aparajita:
I'm not getting any dead keys with that demo.

daboe01:
The online Demo is outdated. Use the repo


Von meinem iPhone gesendet

aparajita:
I did use the repo. No dead keys (or any modifiers other than shift) were working.

daboe01:
i think that this is due to the american key layout requiring metakeys to access the deadkeys. i just worked around some chrome inconsistencies. could you please retry? (clear browser cache)
does it work in FF?

aparajita:
I'm very sorry, but your approach won't work. Not everyone (including me) uses the US keyboard layout, for example I use Dvorak, and it isn't acceptable to have this work with a single keyboard layout. I just don't see any way this will work by scanning keyboard codes.

daboe01:
Does it work on Firefox for you? Here no Scan Codes Arme queried.

Von meinem iPhone gesendet

daboe01:
i just tested myself with dvorak. the current code works well in FF but indeed not in chrome.
unfortunately, the solution from <http://stackoverflow.com/questions/10721499/differency-normal-characters-and-half-characters-on-keydown> is not robust enough.

aparajita:
Not working on Safari either.

aparajita:
It only works on FF because only FF supports injecting key events into text areas. See [here](http://stackoverflow.com/questions/596481/simulate-javascript-key-events).

I think your best bet is to trace what's happening in google docs and figure out how they are doing it.

daboe01:
i just reimplemented the chrome part analogous to the FF approach.
works with dvorak in FF and chrome on my mac.
is this good enough?

aparajita:
Das ist super! There are still a bunch of bugs to work out, but the basic approach is working. Great work!

I posted some issues in the CPTextView repo.

daboe01:
thank you :-)

aparajita:
Thanks for the fixes.

daboe01:
the native input stuff is pretty much finished from my side.
any objectives against including it into this PR?

daboe01:
anyone volunteering to test it in IE?

daboe01:
i just performed extensive testing to ensure this branch works as well as the standalone framework.
it is pretty much finished from my side.
@aparajita is there anything more to do before this can be merged in?
fixes #2357

Dogild:
You should resent with the master :)

daboe01:
@Dogild any idea why travis is failing in CPComboBoxTest wit a "document is not defined"?
i have not touched anything there!

Dogild:
The tests should be fixed now, you should merge the master again. Thx

Dogild:
Are you finished @daboe01 ?

daboe01:
yes :-)

Dogild:
Hi @daboe01, can you make a new pull request with this feature and a new pull request with one commit ? @ahankinson any objections for that (your commits won't appear anymore) ?

Dogild:
HI @daboe01, 

Few things to change :

- When the content of the CPTextView is empty, the position of the cursor is not well positioned.
textviewcursor
- Copy and paste does not work in CPTextView when the CPTextView was created from a xib.
- Font is different between what you get from the xib and then with what your write.
- Position of the texts moves when hitting enter at the end of a line
textviewenter
- Position of the cursor is wrong when hitting shift+leftArrow then leftArrow. The cursors should be at the beginning of the word but it is at the end of the previous line
textviewcursor
- In the test CPTextView, when doing cmd+z after loading the application, it erases everything
- When I redo with cmd+shift+z it select things 
- Not possible to select different parts of the text like in cocoa 
screen shot 2015-09-25 at 2 07 35 pm
- Can you create a test with the font panel ?
- In cocoa, the NSTextView supports in native the cmb+b (bold), cmd+i (italic), cmd+u (underline) and so on.
- Impossible to select the text just after you have deleted something (pattern -> delete - delete - shift+alt+leftArrow). It only happens when you dot with the last letters of the content. Otherwise in the middle of something it will select something weird.

Keep up the good work ;), we are very close to release it !

daboe01:
thank you for testing.
i opened appropriate issues here: https://github.com/daboe01/CPTextView/issues
i cannot fix the xib related issues because i do not work with Xcode.
multiple selections are unsupported for now as is justified text.

Dogild:
Hi @daboe01,

Few other things : 

- When selecting a group of letters and then hitting rightArrow, the cursor is positioned one more space that it should (shift + leftArrow than rightArrow)
cursorposition
- There is not any support of drag and drop of text. See example with the cocoa application.
draganddrop
- Deselection happens with mouseDown, it should be with mouseUp

AppKit/CPText.j
```diff
@@ -46,8 +76,283 @@ CPWritingDirectionLeftToRight   = 0;
...
+    if (selectedRange.length < 1)
+            return;
+
+    var pasteboard = [CPPasteboard generalPasteboard];
+
+    // put plain representation on the pasteboad unconditionally
+    [pasteboard declareTypes:[CPStringPboardType] owner:nil];
+    [pasteboard setString:[[self stringValue] substringWithRange:selectedRange] forType:CPStringPboardType];
+
+    if ([self isRichText] && [self respondsToSelector:@selector(textStorage)])
```
Dogild: This is very related to the CPTextView. It should be there

AppKit/CPText.j
```diff
@@ -46,8 +76,283 @@ CPWritingDirectionLeftToRight   = 0;
...
+{
+    _isSelectable = flag;
+
+    if (!flag)
+        [self setEditable:flag];
+}
+
+- (void)setEditable:(BOOL)flag
+{
+    _isEditable = flag;
```
Dogild: We can bind this ivar. We need to add the willChangeValueForKey and didChangeValueForKey

daboe01: Deselection happens with mouseDown:
I do not see any difference to cocoa with respect to selection and mouseup/down.
could you please elaborate?

Drag and drop:
d&d in textview was also not supported in earlier versions of OSX ;-)

Dogild:
About the selection (it's not a big things), in Cocoa when some letters are selected and when we click on this same selection, the deselection will be with mouseUp and not mouseDown.

daboe01:
ah, i see.
this will be automatically fixed as soon as we support drag and drop at some point in the future.

Dogild:
I just fixed few bugs with the CPTextView coming from a xib !

Dogild:
Hi @daboe01,

Few other issue detected :

- When deleting with backspace, it'll delete one character, then two characters, then one, then two etc etc... 
- Tab seems to not work
- cmd+z does not work with the CPTextView when loading from a xib
- cmd+v seems to not work anymore

daboe01:
i do not see the backspace issue here. for me it only deletes a single character.
any more details on reproducing this?

cmd+z and cmd+v work fine for me in the manual test.
do you have a menu with conflicting keyboard shortcuts in you xib?

Dogild:
Sorry, my bad I can't reproduce anymore :/ I have some issues with the xib and ElCapitan. I'll try to do some fixes this week

AppKit/CPColor.j
```diff
@@ -467,6 +467,16 @@ var cachedBlackColor,
     return [[CPColor alloc] _initWithCSSString: aString];
 }
 
++ (CPColor)selectedTextBackgroundColor
+{
+    return [CPColor colorWithHexString:"99CCFF"];
```
Dogild: This should be in the theme as we can theme a CPColor

AppKit/CPColor.j
```diff
@@ -467,6 +467,16 @@ var cachedBlackColor,
...
+}
+
++ (CPColor)_selectedTextBackgroundColorUnfocussed
+{
+    return [CPColor colorWithHexString:"CCCCCC"];
```
Dogild: Same, it should be in the theme

Dogild:
Hi @daboe01, how do you feel with this PR ? Can you sync it with the master please and do a merge --squash to only have one commit please ?

Dogild:
One more thing, could you improve the test CPTextView with a font panel and the rest of the library you did please

AppKit/CPColor.j
daboe01: @Dogild can you help me with this, please? i am a complete theming noop

daboe01: @Dogild I think that the selection color of CPTextField is not themed, isn't it?
what do you think of using the css system colors like this?
JavaScript ``` element.style.backgroundColor = 'Highlight';
element.style.color = 'HighlightText';
```

Comment (Dogild) [inline AppKit/CPColor.j]: Take a look to the file `AppKit/Themes/Aristo2/ThemeDescriptors.j`, line 104, you have :

```

+ (CPColor)themedColor
{
    var color = [CPColor redColor],
        themedColorValues =
    [
        [@"alternate-selected-control-color", [[CPColor alloc] _initWithRGBA:[0.22, 0.46, 0.84, 1.0]]],
        [@"secondary-selected-control-color", [[CPColor alloc] _initWithRGBA:[0.83, 0.83, 0.83, 1.0]]]
    ];

    [self registerThemeValues:themedColorValues forObject:color];

    return color;
}
```
This is the way you define a color for the theme Aristo2. You need to add two new color here, one for the `selectedTextBackgroundColor` and one for the `selectedTextBackgroundColor`.

Then, in `CPColor.j`, line 470 is the perfect example how to access the color defined in the theme :
```
+ (CPColor)alternateSelectedControlColor
{
    return [[self _cachedThemeColor] valueForThemeAttribute:@"alternate-selected-control-color"];
}
```

You will need to do something similar, but with a different key than `@"alternate-selected-control-color"`

**Comment by igerard** [comment]
Hello,

I am starting to try to use this code for my project, 

I use the CPTextView as a distinct framework, and I put the TextView folder in the Framework folder of my project, but I get in the log this kind of message and I would like to understand why we get it even if it seems to work:

WARNING line 79 in     CPTextView      _textView;
                    ^
WARNING line 79 in http://172.30.22.239:8080/Frameworks/TextView/CPTextContainer.j: Unknown type 'CPTextView' for ivar '_textView'

Thank in advance

**Comment by daboe01** [comment]
@igerard this PR does not work standalone. the standalone framework is here: https://github.com/daboe01/CPTextView

**Comment by daboe01** [comment]
@Dogild @aljungberg last reviewer issue resolved. you can merge it in now.

**Comment by mrcarlberg** [comment]
@daboe01 3 test cases in CPTextViewTest is failing. These need to be resolved.

**Comment by mrcarlberg** [comment]
@igerard The warning on line 79 in CPTextContainer.j is just a warning that the compiler does not recognize the class ```CPTextView``` as the type for the instance variable ```_textView```. ```CPTextView``` is a well known class for us all so this warning is due to a missing ```@import``` of CPTextView.j in CPTextContainer.j or a circular import problem. The later (or both) can be solved by adding ```@class CPTextView``` at the beginning of the file CPTextContainer.j

**Comment by daboe01** [comment]
-#needs-unit-test
-#needs-improvement

**Comment by daboe01** [comment]
+#ready-to-commit

**Comment by daboe01** [comment]
@aljungberg are you still around?

**Comment by daboe01** [comment]
@Dogild may i kindly ask to reassign this PR? @aljungberg is apparently currently not in a position to take care of this. I am thinking of @mrcarlberg because oops is already using this code.

**Comment by Dogild** [comment]
@mrcarlberg any objections ?

**Comment by mrcarlberg** [comment]
I can take care of this.

As this is a new feature in the Cappuccino framework and will not break anything I believe that we can merge this and let everybody start using it even that there are issues with this code. We have started to use it in our project and it is very usable. The more people that are testing the faster it can be fixed. Other people can also make PRs if they want to improve the code.

Please feel free to give feedback on this if you think we should do differently.

I will send an email to @aljungberg to get some feedback from him also.

**Comment by Dogild** [comment]
+1

**Comment by mrcarlberg** [comment]
I have talked to @aljungberg and we both think it is time to merge this. Alexander has some comments that we want to address. Also, I have one performance issues when handling larger texts. I will comment that in a separate comment.

**Comment by mrcarlberg** [comment]
We are using this in our project and it works really well. The only problem is when the text is a little longer. I tested a text with around 50000 characters and it takes many seconds to "open" the text view. After that it is ok in speed when editing the text.

I did a quick profiling in Chrome and found that most of the time is spent in the ```CPLayoutManager``` and ``` ```CPSimpleTypesetter```. 

It looks like it loops over all characters but also has some kind of inne loop from the beginning to the current character. There is also a third inner loop in the _objectsInRange. All this destroys the performance when using larger texts.

The stack trace looks like this:
VM11904:1evaluate_objectsInRange
CPLayoutManager.js:76- [CPLayoutManager setTextContainer:forGlyphRange:]
CPLayoutManager.js:835- [CPSimpleTypesetter _flushRange:lineOrigin:currentContainer:advancements:lineCount:sameLine:]
CPTypesetter.js:146- [CPSimpleTypesetter layoutGlyphsInLayoutManager:startingAtGlyphIndex:maxNumberOfLineFragments:nextGlyphIndex:]
CPTypesetter.js:283- [CPLayoutManager _validateLayoutAndGlyphs]
CPLayoutManager.js:548- [CPLayoutManager textStorage:edited:range:changeInLength:invalidatedRange:]
CPLayoutManager.js:667- [CPTextStorage processEditing]
CPTextStorage.js:123- [CPTextStorage endEditing]
CPTextStorage.js:142- [CPTextStorage replaceCharactersInRange:withAttributedString:]
CPTextStorage.js:204- [WebComponentTabController placeTextInView:]

I'm not very familiar with how ```CPLayoutManager``` works but maybe you (@daboe01) can give us some information if this is possible to speed up?

**Comment by daboe01** [comment]
@mrcarlberg I do not think that the bottleneck is in _objectsInRange which does a binary search in a small sorted list. The initial lag you observe is due to typesetting the full document. While display is lazy, typesetting needs to be done completely before anything will work (e.g. for the scroller to know the document size). This was also the case in  NeXTSTEP and in early cocoa, where i vividly remember the lag after opening large documents in TextEdit.app. I think that apple solved this by putting the layout in a background thread, which we can't do in the web browser. We could, however, cut layouting into chunks that can be executed in the background via setTimeout:. This will increase overall complexity, though. What do you think?

**Comment by mrcarlberg** [comment]
@daboe01 Yeah, I remember that NeXTStep time too. 😄 
But When I profile I find that 1/3 of the time is spend in ```_objectsInRange``` and 99% of the time it returns an empty array when we set the text for the first time. From my understanding it does not do a binary search as it just loops all lines and checks if the range of the current character is anywhere in the already created lines. And it is not.

1/5 of the time is spend in ```_rescuingInvalidFragmentsWasPossibleForGlyphRange:``` and it also has a loop that tries to find the line for the current character.

The more lines that are added more time is going to be spent in these loops and the slower it will become. Maybe doing a binary search to find the lines faster is one way to do it. This function and method are called many many times so another way might be to understand what the ```CPLayoutManager``` is trying to do and do it more efficiently.

I don't think doing this in the background is the way we want to solve this.

**Comment by mrcarlberg** [comment]
Maybe a reverse loop can do the trick in ```_objectsInRange```. I don't know if the lines are always in  order but if they are it can just bail out. Most of the times it will just check the last line and find that the range is after that line and return an empty array.

Maybe this approach can be used in the ```_rescuingInvalidFragmentsWasPossibleForGlyphRange``` method too?

This is just some ideas as I really don't have any understanding how the ```CPLayoutManager``` works

**Comment by daboe01** [comment]
thank you! i will look into optimizing _objectsInRange asap.

**Comment by daboe01** [comment]
@mrcarlberg here is idea that works for text of any size:
split your text into chunks of say 3K and append each chunk individually at the end.
you have to do this via setTimeout or via the runloop.

i think this will mimic the cocoa behaviour in which you immediately see the text and the app is responsive while the layouting is going on in the background (the scroller continuously thrinks).
you could do this in a cover function for replaceCharactersInRange:withAttributedString:

**Comment by mrcarlberg** [comment]
That is a good idea to make it work better but maybe 90% of the time will be on the last chunk 😄 

Great that you will look at ```_objectsInRange``` and ```_rescuingInvalidFragmentsWasPossibleForGlyphRange:```

**Comment by mrcarlberg** [comment]
@daboe01 I might have found one more bug 😄 ```[CPTextView setTextColor:range:]``` does not honor ```[CPTextStorage beginEditing]``` and ```[CPTextStorage endEditing]```

So when I the following it will do the typesetting twice:
``` objective-c
[textStorage beginEditing];
[textView setString:@"Martin"];
[textView setTextColor:myColor];
[textStorage endEditing];

**Comment by daboe01** [comment]
@mrcarlberg i just replaced the linear search with an experimental binary search just to verify your timing. This does not substantially decrease loading time in my testing (this takes around three seconds to render the source code of CPTextView.j as does the linear implementation). What do you think?
```ObjJ
var _objectsInRange = function(aList, aRange)
{
    var firstIndex = [aList _indexOfObject:aRange.location sortedByFunction:_sortRange context:nil],
        lastIndex = [aList _indexOfObject:CPMaxRange(aRange) sortedByFunction:_sortRange context:nil];

    return aList.slice(firstIndex > 0 ? firstIndex - 1 : 0, lastIndex === CPNotFound ? aList.length : lastIndex + 1);
}
```

daboe01:
@mrcarlberg do you know whether setTextColor:range: honors beginEditing / endEditing in apple's implementation? from my understanding, beginEditing / endEditing apply only to the model layer

mrcarlberg:
Yeah, you might be right about ```setTextColor:```. I have to check how Cocoa works.

I guess I have to profile your new ```_objectsInRange``` to know what is happening in it. I don't know how the```_indexOfObject:sortedByFunction:...``` method works as I didn't find it in the Foundation.
I think the easiest trick should be, as I wrote, a reverse loop as the ```_objectsInRange``` functions returns an empty list almost every time when setting a new text string. The list has to be already sorted for that to work and I don't know if that is the case.

An other ways might be possible too? It is a big waste of time trying to figure out if the next character is already included every time by looping a list of ranges and it is always not included. It is still a waste of time if we are doing a binary search instead of looping. Maybe this can be improved somewhere above this function in the typesetting. The ```_objectsInRange``` functions is not need to be called at all in this case.

Should be possible to do huge performance gains.

daboe01:
@mrcarlberg take a look at commit 71f033cc42f1bbcd61cb2e961d023b75626dd818 that i just committed to the standalone version.
this should eliminate the calls of _objectsInRange during typesetting altogether.
however, this commit does not noticeably speed up typesetting in my subjective testing.
keep in mind that the loops are not performed for every character but for every line.
how many lines do you have in your test document?

daboe01:
@mrcarlberg i have now thorougly optimized _objectsInRange. does this fix your performance issues? it seems to me that typesetting performance is highly browser dependent with chrome surprisingly being slowest by far.

mrcarlberg:
@daboe01 I have created a pull request with some small optimisations for the typesetting on your CPTextView repository: https://github.com/daboe01/CPTextView/pull/113

Please test it and let me know if it is better than the current implementation?

daboe01:
hi martin,

thank you very much for these suggestions.
please forgive me, i am a git noop.
can i test this PR locally without merging?

did you do any profiling yourself?
you see a noticeable performance improvement?

please keep in mind that the parts about sizing that you changed are handled differently in the capp PR.
they have been moved to DOM/CPPlatformString.j

best greetings,

daniel

daboe01:
@mrcarlberg i have added some line notes to daboe01/CPTextView#113

daboe01:
@mrcarlberg on my macbook pro from 2011, the current head of standalone version renders the source code of CPTextView.j (>80K) in less than a second (on safari). do you really still have performance issues with the current head of standalone version?

mrcarlberg:
In our project we use your CPTableView framework. I just did a "profile" in Chrome on my MacBook Pro (Retina, Mid 2012) and I think it was ```[CPLayoutManager _validateLayoutAndGlyphs]``` that took 1.7 seconds to execute. After my fixes it takes 350ms. My text has around 49500 characters.

You might not like some of the performance tuning I have done but it sure looks like it is working. I'll answer your line notes one by one at daboe01/CPTextView#113

daboe01:
wow!!
i will have a second look at your changes :-)

mrcarlberg:
To get my version you can do a:
``` bash
git clone https://github.com/mrcarlberg/CPTextView.git
git checkout faster_typesetting
```

daboe01:
@mrcarlberg i fixed a regression in the standalone. thereafter, i ported all your stuff to this PR. Everything works in my testing. Do you see issues in your application?

daboe01:
@mrcarlberg chapeau :-)

mrcarlberg:
Good that you found the last one. I was just writing a comment about that :smile:

mrcarlberg:
What kind of performance gains have you seen when testing?

daboe01:
the users will see mostly none :-( unless they paste twice.
this is because 90% is text sizing anyway.

mrcarlberg:
It is not necessary to paste the same text. You can open a similar or the same text and you get the performance boost.

mrcarlberg:
I think this is a huge improvement. If the browsers will get faster sizing in the future we will benefit from it directly as the typesetting is pretty fast now.

daboe01:
@mrcarlberg you are right! on safari you see the benefit right away :-)

mrcarlberg:
I will test this a little more but I think it is time to merge this PR into the master. This is a big new feature for the Cappuccino framework. Really great work from you. We should write an announcement on the mailing list when it hits the master branch. I think it is better if you write and post it yourself as you have done all the hard work and should get all the credit. The core developers can help you with feedback if you want.

daboe01:
@mrcarlberg we should get PR #2462 in before going public. Otherwise, rich copy and paste is broken. You may ask @kerusan about the background ;-)

mrcarlberg:
Ok, we can take care of that one at the same time. Please prepare what you want to include in the announcement. It is nice if we can have that ready when we merge.

daboe01:
@mrcarlberg sure, i will prepare the announcement and put it into the first comment of this PR for review

daboe01:
@mrcarlberg doen't this case study contradict @aparajita 's idea of eliminating all #defines from the capp sources? shouldn't we re-discuss this on the list?

mrcarlberg:
Are you talking about the replace of, for example, the ```#if PLATFORM(DOM)``` macro with a Javascript if statement that will be executed every time instead of once at the compile time?

aljungberg:
Let's not eliminate defines until we need to. Anything we can do compile time rather than at runtime is a win.

mrcarlberg:
I agree 100%

daboe01:
I was thinking about MIN and MAX specifically

Von meinem iPhone gesendet

mrcarlberg:
Ok, so you mean the difference when MAX is a Javascript function or a preprocessor macro?

I don't think there is something wrong with Aparajitas idea about that. The reason I did remove the use of MAX and MIN was just to save some nano seconds. Some from the function call and some from the store of the result. If we save 100 ns on each character and do it 100000 times we save 10ms. This is not much and usually not worth the trouble except in a case like this when you want it to be super fast.

daboe01:
@mrcarlberg yes that was my point exactly. good to here that the masterplan is ok :-)

daboe01:
@mrcarlberg do you also see a performance regression with chrome 53.0.2785.89?

mrcarlberg:
I first checked with 52 and then upgraded to 53 and I can't find any difference in performance between the two. What are you seeing exactly?

daboe01:
sorry, it has nothing to to with chrome 53. it was a bug that prevented size caching. fixed by 	8cfd3ae

daboe01:
@mrcarlberg 315fc21 is a optimization that speeds up editing very large approximately 10 fold!

daboe01:
CPTextView is super fast now :-)

daboe01:
@mrcarlberg i have updated the first comment of this PR to my proposal for the release message for review.

mrcarlberg:
Merged! Thanks!

Issue #1435, opened 2012-01-19
CPTextField + Bindings: textDidEndEditing sent too late when controller selection changes [+1]


Toxe:
Application setup

a CPArrayController that stores a couple of simple objects
a CPTableView with one column
the table column values are bound to the arrangedObjects of the array controller
a CPTextField
the "value" binding of the text field is bound to a key path of the controller, "selection.translations.en" in this case
If you select an entry from the table the value of the text field changes as well. This is all very basic stuff and nothing fancy as you can see in this screenshot.

The problem

If I enter text into the text field and then select another table entry the textDidEndEditing notification is sent after the selection changed. The consequence is that the newly entered value wont be saved. Let's have a look at some log output:

(1) info: textDidFocus
(2) info: insertText 1info: textDidBeginEditinginfo: textDidChangeinfo: insertText 2info: textDidChangeinfo: insertText 3info: textDidChange
(3) info: (Controller) setValue:[number of indexes: 1 (in 1 range), indexes: (2)] forKeyPath:selectionIndexes
(4) info: setObjectValue:MPEG Video
(5) info: textDidEndEditing
(6) info: (Controller) setValue:MPEG Video forKeyPath:selection.translations.eninfo: (Controller) setValue:MPEG Video forKeyPath:selection.translations.en
(7) info: textDidBlur

I clicked into the text field to start typing.
I started typing 1, 2 and 3 and we see the textDidBeginEditing and textDidChange notifications.
I clicked on another table entry and the controller selection changes.
The value of the newly selected table entry is written into the text field. This overwrites our 123 text we just entered.
Finally the text field receives its textDidEndEditing notification.
Now that the text field believes that editing has ended it informes the controller and sends it what the text field believes to be the newly entered text.
The text field loses its input focus.
See the problem? The text field sends the wrong value to the controller. It should send 123, but instead it sent MPEG Video, which is the text of the newly selected table element.

The textDidEndEditing and textDidBlur notifications should probably be sent right before or after step (3), where the controller changes its selection.

Note that if I enter the text into the text field and instead of selecting a new table entry I simply press enter or click right outside of the text field everything works fine.

Workaround

A simple workaround that seems to work fine in this case was that I subclass CPTextField and overwrite textDidChange: and after every entered character manually send a textDidEndEditing notification.

(void)textDidChange:(CPNotification)note
{
CPLog(@"textDidChange");
[super textDidChange:note];
if (_isEditing) {
    _isEditing = NO;
[...]

PR #2169, opened 2014-08-10
New: CPTextField now supports the CPEditorRegistrationProtocol

daboe01:
previously, changing the selection of a CPArrayController whilst editing a bound text field could result in losing these edits.
this PR makes the textfield invoke the objectDidBeginEditing and objectDidEndEditing methods of the CPArrayController appropriately.
additionally, all dirty changes are committed before the selection changes.

fixes issue #1435


AppKit/CPTextField.j
```diff
@@ -1933,6 +1933,56 @@ var CPTextFieldIsEditableKey            = "CPTextFieldIsEditableKey",
...
+        notificationCenter = [CPNotificationCenter defaultCenter];
+
+    if (theBinding)
+    {
+        [notificationCenter removeObserver:[theBinding._info objectForKey:CPObservedObjectKey]
+                                      name:@"CPControlTextDidBeginEditingNotification"
+                                    object:anObject];
+
+        [notificationCenter removeObserver:[theBinding._info objectForKey:CPObservedObjectKey]
+                                      name:@"CPControlTextDidEndEditingNotification"
```
Dogild: Here you should use the var `name:CPControlTextDidBeginEditingNotification`

AppKit/CPTextField.j
```diff
@@ -1933,6 +1933,56 @@ var CPTextFieldIsEditableKey            = "CPTextFieldIsEditableKey",
...
 
++ (void)unbind:(CPString)aBinding forObject:(id)anObject
+{
+    var theBinding = [self getBinding:aBinding forObject:anObject],
+        notificationCenter = [CPNotificationCenter defaultCenter];
+
+    if (theBinding)
+    {
+        [notificationCenter removeObserver:[theBinding._info objectForKey:CPObservedObjectKey]
+                                      name:@"CPControlTextDidBeginEditingNotification"
```
Dogild: Here you should use the var `CPControlTextDidBeginEditingNotification`

Dogild:
@cacaodev and @daboe01 what's going on with this one ?

daboe01:
imho, we can merge this in.

cacaodev:
This PR effectively solves the issue mentioned but it could be generalized
- The CPEditorRegistrationProtocol can be implemented in all controls that support editing, not just CPTextField.
- In CPArrayController there are other cases than selection change when we need to review all editor pending changes. They should be covered, including the selection change, by the wider concept described by the methods commitEditing: (forces to end editing) and discardEditing: (pending changes are lost).

@daboe01 Do you want to work on this or do you prefer to merge as it is ? In this case, I will just ask you to document the fact that this is a partial implementation. In code comments or by opening a new PR.

cacaodev:
Also, it seems that with this PR, when the array controller selection changes, editor changes are lost. Is it true and did you check the cocoa behavior ?

cacaodev:
Take a look at CPController, there is a commitEditing: method implemented there that calls commitEditing: on the editors. The solution could be as simple as calling it from the _selectionWillChange method in CPArrayController add a commitEditing protocol to CPObject that will just reverseSet the binding.

One last thing: when an editor commits its value, it needs to be removed from the list of active editors in CPController, don't you think ?

daboe01:
@cacaodev thank you for the valuable insight. i will look into all this as soon as possible.
+#needs-improvement

daboe01:
-#needs-review

daboe01:
-#needs-confirmation

daboe01:
@cacaodev it is removed by the call to objectDidEndEditing that is triggered by the CPControlTextDidEndEditingNotification

daboe01:
@cacaodev @mrcarlberg i think it is time to merge this in as it improves the structure and cocoa compatibility of our code-base right now.
i will work on the generalization with future pull requests as soon as the need arises on my side.

mrcarlberg:
@cacaodev What do you think about this PR? How should we move this forward?

cacaodev:
@mrcarlberg As I said before, as far as I know, CPEditorRegistrationProtocol is not really for user consumption but should be used inside CPArrayController.
We can merge this but without a change in CPArrayController I am afraid this will just be a nice sleeping feature (_la belle endormie_ in french).

mrcarlberg:
@cacaodev I do agree that it should be good if this PR can cover more cases and be more generic. But it does give us some basic functions that can be used and built upon for future enhancements. I think we should merge this if we document in the code what is missing.

@daboe01 Can you add comments about the issues @cacaodev talks about above. It is good if we can have this documented in the code.

cacaodev:
@mrcarlberg that's good for me

daboe01:
@cacaodev @mrcarlberg i am a bit unsure where to place the comments and what to comment exactly.

should i add FIXME-comments to all our editable controls that support binding to call objectDidBeginEditing: and objectDidEndEditing?

mrcarlberg:
Yes, you can add FIXME comments on the controls. I also think it is a good idea to add the two issues in the comment from 14 Dec 2016 to some file. Maybe in the CPObjectController.j file above the added method is a good place?

daboe01:
milestone=1.1

daboe01:
-#needs-improvement
+#ready-to-commit

daboe01:
@mrcarlberg can we merge this one in, eventually?

mrcarlberg:
Merged! Thanks!

PR #2227, opened 2014-10-23
Fixed: constructor of CPDate can take string param

Dogild:
Previously, when making a CPDate with some constructors, we could pass a string or nil to create a date. For example it was allowed to do var date = [CPDate dateWithTimeIntervalSince1970:"10"];

Now this is not possible anymore and it works like in Cocoa, only number are authorized.


primalmotion:
would be cool to put a warning first before going hardcore

primalmotion:
+#needs-improvement

Dogild:
Done !

primalmotion:
+#fixed

Thanks!

PR #2240, opened 2014-10-30
New: added method initWithWindow in CPPlatformWindow 

Dogild:
The constructor initWithWindow: initializes a new CPPlatformWindow and set the given CPWindow as a fullPlatformWindow and bridgless window.
When using this constructor, you don't need to work with the platformWindow anymore, you can use the method orderFront: and orderOut: of the CPWindow to open or close the CPPlatformWindow.
You can also use the method setFrame: of the CPWindow to automatically change the contentRect of the CPPlatformWindow. In one word we assume that the given CPWindow and the CPPlatformWindow will have the same behavior. When not using the constructor initWithWindow:, the method setFrame: of the given CPWindow won't do anything to the CPPlatformWindow.

Then previously, when calling the method orderOut: on the full platform window of a CPPlatformWindow, cappuccino didn't close the parent CPPlatformWindow. When he does, this only works if the parent platformWindow is not the primary platform.

Test app in Tests/Manual/CPPlatformWindow/


Dogild:
@cacaodev @primalmotion any chance to take a look to this one ?

primalmotion:
+#fixed

Thanks

PR #2243, opened 2014-11-01
New: CFError and Updated CPError

ahankinson:
This pull request re-works and expands the existing CPError by providing an underlying implementation in CFError(). CPError is now toll-free bridged with CFError, mirroring the implementation in Cocoa.

Additionally, this pull request adds support for CPURLErrors and CFNetworkErrors, and maintains the same error codes used in Cocoa. These have been copied verbatim and may need to be adjusted on a case-by-case basis for their applicability in a Cappuccino context.

Basic unit tests have been provided for both CPError and CFError.

Finally, a small change was made to the CPCoder class, where `encodeValueOfObjCType` was renamed to `encodeValueOfObjJType`.


primalmotion:
+#fixed

Thanks!

PR #2252, opened 2014-11-06
Fixed: Doxygen configuration

ahankinson:
The Doxygen configuration file caused warnings to appear about obsolete tags. (The version of doxygen that the file was created for was 1.7.1, which was released in 2010.)

This commit uses the automatic doxygen updater (doxygen -u) to update it to a more recent version (1.8.7).

Additionally, this commit changes the .doxygen file to build a docset by default. The effect of this is to simply emit a `Makefile` in the output directory that can be further used to compile a docset. Parameters for the project publisher and name were added for this docset.


primalmotion:
Doesn't seem to work. That's what I get after doing a jake docs
screen shot 2015-01-08 at 17 32 20

primalmotion:
there is no class list or anything

ahankinson:
I must have missed that config. I'll fix it later today.

primalmotion:
-#needs-review
+#needs-improvement

ahankinson:
I just updated the doxygen configuration file to add the sidebar with `jake docs`.

-#needs-improvement
+#needs-review

primalmotion:
+#fixed

Thanks

PR #2264, opened 2014-11-26
Fixed: Memory leak in CPTableView, CPScrollView [+1]

Dogild:
Prevously, when removing a CPTableView or a CPScrollView, the CPNotificationCenter kept a reference of these observers in the notification center.

Now, the CPNotificationCenter does only have a observer when necessary.


primalmotion:
+#fixed

Thanks!

PR #2272, opened 2014-12-04
Rework of withCredentials, CPURLRequest, and CPURLConnection

ahankinson:
This Pull Request deals with three inter-related issues.

1. The presence of `withCredentials` on CPURLConnection required that the connection be initialized with credentials or not. This commit moves `withCredentials` to CPURLRequest, where it can be set prior to initializing the connection. This setting will then get picked up and transferred to the underlying CFHTTPRequest as needed. This also simplifies the implementation of `withCredentials` in Capp.

2. CPURLConnection was missing the `originalRequest` and `currentRequest` methods, present in Cocoa since 10.8. These were added.

3. CPURLRequests were not able to be deep copied using the `copy` method (provided by CPCopying). The CPCopying protocol was added to CPURLRequest to implement the copy method allowing CPURLRequest objects to be deep copied.

Refs: #2158, #2237, #2209.

### DIFF
--- a/Foundation/CPURLConnection.j
+++ b/Foundation/CPURLConnection.j
@@ -81,8 +81,6 @@ var CPURLConnectionDelegate = nil;
     BOOL            _isCanceled;
     BOOL            _isLocalFileConnection;
 
-    BOOL            _withCredentials    @accessors(property=withCredentials);
-
     HTTPRequest     _HTTPRequest;
 }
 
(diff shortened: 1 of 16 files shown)


primalmotion:
Seems to work well. 

@ahankinson You mention some issue with IE. are they introduced with this PR and was working before, or it was not working anyway?

ahankinson:
I don't think it would work anyway. The issue is that IE doesn't implement CORS in the way the other browsers do. You can very easily test the behaviour using the manual test I've included in the PR. There's a README in that test that explains how to set it up.

primalmotion:
@ahankinson ok so are you done with this one? it looks good to me

ahankinson:
Yep, I'm done with it.

primalmotion:
+#fixed

Thanks!

PR #2278, opened 2014-12-09
Animations: CPAnimationContext & animator

cacaodev:
Here it is. Animations for cappuccino. Animations are implicit from the capp API point of vue, but explicit in the DOM. You define animations for a given key path on an target, then ask the -animator of this target to change the keypath value and the animation is run implicitly. In the backstage, a css animation is created, inserted into the stylesheet, and run.

Historically, in cocoa and cocoaTouch, several APIs have been added for animations. The CPAnimationContext & animator is the one currently used in cocoa, including the most recent versions. I think this approach is great for built-in animations in the framework. For example, it's easy to decide that when a CPViewController loads its view, the subviews will automatically fade-in when presented.
But from a user perspective, the API may seem a bit complex with the need to set default properties on CPAnimationContext and use the -animator instead of calling directly the view you want to animate. That's why we may consider to add the UIKit API where you can just do this on CPView:
```
+ (void)animateWithDuration:(NSTimeInterval)duration
                 animations:(void (^)(void))animations
                 completion:(void (^)(BOOL finished))completion
```

In this commit there are no built-in animations in the framework. What are the animations would you like to see when you interact with controls in cappuccino ?

There is a test there:Tests/Manual/CPAnimatablePropertyContainerTest/


primalmotion:
@cacaodev Looking at the test, it works great. awesome PR that I'm planning to use :)

I'm not very familiar with Core Animation, so I would trust you on this, but is there any breaking changes in the way people used the CPViewAnimation, or everything is well separated?

If it's well separated, I think it would safe to merge it in, let people start using it, and deal with bugs with the flow.

@aparajita @aljungberg comments?

cacaodev:
@primalmotion CPViewAnimation and the CPAnimatablePropertyContainer protocol are totally different. You can continue to use CPViewAnimation like before but anything that is done with CPViewAnimation can be done with the animator proxy, and you get css animation support for free.

primalmotion:
@cacaodev is it possible to have automated tests for this one?

cacaodev:
@primalmotion It is really not easy to ojtest this thing because the css animation generated is run asynchronously. For example, you cannot assert anything based on the end values. If there are other parts you can think of, I can try to make tests.
From the console, I can tell you that the duration between setting the animated key path and the resulting  css animation seems ok: less than 10 ms.

primalmotion:
I think we can give it a go. Can you sync up with master and that in? Can't wait to play with it :)

cacaodev:
OK but then I will need some help with the "file stripping" feature for deploy apps. I really want to make sure that if an application does not use animation (i.e. ask for the -animator proxy), none of this stuff is imported, including the js files. Maybe this is something we should see in relation with the new compiler.

We also desperately need automated tests that can manipulate the DOM. I feel bad releasing this without ojtests :)

Then we can discuss what built-in animations for controls we would like to have first. I already have a very cool animation for the segmented control selection in a branch. I would like to also support CPViewController (fade-in subviews) and the check box image but I realize that it would require some substantial rework of these classes layout code.

Dogild:
One solution for manipulating the DOM could be Cucapp and Cucumber. 

@cappuccino/core-devs Is it possible to add that to travis ? Any other way to manipulate the DOM ?

cacaodev:
There is http://phantomjs.org . I never used it but it seems to be popular. And Travis CI compatible.

Dogild:
Yep right, Travis comes with phantomjs or firefox. I would prefer firefox as it's a real browser

primalmotion:
@cacaodev are you feeling ok with this one?

Dogild:
@primalmotion I'll let you merge this one asap

primalmotion:
Awesome! Thanks!

PR #2286, opened 2014-12-14
CPURLConnection +sendAsynchronousRequest:queue:completionHandler:

cacaodev:
NEW: CPURLConnection -operation, CPError CPURLErrorDomain

CPURLConnection -operation gives access to the operation created with the completionHandler argument. This is an addition compared to cocoa.
This allows to create dependencies between operations and setup priorities.
When the connection fails (with a status code indicating an error) or is canceled, the operation is also canceled and the next operation in the queue is started.

With test Manual/CPURLAsynConnectionTest/ that requires to be run in a web server with php.
Known issue: if you cancel a connection before it received any response, the dependent operations are not performed. I need to check if this is a bug in CPOperation or not. [FIXED by #2287]


cacaodev:
@ahankinson : i merged against your recent changes in master. Does it look ok to you.
known bug: check statusCode == 0
It would also be great to have more custom errors & descriptions for the most common network errors.

Foundation/CPURLConnection.j
```diff
@@ -138,6 +143,18 @@ var CPURLConnectionDelegate = nil;
 }
 
 /*
+    Loads the data for a URL request and executes a function on an operation queue when the request completes or fails.
+    @param aRequest contains the URL to obtain data from.
+    @param aQueue The operation queue to which the function is dispatched when the request completes or failed.
+    @param aHandler The function to execute.
+    @discussion If the request completes successfully, the data parameter of the function contains the resource data, and the error parameter is nil. If the request fails, the data parameter is nil and the error parameter contain information about the failure.
+*/
++ (CPURLConnection)sendAsynchronousRequest:(CPURLRequest)aRequest queue:(CPOperationQueue)aQueue completionHandler:(Function)aHandler
```
Dogild: Can we use `usingBlock:` instead of `completionHandler:`. It's more consistent with the other API of Cappuccino

Foundation/CPURLConnection.j
```diff
@@ -161,26 +178,51 @@ var CPURLConnectionDelegate = nil;
...
+    return self;
+}
+
+- (void)_initWithRequest:(CPURLRequest)aRequest
+{
+	_request = aRequest;
+    _originalRequest = [aRequest copy];
+	_isCanceled = NO;
+
+	var URL = [_request URL],
```
Dogild: Style here, too many spaces or tab

Dogild:
Any chance to ass a manual test ?

cacaodev:
There is a manual test under Manual/CPURLConnectionAsyncTest, although you may find it to be not very explicit. Tell me what you think. It mainly tests the operation queue.

Dogild:
I don't see this test (Manual/CPURLAsynConnectionTest/) in the PR. You have probably removed it when rebasing/reverting (whatever) .

cacaodev:
@Dogild sorry for that, here it is.

Dogild:
Tested and looks good to me ! Thanks

PR #2287, opened 2014-12-16
FIXED : a CPOperation marked as cancelled was never started 

cacaodev:
... (CPOperation -start not called), when managed in a queue.

FIXED : After an operation started, either from a queue or explicitely, its finished property was never set to true.

This was causing dependant operation to never be executed.
After this commit, cancelled operations are started (but not executed) and correctly marked as finished.

Tests: CPOperationQueueTest and CPOperationTest for an operation managed by a queue or run explicitely.


primalmotion:
+#fixed

Thanks!

PR #2288, opened 2015-01-02
Fixed: -showRelativeToRect:ofView:preferredEdge is ignoring the positioning rectangle

daboe01:
previously, the popover was positioned centered to the positioning view regardless of the given positioning rectangle.

this bug had been introduced by commit 8b29369 and all credits for the fix go to @Dogild

### DIFF
--- a/AppKit/_CPPopoverWindow.j
+++ b/AppKit/_CPPopoverWindow.j
@@ -370,13 +370,6 @@ var _CPPopoverWindow_shouldClose_    = 1 << 4,
         _targetView = positioningView;
     }
 
-    if (positioningView !== _targetView)
-    {
-        [[_targetView window] removeChildWindow:self];
-        [self _removeFrameObserver];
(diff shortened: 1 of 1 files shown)


AppKit/_CPPopoverWindow.j
```diff
@@ -369,6 +370,13 @@ var _CPPopoverWindow_shouldClose_    = 1 << 4,
         _targetView = positioningView;
     }
 
+    if (positioningView !== _targetView)
+    {
+        [[_targetView window] removeChildWindow:self];
+        [self _removeFrameObserver];
+        _targetView = positioningView;
+    }
+
```
Dogild: You have now two times the same condition here

daboe01:
thanks, fixed.

daboe01:
it is still working.
best greetings,
daniel

primalmotion:
+#fixed

Thanks!

PR #2292, opened 2015-01-13
New: option to active or not xCodeCapp when opening the errors & warnings panel

Dogild:
Previously, when xCodeCapp opened the errors & warnings panel, xcc was always the new active application of the system.
Now an option is available in the preferences of xcc to disable this possibility. If disable the errors & warnings panel will open itself, but your favorite text editor will still be the active application of your system (you will keep the focus).


primalmotion:
+#fixed

Thanks

PR #2294, opened 2015-01-14
Fixed: CPStepper class method returns `id` instead of CPStepper type

### LINKED ISSUE #2293: AppKit/CPStepper.j initializers use CPStepper not self
In AppKit/CPStepper.j the class methods stepperWithInitialValue and stepper are declared to return (CPStepper) instead of (id) and use [CPStepper alloc] instead of [self alloc], which makes subclassing harder when using those methods

ahankinson:
Previously the CPStepper class returned the `CPStepper` type instead of the `id` type. This caused warnings to be raised when initializing subclasses of CPStepper.

This commit fixes this issue by specifying the `id` type in the return values.

This commit also corrects instances of `==` to `===` in CPStepper.

Fixes #2293


ahankinson:
-#new
+AppKit
+bug
+#needs-review
milestone=0.9.8

primalmotion:
It doesn't seem to merge automatically. Can you take a look? Thanks

ahankinson:
Ok, I just re-synced it with master. Does that help (It's showing OK for me)

primalmotion:
+#fixed

Thanks!

PR #2295, opened 2015-01-15
Fixed: showcase of themes are broken

Dogild:
Previously the showcase of the themes were broken. This was due to a change of the CPCollectionView, previously we used to use the delegate collectionViewDidChange. This delegate method has been removed. We now use an observer on the selectionIndexes.


ahankinson:
Tests fine. Thanks!

-#new
+bug
+Theme
+#ready-to-commit
milestone=0.9.8

primalmotion:
+#fixed

Thanks!

PR #2302, opened 2015-01-22
Formatting: CPView

ahankinson:
Previously CPView used "==" as an equality checker. This commit changes these to use "===" for consistency.

### DIFF
--- a/AppKit/CPView.j
+++ b/AppKit/CPView.j
@@ -380,7 +380,7 @@ var CPViewFlags                     = { },
 */
 - (void)setToolTip:(CPString)aToolTip
 {
-    if (_toolTip == aToolTip)
+    if (_toolTip === aToolTip)
         return;
 
     if (aToolTip && ![aToolTip isKindOfClass:CPString])
@@ -525,7 +525,7 @@ var CPViewFlags                     = { },
     [[self window] _dirtyKeyViewLoop];
 
     // If this is already one of our subviews, remove it.
-    if (aSubview._superview == self)
+    if (aSubview._superview === self)
     {
         var index = [_subviews indexOfObjectIdenticalTo:aSubview];
 
@@ -796,7 +796,7 @@ var CPViewFlags                     = { },
 
     do
     {
-        if (view == aView)
+        if (view === aView)
             return YES;
     } while(view = [view superview])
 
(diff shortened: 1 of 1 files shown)


AppKit/CPView.j
```diff
@@ -1900,10 +1900,12 @@ var CPViewFlags                     = { },
             CPDOMDisplayServerSetStyleSize(_DOMImageParts[0], size.width, size.height);
         }
         else
+        {
             _DOMElement.style.background = colorCSS;
 
             if (patternImage)
```
primalmotion: are you sure this was in the if else condition?

primalmotion:
-#new 
+#needs-improvement

primalmotion:
@ahankinson any update on my previous comment?

AppKit/CPView.j
aparajita: Looks like it should be part of the else block (and thus this was a bug), since lines 1891-1892 test for patternImage also and call CPDOMDisplayServerSetStyleBackgroundSize with very different args.

daboe01:
i came to the same conclusion :-)

AppKit/CPView.j
```diff
@@ -796,7 +796,7 @@ var CPViewFlags                     = { },
 
     do
     {
-        if (view == aView)
+        if (view === aView)
             return YES;
     } while(view = [view superview])
```
aparajita: I'm pretty sure the style guidelines say the while should be below the }.

AppKit/CPView.j
```diff
@@ -1525,7 +1525,7 @@ var CPViewFlags                     = { },
         {
             do
             {
-               if (self == view)
```
aparajita: Only 3 spaces of indent in this block and the inner block.

AppKit/CPView.j
daboe01: @aparajita i did not find the newline explicitly demanded here: https://github.com/cappuccino/cappuccino/blob/master/CONTRIBUTING.md#cappuccino-coding-style-guidelines

daboe01:
-#needs-improvement
+#ready-to-commit

AppKit/CPView.j
aparajita: @daboe01 The guidelines state that braces are always on a separate line.

daboe01: ok, i see.
@ahankinson could you please insert a newline between brace and else?

mrcarlberg:
There are some formatting issues in the CPView.j file as noted above. This PR fixes a bug and use ```===``` instead ```==``` etc. for consistency and has nothing to do with the formatting issues.

mrcarlberg:
Merged! Thanks!

PR #2303, opened 2015-01-28
FIXED: The text color of a checkbox inside a selected table row was black

cacaodev:
Now it is white. The text-color values match the CPTextField values for the same theme states.

TEST: Tests/Manual/CPTableViewLion/


primalmotion:
+#fixed

Thanks

PR #2304, opened 2015-02-04
New: A private notification is sent to every subviews when the bounds or the frame of a view changed

Dogild:
Previously, when the frame or the bounds of a view changed, the subviews of this view were not notified by this event.
Now they are, every subview will send a private notification that one of their parent's frame/bounds has changed.

This new feature fixes the automatic positioning of a popover when the targeted view's frame change.


primalmotion:
that's a freaking lot of notifications. @aljungberg @aparajita comments?

primalmotion:
+#needs-review

aparajita:
Sorry I didn't comment sooner. There has to be a better way to deal with the popover problem than this.

Dogild:
This PR has been closed. I'm about to propose a new way to fix the popover's problem.

PR #2305, opened 2015-02-04
Fixed: Duplicate unit-tests

Dogild:
Previously, some unit-tests had the same name in a test file. Now they have unique name.


primalmotion:
+#fixed

Thanks

PR #2309, opened 2015-02-10
Fixed: CPSegmentedControl divider wrong color

Dogild:
Previously, when clicking on segment, the divider did not have the expected color. Now they does.


primalmotion:
+#fixed

Thanks

PR #2311, opened 2015-02-11
Fixed: CPPopover position does not update each time

Dogild:
Previously, when the frame of a superview of the targetedView would update, the popover did not update its position.
Now it does by observing each frame of each superviews of the targetedView.


primalmotion:
Way better fix :)

+#fixed

PR #2315, opened 2015-02-20
New: Added method viewAtColumn:row:makeIfNecessary: in CPTableView

Dogild:
This PR adds the method viewAtColumn:row:makeIfNecessary: in CPTableView.
This method first attempts to return an available view, which is generally in the visible area. If there is no available view, and makeIfNecessary is YES, a prepared temporary view is returned. If makeIfNecessary is NO, and the view is not available, nil will be returned.
An exception will be thrown if row is an invalid row index and if column is an invalid column index.
The returned result should generally not be held onto for longer than the current run loop cycle. Instead they should re-query the table view for the row view.

UnitTests in Tests/AppKit/CPTableViewTests.j

### DIFF
--- a/AppKit/CPTableView.j
+++ b/AppKit/CPTableView.j
@@ -3403,7 +3403,6 @@ Your delegate can implement this method to avoid subclassing the tableview to ad
 */
 - (void)load
 {
-    console.error(@"load");
     if (_reloadAllRows)
     {
         [self _unloadDataViewsInRows:_exposedRows columns:_exposedColumns];
--- a/AppKit/CPTableView.j
+++ b/AppKit/CPTableView.j
@@ -1622,7 +1622,7 @@ NOT YET IMPLEMENTED
 */
 - (int)numberOfRows
 {
-        return _numberOfRows;
+    return _numberOfRows;
 }
 
 - (int)_numberOfRows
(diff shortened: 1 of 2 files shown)


AppKit/CPTableView.j
```diff
@@ -3706,6 +3706,64 @@ Your delegate can implement this method to avoid subclassing the tableview to ad
...
+        [CPException raise:CPInvalidArgumentException
+                    reason:@"Column " + column + " out of row range [0-" + (NUMBER_OF_COLUMNS ()- 1) + "] for rowViewAtRow:createIfNeeded:"];
+
+    var dataViewsForRow = _dataViewsForRows[row],
+        tableColumn = _tableColumns[column],
+        tableColumnUID = [tableColumn UID],
+        view = dataViewsForRow ? dataViewsForRow[tableColumnUID] : nil;
+
+    if (!makeIfNecessary)
+        return view || nil;
```
primalmotion: if !view, view is nil, so this seems to be useless, no?

Dogild: You get undefined, or we want to send back nil

primalmotion: just above it you have 

```
view = dataViewsForRow ? dataViewsForRow[tableColumnUID] : nil;
```

Dogild: dataViewsForRow can be defined but not dataViewsForRow[tableColumnUID]

primalmotion: valid

primalmotion:
+#fixed

Thanks

PR #2316, opened 2015-02-20
New: added the method setNeedsLayout:

Dogild:
This PR adds the possibility to layout or not a CPView.
Previously, once setNeedsLayout was called on a CPView, it wasn't possible to cancel the layout of the view.
Now we can as in cocoa. The method setNeedsLayout will still work (it calls the method setNeedsLayout: with YES).

UnitTests in Tests/AppKit/CPViewTest.j


primalmotion:
+#fixed

Thanks

PR #2318, opened 2015-02-24
FIXED: Exception when rows were removed in a table with binded columns.

### LINKED ISSUE #2317: Changing indexes of TableView suddenly gives uncaught CPRangeException
Today I updated my Cappuccino install. And suddenly the next error:

Frameworks/Debug/Foundation/CPException.j:51 Uncaught CPRangeException: -[_CPObservableArray objectAtIndex:]: index (1) beyond bounds (1)

When I change the items in a View-based CPTableView or when I reload the data. The error is not appearing on every item, but I can't find out any difference. The value of the table columns are bound to an arraycontroller, 

The project is pretty big, so hard to isolate the problem. It never happened before, so I think the problem is in the Cappuccino-code. Does it maybe has to do with the memory leak or enumeration of rows fixes?

I have the idea this error occurs when the items of the arraycontroller change. It really breaks the complete app.

I included the stack trace.

Any idea?

shot

Update:
I created a simple example: http://cl.ly/122v0M1O361U

cacaodev:
Before this fix, the CPTableColumn binder was not correctly reloading
the table when the number of rows changed. Now we reload fully the
dataviews when the number of rows changes. If no rows are
inserted/deleted, there is an optimization: we just need to reload the
objectValues and leave the dataviews untouched.

With Test in Tests/AppKitCPTableViewTest.j
Fixes #2317

### DIFF
--- a/AppKit/CPTableColumn.j
+++ b/AppKit/CPTableColumn.j
@@ -546,13 +546,22 @@ CPTableColumnUserResizingMask   = 1 << 1;
 - (void)setValueFor:(CPString)aBinding
 {
     var tableView = [_source tableView],
-        column = [[tableView tableColumns] indexOfObjectIdenticalTo:_source],
-        rowIndexes = [CPIndexSet indexSetWithIndexesInRange:CPMakeRange(0, [tableView numberOfRows])],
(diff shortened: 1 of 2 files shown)


primalmotion:
+#fixed

Thanks

PR #2319, opened 2015-02-25
Fixed: CPApplication dispatch the event when having the auto complete menu of a CPTokenField opened

Dogild:
Previously, the CPApplication dispatched the current event when having the auto complete menu of a CPTokenField opened.
This occurs weird behavior. For instance when hitting enter on the menu and having a default button, the action of the button was triggered.
Now, the CPTokenField handles the key enter when the autocomplete menu is opened.


primalmotion:
Travis seems to have some issues. The tests are actually all green, but Travis continues to wait for I don't know what

primalmotion:
+#fixed

Thanks

PR #2321, opened 2015-02-27
New: added the CPCache class

william57m:
This is an implementation of NSCache https://developer.apple.com/library/mac/documentation/Cocoa/Reference/NSCache_Class/index.html
I tried to reproduct the default behaviour, but I had to give up the implementation of NSDiscardableContent, because Cappuccino don't provide a mechanism to count objects usage, so the property evictsObjectsWithDiscardedContent also disappear from my implementation.

### DIFF
--- a/Foundation/CPCache.j
+++ b/Foundation/CPCache.j
@@ -22,39 +22,10 @@
 
 
 @import "CPObject.j"
+@import "CPDictionary.j"
+@import "CPString.j"
 
-/*
- * Class _CPCacheItem
- * Represent an item of CPCache
- * This class allow to associate a cost and a position to an object
- *
- * Attributes:
- * - object: the stored object
- * - cost: represent the cost (of memory) of the object
- * - position: represent the insertion order to determine the oldest object
- */
-@implementation _CPCacheItem : CPObject
-{
-    CPObject    _object      @accessors(property=object);
-    int         _cost        @accessors(property=cost);
-    int         _position    @accessors(property=position);
-}
-
-+ (id)cacheItemWithObject:(CPObject)anObject cost:(int)aCost position:(int)aPosition
-{
-    var cacheItem = [[super alloc] init];
-
-    if (cacheItem)
-    {
-        cacheItem._object = anObject;
-        cacheItem._cost = aCost;
-        cacheItem._position = aPosition;
-    }
-
-    return cacheItem;
-}
-
-@end
+@class _CPCacheItem;
 
 
 /*
@@ -71,8 +42,8 @@
 
 @end
 
+var CPCacheDelegate_cache_willEvictObject_ = 1 << 1;
 
-var CPCacheDelegate_cache_WillEvictObject = 1 << 1;
 
 /*!
     @class CPCache
@@ -91,17 +62,19 @@ var CPCacheDelegate_cache_WillEvictObject = 1 << 1;
  */
 @implementation CPCache : CPObject
 {
-    CPDictionary            _items              @accessors(readonly);  // TOFIX: delete accessors which is only needed for test
+    CPDictionary            _items;
     int                     _currentPosition;
-    int                     _countLimit;
-    int                     _totalCostLimit;
-    id <CPCacheDelegate>    _delegate;
-
-    // Set of bit to determine which delegate has to respond
+    BOOL                    _totalCostCache;
     unsigned                _implementedDelegateMethods;
+
+    CPString                _name               @accessors(property=name);
+    int                     _countLimit         @accessors(property=countLimit);
+    int                     _totalCostLimit     @accessors(property=totalCostLimit);
+    id <CPCacheDelegate>    _delegate           @accessors(property=delegate);
 }
 
 
+#pragma mark -
 #pragma mark Initialization
 
 /*!
@@ -110,12 +83,14 @@ var CPCacheDelegate_cache_WillEvictObject = 1 << 1;
 */
 - (id)init
 {
-    self = [super init];
-
-    if (self)
+    if (self = [super init])
     {
         _items = [[CPDictionary alloc] init];
         _currentPosition = 0;
+        _totalCostCache = -1;
+        _implementedDelegateMethods = 0;
+
+        _name = @"";
         _countLimit = 0;
         _totalCostLimit = 0;
         _delegate = nil;
@@ -125,6 +100,7 @@ var CPCacheDelegate_cache_WillEvictObject = 1 << 1;
 }
 
 
+#pragma mark -
 #pragma mark Managing cache
 
 /*!
@@ -162,6 +138,9 @@ var CPCacheDelegate_cache_WillEvictObject = 1 << 1;
     // Add object
     [_items setObject:[_CPCacheItem cacheItemWithObject:anObject cost:aCost position:++_currentPosition] forKey:aKey];
 
+    // Invalid cost cache
+    _totalCostCache = -1;
+
     // Clean cache to satisfy condition (< totalCostLimit & < countLimit) if necessary
     [self _cleanCache];
 }
@@ -173,10 +152,13 @@ var CPCacheDelegate_cache_WillEvictObject = 1 << 1;
 - (void)removeObjectForKey:(id)aKey
 {
     // Call delegate method to warn that the object is going to be removed
-    if (_implementedDelegateMethods & CPCacheDelegate_cache_WillEvictObject)
-        [_delegate cache:self willEvictObject:[[_items objectForKey:aKey] object]];
+    [self _sendDelegateWillEvictObjectForKey:aKey];
 
+    // Remove object
     [_items removeObjectForKey:aKey];
+
+    // Invalid cost cache
+    _totalCostCache = -1;
 }
 
 /*!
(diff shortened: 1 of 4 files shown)


primalmotion:
Thank you @william57m 

@cappuccino/core-devs please don't merge this in, this is an actual interview question :) I'll review it next week.

primalmotion:
Well, you have the week end to make your tests to pass, but looking at the error, should not be a big deal ;)

Foundation/CPCache.j
```diff
@@ -0,0 +1,350 @@
...
+ * Class _CPCacheItem
+ * Represent an item of CPCache
+ * This class allow to associate a cost and a position to an object
+ *
+ * Attributes:
+ * - object: the stored object
+ * - cost: represent the cost (of memory) of the object
+ * - position: represent the insertion order to determine the oldest object
+ */
+@implementation _CPCacheItem : CPObject
```
primalmotion: we usually put private classes under the main one.
You'll get a warning about an unknown class when compiling this file, but you can fix that by adding a `@class _CPCacheItem` right after the imports section

Foundation/CPCache.j
```diff
@@ -0,0 +1,350 @@
...
+{
+    CPObject    _object      @accessors(property=object);
+    int         _cost        @accessors(property=cost);
+    int         _position    @accessors(property=position);
+}
+
++ (id)cacheItemWithObject:(CPObject)anObject cost:(int)aCost position:(int)aPosition
+{
+    var cacheItem = [[super alloc] init];
```
primalmotion: this is working, but you're lucky _CPCacheItem doesn't do anything in it's init method.

should be `var cacheItem = [[self alloc] init]`

Foundation/CPCache.j
```diff
@@ -0,0 +1,350 @@
...
+
+    if (cacheItem)
+    {
+        cacheItem._object = anObject;
```
primalmotion: if you have accessors you should use them

`[cacheItem setObject:anObject] `

Foundation/CPCache.j
```diff
@@ -0,0 +1,350 @@
...
+ */
+@protocol CPCacheDelegate <CPObject>
+
+@optional
+- (void)cache:(CPCache)cache willEvictObject:(id)obj;
+
+@end
+
+
+var CPCacheDelegate_cache_WillEvictObject = 1 << 1;
```
primalmotion: style:

Should be `var  CPCacheDelegate_cache_willEvictObject_` (replace all `:` with `_`)

Foundation/CPCache.j
```diff
@@ -0,0 +1,350 @@
...
+    similar to the CPDictionary class. Developers often incorporate caches to temporarily
+    store objects with transient data that are expensive to create.
+
+    Reusing these objects can provide performance benefits, because their values do not have to be recalculated.
+    However, the objects are not critical to the application and can be discarded if memory is tight.
+    If discarded, their values will have to be recomputed again when needed.
+ */
+@implementation CPCache : CPObject
+{
+    CPDictionary            _items              @accessors(readonly);  // TOFIX: delete accessors which is only needed for test
```
primalmotion: if it's only needed for the tests, directly access it, don't pollute the main class:

in your test:

```objj
var cache = [CPCache new];
// [...]
var items = cache. _items;
```

Foundation/CPCache.j
```diff
@@ -0,0 +1,350 @@
...
+
+/*!
+    Initializes the cache with default values
+    @return the initialized cache
+*/
+- (id)init
+{
+    self = [super init];
+
+    if (self)
```
primalmotion: style.

We usually do 

```
- (id)init
{
    if (self = [super init])
    {
        // init
    }

    return self;
}
```

I guess that's the only place assignation in a `if` should be tolerated in the world :)

Foundation/CPCache.j
```diff
@@ -0,0 +1,350 @@
...
+}
+
+/*!
+    Removes the object from the cache for the given key.
+    @param aKey the key of the object to be removed
+*/
+- (void)removeObjectForKey:(id)aKey
+{
+    // Call delegate method to warn that the object is going to be removed
+    if (_implementedDelegateMethods & CPCacheDelegate_cache_WillEvictObject)
```
primalmotion: we normally create private methods to send call to delegates.

for instance:

```objj
- (void)_sendDelegateWillEvictObjectForKey:(CSPtring)aKey
{
    if (_implementedDelegateMethods & CPCacheDelegate_cache_WillEvictObject)
        [_delegate cache:self willEvictObject:[[_items objectForKey:aKey] object]];
}
```

then 

```objj
- (void)removeObjectForKey:(id)aKey
{
    [self _sendDelegateWillEvictObjectForKey:aKey];
    [_items removeObjectForKey:aKey];
}
```

Foundation/CPCache.j
```diff
@@ -0,0 +1,350 @@
...
+    [_items removeAllObjects];
+}
+
+
+#pragma mark Accessors
+
+/*!
+    Returns the count limit of the cache
+*/
+- (int)countLimit
```
primalmotion: you could use for instance `@accessors(property=countLimit)` in the class ivar list instead of adding all these methods manually.

You can also declare a delegate and override the setter or getter if you need to, that way you get code for free, and you can add your own only where needed

Foundation/CPCache.j
```diff
@@ -0,0 +1,350 @@
...
+ */
+- (int)_count
+{
+    return [_items count];
+}
+
+/*
+ * This method return the total cost (addition of all object's cost in the cache)
+ */
+- (int)_totalCost
```
primalmotion: what about some caching here?

- You have a variable `_totalCountCacheValid`
- in `- (int)_totalCost`
    - if `! _totalCountCacheValid`, you compute the value, cache it and return it
    - if  `_totalCountCacheValid`, you simply return the cached value
- when you add or remove an item, you simply mark the cache as invalid

Foundation/CPCache.j
```diff
@@ -0,0 +1,350 @@
...
+}
+
+/*
+ * This method clean the cache if the totalCost or the count exceeds the totalCostLimit or the countLimit
+ * until to satisfy conditions (totalCost < totalCostLimit and count < countLimit)
+ */
+- (void)_cleanCache
+{
+    // Check if the condition is satisfied (totalCost < totalCostLimit and count < countLimit)
+    if (([self _totalCost] > _totalCostLimit && _totalCostLimit > 0) || ([self _count] > _countLimit && _countLimit > 0))
```
primalmotion: we usually filter out this conditions that are "if that don't do anything"

you should do a reverse if, and if it matches, you simply `return`

Foundation/CPCache.j
```diff
@@ -0,0 +1,350 @@
...
+    However, the objects are not critical to the application and can be discarded if memory is tight.
+    If discarded, their values will have to be recomputed again when needed.
+ */
+@implementation CPCache : CPObject
+{
+    CPDictionary            _items              @accessors(readonly);  // TOFIX: delete accessors which is only needed for test
+    int                     _currentPosition;
+    int                     _countLimit;
+    int                     _totalCostLimit;
+    id <CPCacheDelegate>    _delegate;
```
primalmotion: NSCache references states a cache can have a `name`, seems to be missing here

primalmotion:
+#needs-improvement

Thanks!

There a few issues here an there, but it's pretty good. Please don't rebase your commit, simply push your new patches to your branch. The PR will be automatically updated in Github.

Also please ensure running `objj` on your file doesn't generate any warning (hint: it is :)), and also run `capp_lint` for style checking

william57m:
Thank's for the tips, I'm gonna to correct that.

Dogild:
Looks good to me !

cacaodev:
This is cool and I think there is a great potential here. Of course an important feature cannot implemented in javascript: the ability to automatically evict objects in response to a "memory pressure". Even if it's undocumented, form what i've read i believe this is what cocoa does.
One question: why do you say you can't implement the CPDiscardableContent protocol ?. Even without the same memory management objc has, i believe it is possible. -beginContentAccess just means "i'm dealing with this object, please don't remove it now" and - discardContentIfPossible means "remove this object (or not) if we are not accessing it".

william57m:
Ok I also thought, but the -discardContentIfPossible method doesn't return a boolean value, to say to the cache : "ok you can discard me or not".
So in Objective-C I presume that the NSCache check the counter usage in the case where it's a "Discardable object".
So we have no way to know if an object has to be retained or not, unless to reinvent the protocol or to dig more depth in Cappuccino.

For the rest, the method -isContentDiscarded could be useful for the cache to remove object which is not used anymore, but I didn't want to implement a class if we only can exploit a half.

cacaodev:
You're right, discardContentIfPossible  gives an opportunity to the object to remove its content and free up some memory, it is not a condition for object removal itself. The good news is that we can also do that in javascript.
We don't know what is the counter usage but we can know that the object is in use thanks to -beginContentAccess and just disallow removal/discard for this object.
So, to conclude, I think the CPDiscardableContent can be adapted and stay consistent.

Dogild:
@cacaodev we know when we are using the object with `beginContentAccess` and when retrieving the object. But how can we know we don't use it anymore ?

william57m:
@cacaodev sorry but I don't see what you mean, the beginContentAccess method is used by the object to increment its counter and to avoid to be discarded, so we cannot use this method from the cache to check if we can evict an object or not. But actually it's not what the cache must does.

Apple provides a class "NSPurgeableData" which implement the NSDiscardableContent protocol, I took a look at the interface file and I saw a counter in the list of attributes... so you can implement the protocol like you want.

After some tests in xCode with the original class, the behaviour is not what we (or I) expected... even if you use a NSDiscardableContent object, it may be removed from the cache whatever the objects in the cache. Because the role of the cache is not to retain object.

And according to the Apple's documentation it makes sense

"Storing this type of object in a cache has benefits, because its content can be discarded when it is not needed anymore" -> Yes but it's the role of the user to manage its object (from the method "discardContentIfPossible", and destruct it), it's not the role of the cache.

"By default, NSDiscardableContent objects in the cache are automatically removed from the cache if their content is discarded, although this automatic removal policy can be changed." -> When the cache is used, it checks if there are discarded objects and it removed them (according to "this automatic removal policy" which is also a mystery...), but there is no way to retain an object in the cache.

So to conclude, yes I can implement it.

primalmotion:
I think we can implement memory management later.

This PR looks good to me.

primalmotion:
+#fixed

Thanks!

PR #2323, opened 2015-03-02
Fixed: Method scrollRectToVisible:(CGRect)aRect in CPView didn’t work...

mrcarlberg:
... if aRect is larger then the visible rect.

If aRect is larger then the visible rect this method scrolled all they way to the opposite edge instead of the nearest.

I have also added test case for this.

I made a test application in Cappuccino that can be tested here: http://mini.carlberg.org/dev/ScrollRectToVisibleWithToLargeViewCappTest/

A Cocoa version of the same application can be dowloaded here: http://mini.carlberg.org/dev/ScrollRectToVisibleWithToLargeViewTest.tgz

### DIFF
--- a/AppKit/CPView.j
+++ b/AppKit/CPView.j
@@ -2723,18 +2723,47 @@ setBoundsOrigin:
     if (CGRectContainsRect(documentViewVisibleRect, rectInDocumentView))
         return NO;
 
-    var scrollPoint = CGPointMakeCopy(documentViewVisibleRect.origin);
+    var currentScrollPoint = documentViewVisibleRect.origin,
+        scrollPoint = CGPointMakeCopy(currentScrollPoint),
+        rectInDocumentViewMinX = CGRectGetMinX(rectInDocumentView),
+        documentViewVisibleRectMinX = CGRectGetMinX(documentViewVisibleRect),
+        doesItFitForWidth = documentViewVisibleRect.size.width >= rectInDocumentView.size.width;
 
     // One of the following has to be true since our current visible rect didn't contain aRect.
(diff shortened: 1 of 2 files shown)


AppKit/CPView.j
```diff
@@ -2723,18 +2723,47 @@ setBoundsOrigin:
...
-    if (CGRectGetMinX(rectInDocumentView) < CGRectGetMinX(documentViewVisibleRect))
-        scrollPoint.x = CGRectGetMinX(rectInDocumentView);
-    else if (CGRectGetMaxX(rectInDocumentView) > CGRectGetMaxX(documentViewVisibleRect))
-        scrollPoint.x += CGRectGetMaxX(rectInDocumentView) - CGRectGetMaxX(documentViewVisibleRect);
-
-    if (CGRectGetMinY(rectInDocumentView) < CGRectGetMinY(documentViewVisibleRect))
-        scrollPoint.y = CGRectGetMinY(rectInDocumentView);
-    else if (CGRectGetMaxY(rectInDocumentView) > CGRectGetMaxY(documentViewVisibleRect))
-        scrollPoint.y += CGRectGetMaxY(rectInDocumentView) - CGRectGetMaxY(documentViewVisibleRect);
+    if (rectInDocumentViewMinX < rectInDocumentViewMinX && doesItFitForWidth)
```
Dogild: Not sure this is useful if (rectInDocumentViewMinX < rectInDocumentViewMinX) !

mrcarlberg:
Good catch!

I have now fixed this and updated the test case to catch it too.

Thanks!

Dogild:
Thanks ! I need it right now :)

PR #2324, opened 2015-03-02
CPSegmentedControl fixes

cacaodev:
New: insertSegments:AtIndexes: removeSegmentsAtIndexes: private methods necessary to implement CPTabView -insertTabViewItem:atIndex:
Reworked -tileWithChangedSegment: it now tiles all segments views at once when multiple segments are inserted.
The selection now correctly updates when the selected segment is removed.
Fixed widthForSegment: and frameForSegment: The width property of a segment was updated based on its current frame.

ojtest and manual test Manual/CPSegmentedControlTest/


AppKit/CPSegmentedControl.j
```diff
@@ -677,63 +707,39 @@ CPSegmentSwitchTrackingMomentary = 2;
 {
 }
 
-- (void)tileWithChangedSegment:(unsigned)aSegment
```
Dogild: This method is public right now. It's a method which not exist in the AppKit or in the UIKit.

Should we pass it in private even if it's was public for a while ?

cacaodev:
It seems to me it should be private because it's only used by other public methods managing the segment width like setLabel:forSegment: or setSegmentCount:. But maybe there are some other reasons  i don't know explaining why it was made public in the first place ...

AppKit/CPSegmentedControl.j
```diff
@@ -677,63 +707,39 @@ CPSegmentSwitchTrackingMomentary = 2;
...
-    {
-        [_segments[index] frame].origin.x += delta;
+    [self setNeedsLayout];
+    [self setNeedsDisplay:YES];
+}
 
-        [self drawSegmentBezel:index highlight:NO];
-        [self drawSegment:index highlight:NO];
-    }
+- (CGSize)intrinsicContentSize
```
Dogild: The method could be private

Dogild:
Yep...I don't know why I should use that personally...

Other remarks, it could be great to have some documentations/comment for the new methods. Specially because they don't exist in Cocoa.

Dogild:
We have a difference between the current version and yours. I'll take a look in the next days to see of you or we are wrong :)
 
screen shot 2015-03-05 at 6 08 38 pm

AppKit/CPSegmentedControl.j
cacaodev: It's public in cocoa but since this is the only occurence in AppKit and it won't be documented, I will privatize it. Although i think all controls should implement this method because it gives a better separation between metrics and actual layout.

cacaodev:
@Dogild (set | insert | remove)segments: methods are private. I didn't underscore them because they may be used in the future as indexed accessors. I can add a code comment about that if you want.

AppKit/CPSegmentedControl.j
Dogild: I did not know this method. Let it public ! You are right

aparajita: Better to make it public now and worry about documenting it later.

aparajita:
Private methods don't have to be underscored. We can hide them from the documentation using a doxygen directive.

Dogild:
The difference came from our side (@primalmotion  does not now how to use a `CPSegmentedControl` :D). Everything looks good for us !

Dogild:
`- (void)tileWithChangedSegment:(unsigned)aSegment`, maybe we should create a method `-(void)tile` which call `[self tileWithChangedSegment:0]` ?

cacaodev:
There is something interesting about CPSegmentedControl: the frameSize is completely determined by its content and the content only. Meaning it's not externally resizable. Maybe we should take this brute fact into account this and disable -setFrameSize: with a noop. This would optimize any external attempt to resize the control, including the autosize mechanism. When the content changes, we could just update the frameSize property. We need to check first if it doesn't break anything, that is why I didn't make the change for now.

Dogild:
Yep, let's do that in another PR/commit.

Issue #362, opened 2009-12-08
[OBJJ] Parser doesn't detect double instance variable declaration


p3consulting:

mrcarlberg:
+#fixed
by #2326

cappbot:
**Milestone:** 1.0.  **Labels:** #fixed, Objective-J, bug.  **What's next?** This issue is considered successfully resolved.

PR #2326, opened 2015-03-03
Duplicate ivar on superclass

### LINKED ISSUE #2271: About Private Instance Variables in the Objective-J
```

@import <AppKit/AppKit.j>



@implementation A: CPObject
{
    int _i;
}

- (id)init
{
    if (!(self = [super init]))
        return nil;

    _i = 7;

    return self;
}

- (void)printA
{
    CPLog.info("i = " + _i);
}

@end

@implementation B: A
{
    int _i;
}

- (id)init
{
    if (!(self = [super init]))
        return nil;

    _i = 8;

    return self;
}

- (void)printB
{
    CPLog.info("i = " + _i);
}

@end

```

When executing the below code:
```
var a = [[A alloc] init];
[a printA];
var b = [[B alloc] init];
[b printA];
[b printB];
```
it will print:
```
2014-12-04 16:16:14.561 Cappuccino [info]: i = 7
2014-12-04 16:16:14.844 Cappuccino [info]: i = 8
2014-12-04 16:16:14.845 Cappuccino [info]: i = 8
```

I can't make sure this is a bug, but if it's not, that means we must take a look at both the interface and implementation of the superclass when inheriting from it. At least I want to inherit from a class only by its interface showed in the API docs.

The corresponding C++ code:
```

#include <iostream>
#include <string>
#include <vector>
#include <map>
#include <algorithm>



using namespace std;

class A {
public:
        A(void)
        {
                _i = 7;
        }

        void printA(void)
        {
                cout << "i = " << _i << endl;
        }

protected:
        int _i;
};

class B: public A {
public:
        B(void): A()
        {
                _i = 8;
        }

        void printB(void)
        {
                cout << "i = " << _i << endl;
        }

        void printBA(void)
        {
                cout << "i = " << A::_i << endl;
        }

protected:
        int _i;
};

int main(int argc, char *argv[])
{
        A a;
        B b;

        a.printA();
        b.printA();
        b.printB();
        b.printBA();
        return 0;
}

```

It will print:
```
i = 7
i = 7
i = 8
i = 7
```

Help me.

mrcarlberg:
Today you can declare a duplicate ivar in a subclass. Now the compiler will generate an error.

This fixes #2271 and #362

### DIFF
--- a/AppKit/CPColorPanel.j
+++ b/AppKit/CPColorPanel.j
@@ -71,7 +71,6 @@ var SharedColorPanel = nil,
 */
 @implementation CPColorPanel : CPPanel
 {
-    _CPColorPanelToolbar    _toolbar;
     _CPColorPanelSwatches   _swatchView;
     _CPColorPanelPreview    _previewView;
 
--- a/AppKit/CPMenu/_CPMenuBarWindow.j
+++ b/AppKit/CPMenu/_CPMenuBarWindow.j
@@ -32,7 +32,6 @@
 
 @implementation _CPMenuBarWindow : CPPanel
 {
-    CPMenu      _menu;
     CPView      _highlightView;
     CPArray     _menuItemViews;
 
--- a/AppKit/CPRuleEditor/_CPRuleEditorViewSlice.j
+++ b/AppKit/CPRuleEditor/_CPRuleEditorViewSlice.j
@@ -29,7 +29,6 @@
     CGRect       _animationTargetRect   @accessors(property=_animationTargetRect);
     BOOL         _selected              @accessors(getter=_isSelected, setter=_setSelected:);
     BOOL         _lastSelected          @accessors(getter=_isLastSelected, setter=_setLastSelected:);
-    CPColor      _backgroundColor       @accessors(property=backgroundColor);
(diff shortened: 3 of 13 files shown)


Dogild:
Hi Martin,

In Cocoa we can't duplicate ivars but we can duplicate something like `@property NSString *firstName;`. I have no idea where it could be useful though.

mrcarlberg:
It can be useful in Objective-C if you have a immutable class and want to do a mutable subclass. Then you can change a property from ```readonly``` to ```readwrite```.

In Objective-J we don't have properties as it is hard to implement as a superset on top of Javascript so this does not apply to us in the same way. We have ```@accessors``` that create getters and setters. It might be cool to be able to generate a setter in a subclass for an ivar that is declared in a superclass.

Although this is a different feature then what this pull request is trying to solve...

Dogild:
Thanks ! So it won't be possible to declare (in a subclass) a same ivar with others/new `@accessors`.

mrcarlberg:
No, you have to declare the setter yourself as a regular method

Dogild:
Tested with our application and it works !

Dogild:
I'll let you merge it ;)

mrcarlberg:
+#fixed

Thanks!

Issue #1169, opened 2011-03-14
CPTableView does not update _clickedRow


BlairDuncan:
Currently the only time the _clickedRow is updated is when the user doubleClicks. Thus you can single click all over the place and it will always return the last double clicked row and not the row you just clicked on. 
What I believe should happen is on mouse down it should be set to -1 (for drag support) and upon mouse up set to the clicked row.

daboe01:
this issue is invalid as discussed here: https://github.com/cappuccino/cappuccino/pull/2328#issuecomment-78207574
+#wont-fix

cappbot:
**Milestone:** Someday.  **Label:** #wont-fix.  **What's next?** A reviewer or core team member has decided against acting upon this issue.

cappbot:
**Milestone:** Someday.  **Label:** #wont-fix.  **What's next?** A reviewer or core team member has decided against acting upon this issue.

PR #2328, opened 2015-03-08
_clickedRow and _clickedColumn are only updated after double clicking

daboe01:
this PR makes both ivars update after every click instead of only after double-clicking

fixes issue #1169

### DIFF
--- a/AppKit/CPTableView.j
+++ b/AppKit/CPTableView.j
@@ -4664,13 +4664,12 @@ Your delegate can implement this method to avoid subclassing the tableview to ad
 
     } //end of editing conditional
 
+    _clickedRow = [self rowAtPoint:aPoint];
+    _clickedColumn = [self columnAtPoint:lastPoint];
+
     //double click actions
     if ([[CPApp currentEvent] clickCount] === 2 && _doubleAction)
-    {
-        _clickedRow = [self rowAtPoint:aPoint];
-        _clickedColumn = [self columnAtPoint:lastPoint];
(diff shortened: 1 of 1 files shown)


Dogild:
Hi Daniel,

According to the documentation : The value of this property is meaningful in the target object’s implementation of the action and double-action methods.

This both values are only updated in the "scope" of the `action` and `doubleAction` method of the tableView.

For instance, in cocoa

```
- (IBAction)actionButton:(id)sender
{
    NSLog(@"%ld", (long)[_t clickedRow]);
}

- (IBAction)actionTableView:(id)sender
{
    NSLog(@"%ld", (long)[_t clickedRow]);
}
```
The log in `actionTableView` will print the expected clickedRow but in `actionButton` wi will always get -1

daboe01:
@Dogild the current implementation behaves the way you describe.
Is issue #1169 invalid then?

Dogild:
Yep the issue #1169 is invalid.

I just took a look in cocoa (again), the `clickedRow and `clickedColumn` properties can be only used in the scope of a mouse event. In the tableView, the clickedRow and clickedColumn will be set to the appropriate value from the method `- (BOOL)startTrackingAt:(CGPoint)aPoint` to `- (void)stopTracking:(CGPoint)lastPoint at:(CGPoint)aPoint mouseIsUp:(BOOL)mouseIsUp`. Then it will be set to -1.

So every delegate methods called in these methods could use `clickedRow` and `clickedColumn`, but after `stopTracking` they (the properties) should be set to -1 again.

Issue #10, opened 2009-05-03
-hash is broken on CPArray


cappuccino:
It breaks the contract that two arrays which are -isEqual: also need to report the same -hash value.

Patch and tests included

It&rsquo;s also the same way that GnuStep and Cocotron implement this - I was quite surprised how simple they implemented it.

:)

[original LH ticket](http://cappuccino.lighthouseapp.com/projects/16499/tickets/286)

 This ticket has 1 attachment(s).

daboe01:
given that a lot of code has been written since 2009,  i would suggest closing this issue and documenting the current behaviour.

@aparajita, any objections?

aparajita:
@daboe01 Go ahead and close it, but document the problem with the current implementation compared to Cocoa.

daboe01:
#fixed

cappbot:
**Assignee:** [boucher](https://github.com/boucher).  **Milestone:** 1.0.  **Labels:** #fixed, Foundation, bug.  **What's next?** This issue is considered successfully resolved.

cappbot:
**Assignee:** [boucher](https://github.com/boucher).  **Milestone:** 1.0.  **Labels:** #fixed, Foundation, bug.  **What's next?** This issue is considered successfully resolved.

PR #2329, opened 2015-03-08
Added documentation for CPArray's hash method

daboe01:
The implementation of CPArray's hash method deliberately differs from cocoa.
This PR adds the appropriate documentation.

Fixes issue #10

### DIFF
--- a/Foundation/CPArray/_CPArray.j
+++ b/Foundation/CPArray/_CPArray.j
@@ -209,7 +209,7 @@ var concat = Array.prototype.concat,
 }
 
 /*!
-    Returns a hash for the object. Other than in Cocoa, the hash value does not take the content into account. To distinct instances of CPArray with identical content (that are \c isEqual:) may return different \c -hash values.
+Returns a hash for the object. Unlike Cocoa, the hash value does not take content into account, so two arrays with the same content (\c isEqual: === YES) will not generate the same hash.
 */
 - (unsigned)hash
 {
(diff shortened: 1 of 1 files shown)


Foundation/CPArray/_CPArray.j
```diff
@@ -209,6 +209,14 @@ var concat = Array.prototype.concat,
 }
 
 /*!
+    Returns a hash for the object. Other than in Cocoa, the hash value does not take the content into account. To distinct instances of CPArray with identical content (that are \c isEqual:) may return different \c -hash values.
```
aparajita: Please change to this:

```
Returns a hash for the object. Unlike Cocoa, the hash value does not take content into account,
so two arrays with the same content (\c isEqual: === YES) will not generate the same hash.
```

Dogild:
@aparajita What is the problem to return the count of the array as in cocoa ? When we call the method isEqual on a CPArray we won't compare the UID of the array anyway.

aparajita:
@Dogild See the [original discussion](https://github.com/cappuccino/cappuccino/issues/10#issuecomment-5021193). Even if we change the implementation without breaking anything in Cappuccino (which will take a lot of work and will most likely fail on our first few attempts), we would have to warn all users that CPArray -hash does NOT create a unique value and thus may break their code.

I don't see how fixing this one particular "bug" (if being different than Cocoa is a bug) is worth it.

Dogild:
Not sure if a lot of people use the `hash` method but it's not going to make a huge difference to keep this thing.

Dogild:
Thanks !

Issue #1686, opened 2012-11-09
programmatic use of CAArrayController w/ TableView bindings broken


core-code:
it seems to me that binding a CAArrayController to a CPTableView only works with a XIB file and not programmatically.

consider this app (not written by me): http://pastebin.com/dweztETS

running it you don't see the intended output "Foo One" etc in the table, but rather debug information about KVO objects. i can't see anything wrong in the code (and writing similar code myself has yielded the same problem) so there must be something wrong in cappucino. binding tables with XIB files doesn't seem to exhibit the issue.

daboe01:
given our limited ressources, i think we are better of with documenting the current behavior instead of patching.
i would advocate closing this one.
what do you core devs think?

daboe01:
@aljungberg any objections against closing this one?

aparajita:
We can close it, but only after the current behavior has been documented.

daboe01:
-#needs-patch
-bug
+documentation

cappbot:
**Milestone:** Someday.  **Labels:** #accepted, documentation.  **What's next?** A reviewer should examine this issue.

daboe01:
+fixed

daboe01:
+#fixed

cappbot:
**Milestone:** Someday.  **Labels:** #fixed, documentation.  **What's next?** This issue is considered successfully resolved.

cappbot:
**Milestone:** Someday.  **Labels:** #fixed, documentation.  **What's next?** This issue is considered successfully resolved.

PR #2330, opened 2015-03-08
Clarification for CPTableColumn's binding documentation

daboe01:
This PR updates the documentation to clarify that bindings work only after a column has been added to a tableview.

fixes issue #1686

### DIFF
--- a/AppKit/CPTableColumn.j
+++ b/AppKit/CPTableColumn.j
@@ -601,7 +601,7 @@ CPTableColumnUserResizingMask   = 1 << 1;
 }
 
 /*!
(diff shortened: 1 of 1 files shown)


AppKit/CPTableColumn.j
```diff
@@ -601,7 +601,7 @@ CPTableColumnUserResizingMask   = 1 << 1;
 }
 
 /*!
-    Binds the receiver to an object.
+    Binds the receiver to an object. Note that other than in Cocoa, this will only work when the receiver has been added to a \c CPTableView.
```
aparajita: Change to:

```
Binds the receiver to an object. Note that unlike Cocoa,
this works only *after* the receiver has been added to
a \c CPTableView.
```

aparajita:
Merged, thanks!

PR #2334, opened 2015-03-14
Fixed: reloadData in CPTableView run the runLoop

Dogild:
Previously, when reloading a CPTableView the run loop was explicitly call to layout the tableView. This is not the case in Cocoa.
You can call several times the method reloadData and this will only lay out the tableView one time.

### DIFF
--- a/AppKit/CPTableView.j
+++ b/AppKit/CPTableView.j
@@ -1145,6 +1145,7 @@ NOT YET IMPLEMENTED
         _dirtyTableColumnRangeIndex = MIN(index, _dirtyTableColumnRangeIndex);
 
     [self reloadData];
+    [[CPRunLoop currentRunLoop] limitDateForMode:CPDefaultRunLoopMode];
 }
 
 /*!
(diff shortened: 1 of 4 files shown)


Dogild:
@cacaodev Any reason you add this line of code ?

cacaodev:
I added this line because I thought it was the right behavior from the developer point of view but if you think it has to be removed, then let's do it. If you do this, there are 3 others places where I believe we should perform the run loop: CPTableView-removeTableColumn, CPTableView-enumerateAvailableRows and CPTableColumnBinder-setValueFor:.
We also need to keep in mind that -reloadData can change the numberOfRows.
I don't think we can really compare -reloadData to cocoa because they are not loading/unloading views in -layout like we do, it's more about was is needed and consistent for us.

Dogild:
Ok, thanks for the info. I'll add the the code to perform the run loop where it's needed.

I believe it's better like this, in our application we sometimes call several times the method reloadData in the same stack, for a question of performance it's better like this.

Dogild:
@cacaodev can you take a look and let me know if it's good ?

I really believe that the developer should not perform the run loop to get basics informations from the tableView and that the tableView perform itself the run loop when it's only necessary.

primalmotion:
+#fixed
Thanks

PR #2335, opened 2015-03-18
Fixed: CPSearchField alignment from nib2cib

t00f:
CPSearchField were not properly aligned in Aristo2 theme.
This fix uses the nib2cib-adjustment-frame information to align it.
It also improve the Nib2CibAlignement test.


Dogild:
What about Aristo 1 ?

Dogild:
I just tested with Aristo and Aristo2, issues are the following :

Aristo:
- The CPSearchField is to large

Aristo2:
- The height of the control is different between normal and focused mode

Dogild:
Hi @t00f, in Aristo 1 the content is jumping when focusing and leaving the CPSearchField.

goodbehavior4

Dogild:
Hi Chris,

Last small things and we are good ! In Aristo 1, the normal mode, the margin between the content and the search icon seems to big.

screen shot 2015-03-22 at 5 07 33 pm

Dogild:
Thanks

PR #2336, opened 2015-03-20
Fixed: Encoding a [CPDecimalNumber zero] causes a corrupt CPDecimalNumbe...

### LINKED ISSUE #2332: Encoding a [CPDecimalNumber zero] causes a corrupt CPDecimalNumber after decoding
Encoding a CPDecimalNumber zero causes an error on decoding. The CPDecimalNumber object will refer to itself in the _exponent and _mantisa properties. This will lead to an infinite loop when one of this properties is accessed. Here is my reduction of the problem:

```Objective-J
	var aDecimalNumber = [CPDecimalNumber zero];	
	var encodedDecimalNumber 	= [CPKeyedArchiver archivedDataWithRootObject:aNumber];	
	var data 					= [CPData dataWithRawString:[encodedNumber rawString]];
	var unarchiver				= [[CPKeyedUnarchiver alloc] initForReadingWithData:data];
	var decodedDecimalNumber 	= [unarchiver decodeObjectForKey:@"root"];
```

I already found the cause of this problem. In the function _CPKeyedArchiverEncodeObject of the class CPKeyedArchiver following line

```Objective-J
    // Get the proper replacement object
    var GUID = [anObject UID],
        object = [self._replacementObjects objectForKey:GUID];
```

will get a GUID of an object an returns a replacementObject if this GUID is already in the Dictonary. Everything is okay so far. The problem comes with the implemented method UID on CPNumber.

 ```Objective-J
- (CPString)UID
{
    var UID = CPNumberUIDs.valueForKey(self);

    if (!UID)
    {
        UID = objj_generateObjectUID();
        CPNumberUIDs.setValueForKey(self, UID);
    }

    return UID + "";
}
```

When this method is called on [CPDecimalNumber zero] for the first time, a new UID will be created, because the CPNumberUIDs-Dictonary is empty. This UID (for example 1234) will be added to the CPNumberUIDs-Dictonary an returned to _CPKeyedArchiverEncodeObject . The next step is to encode the exponent of the DecimalNumber which is a CPNumber with the value of 0. The method UID is also called on this CPNumber. You would expect that a new UID will be generated because this CPNumber cant be found in the  CPNumberUIDs-Dictonary. But that's not the case. It returns the well known CPDecimalNumber with the UID 1234. Just because CPNumber with the value 0 is also the Key in the CPNumberUIDs-Dictionary for the CPDecimalNumber. That's why the the CPDecimalNumber will later refer to itself.

My workaround for this problem is a Category for CPDecimalNumber:

```Objective-J
var CPDecimalNumberUIDs    = new CFMutableDictionary();

@implementation CPDecimalNumber (EncodeFix)
{}

- (CPString)UID
{
    var UID = CPDecimalNumberUIDs.valueForKey(self);

    if (!UID)
    {
        UID = objj_generateObjectUID();
        CPDecimalNumberUIDs.setValueForKey(self, UID);
    }

    return UID + "";
}
```
 
CPDecimalNumbers now have their own UID-Dictionary what solves the problem (hopefully side-effect free). Maybe it makes sense to implement -(CPString)UID on CPDecimalNumber with the code above. I am new to github and have no clue how to contribute, so I would  appreciate if someone else could do this ;) Maybe there are better solutions for this interesting problem :)

cpasslack:
...r after decoding

Previously, CPNumber and CPDecimalNumber shared the same method UID and CPNumberUIDs-Dictionary.
This leads to some unexpected errors for example a [CPDecimalNumber zero] which refers to itself after
decoding. CPDecimalNumber now overwrites the method UID and has its own UID-Dictionary.

Fixes #2332

### DIFF
--- a/Foundation/CPDecimalNumber.j
+++ b/Foundation/CPDecimalNumber.j
@@ -28,7 +28,7 @@
 // The default global behavior class, created lazily
 var CPDefaultDcmHandler = nil;
 
-var CPDecimalNumberUIDs	= new CFMutableDictionary();
+var CPDecimalNumberUIDs = new CFMutableDictionary();
 
 /*!
     @class CPDecimalNumberHandler
--- a/Foundation/CPDecimalNumber.j
+++ b/Foundation/CPDecimalNumber.j
@@ -28,6 +28,8 @@
 // The default global behavior class, created lazily
 var CPDefaultDcmHandler = nil;
 
+var CPDecimalNumberUIDs	= new CFMutableDictionary();
+
 /*!
     @class CPDecimalNumberHandler
     @ingroup foundation
@@ -531,6 +533,20 @@ var CPDecimalNumberHandlerRoundingModeKey       = @"CPDecimalNumberHandlerRoundi
 }
 
 // instance methods
+
+- (CPString)UID
+{
+    var UID = CPDecimalNumberUIDs.valueForKey(self);
+
+    if (!UID)
+    {
+        UID = objj_generateObjectUID();
+        CPDecimalNumberUIDs.setValueForKey(self, UID);
+    }
+
+    return UID + "";
+}
+
 /*!
     Returns a new CPDecimalNumber object with the result of the summation of
     the receiver object and \c decimalNumber. If overflow occurs then the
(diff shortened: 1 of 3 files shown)


cpasslack:
CPNumberFormatterTest failed because:

````Objective-J
    [self assert:10 equals:objectValue];
````

objectValue is a CPDecimalNumber with a different UID --> isEqual returns "false"

Possible solutions:

change objectValue to [objectValue intValue]

or let the NumberFormater generate CPNumbers

[numberFormatter setGeneratesDecimalNumbers:NO]

primalmotion:
+#fixed

Thanks!

CyrilPeponnet:
This PR seems to break a lot of things...

PR #2337, opened 2015-03-20
FIXED: CPSegmentedControl -setSelectedSegment:-1

cacaodev:
previously, -setSelectedSegment:-1 was not visually deselecting the
segments and selectedSegment was not set to -1.

With tests.


Dogild:
Looks go to me

Issue #2341, opened 2015-04-06
0.9.8 CPSegmentedControl wrong vertical positioning...


didierkorthoudt:
Hi Folks !

I've just found a problem with CPSegmentedControl vertical positioning...

You can reproduce this very easily with a new project and adding a CPSegmentedControl vertically aligned with, say, a text field.

For what I've seen, the offset is (is this a symptom or just a coincidence ?) 24 pixels, which is the height of the CPSegmentedControl...

Here is the IB layout : 

![capture d ecran 2015-04-06 a 19 57 08](https://cloud.githubusercontent.com/assets/2394110/7009034/ca3f7792-dc97-11e4-81c6-cc8d7e0ae282.png)

and this is the result : 

![capture d ecran 2015-04-06 a 19 56 56](https://cloud.githubusercontent.com/assets/2394110/7009037/d56bade8-dc97-11e4-9209-4cadc74f72de.png)

Hope this can help,

Didier.

didierkorthoudt:
Forgot to mention that everything was OK in 0.9.7-1...

Maybe a XCodeCapp issue ?

Dogild:
Hi Didier, 

Can you share the app please ? 

Thanks

Dogild:
No this is perfect ! Easy app to reproduce the problem.

@cacaodev can you take a look please ? It looks like this commit a9a6e58d60db02413168b39a42857c14f0e19969 is the issue

cacaodev:
@didierkorthoudt does it work for you ?

didierkorthoudt:
@cacaodev : yes, perfect ! Thank you !

cappbot:
**Assignee:** [cacaodev](https://github.com/cacaodev).  **Labels:** #accepted, #needs-patch, AppKit, regression.  **What's next?** This issue needs a volunteer to write and submit code to address it.

PR #2342, opened 2015-04-07
FIXED: segmented control height in nib2cib

cacaodev:
The -tile method now leaves the frame height unchanged instead of
sizingToFit the theme attribute « min-size ».
The « min-size » is still in use when we layout the ephemeral subviews.

This is how other controls with a min-size work.
Also added -minimumFrameSize. This CPControl subclass will return the
frame for sizeToFit, i.e. when you want to sync the frame property
with the min-size theme attribute.

Tests : CPSegmentedControlTest & Nib2CibAlignment manual tests.

Fixes: #2341


Dogild:
You give up the method `- (CGSize)intrinsicContentSize` finally ?

cacaodev:
Yes because it's really a contraints thing. I prefer to conform to our current pattern we use for -sizeToFit. The difference between _minimumFrameSize and intrinsicContentSize is that in the latest, the width or height can be -1 if the control is flexible.

PR #2345, opened 2015-04-23
New: compilation gives more informations when getting an error or warnin...

Dogild:
...g

Previously, when the compilation of a project failed, we did not have so many informations about the cause of the problem.
Now, when gcc returns an error or a warning, we print in the console. This help to know what is the issue of the build we would like to have.


Dogild:
I close this pull request; it seems cappuccino does not build itself with this patch. The error is when compiling the ThemeDescriptors of a theme.

It seems that the issues comes from narwhal -> io-engine.cc, line ~230, the read of the file descriptor is blocking. (if someone want to have fun ^^)

PR #2346, opened 2015-04-23
FIXED: Closes all streams after a OS.popen()

primalmotion:
Previously, when popen was called, the 3 streams (stdin, stderr and stdourt) where never closed. This was the reason of the having an impossible number of open files.

This patch ensure all streams are closed after using them. This means that the ulimit trick is no more necessary

### DIFF
--- a/Tools/nib2cib/Converter.j
+++ b/Tools/nib2cib/Converter.j
@@ -119,8 +119,14 @@ ConverterConversionException = @"ConverterConversionException";
             // Compile xib or nib to make sure we have a non-new format nib.
             temporaryNibFilePath = FILE.join("/tmp", FILE.basename(aFilePath) + ".tmp.nib");
 
-            if (OS.popen(["/usr/bin/ibtool", aFilePath, "--compile", temporaryNibFilePath]).wait() === 1)
+            var p = OS.popen(["/usr/bin/ibtool", aFilePath, "--compile", temporaryNibFilePath]);
+
+            if (p.wait() === 1)
                 [CPException raise:ConverterConversionException reason:@"Could not compile file: " + aFilePath];
+
+            p.stdin.close();
+            p.stdout.close();
+            p.stderr.close();
         }
         else
         {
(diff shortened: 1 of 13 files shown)


Tools/nib2cib/Converter.j
```diff
@@ -119,8 +119,14 @@ ConverterConversionException = @"ConverterConversionException";
             // Compile xib or nib to make sure we have a non-new format nib.
             temporaryNibFilePath = FILE.join("/tmp", FILE.basename(aFilePath) + ".tmp.nib");
 
-            if (OS.popen(["/usr/bin/ibtool", aFilePath, "--compile", temporaryNibFilePath]).wait() === 1)
+            var p = OS.popen(["/usr/bin/ibtool", aFilePath, "--compile", temporaryNibFilePath]);
+
+            if (p.wait() === 1)
                 [CPException raise:ConverterConversionException reason:@"Could not compile file: " + aFilePath];
+
+            p.stdin.close();
```
aljungberg: This won't be called if the exception on line 125 was raised.

Tools/nib2cib/Converter.j
```diff
@@ -130,9 +136,15 @@ ConverterConversionException = @"ConverterConversionException";
         // Convert from binary plist to XML plist
         var temporaryPlistFilePath = FILE.join("/tmp", FILE.basename(aFilePath) + ".tmp.plist");
 
-        if (OS.popen(["/usr/bin/plutil", "-convert", "xml1", temporaryNibFilePath, "-o", temporaryPlistFilePath]).wait() === 1)
+        var p = OS.popen(["/usr/bin/plutil", "-convert", "xml1", temporaryNibFilePath, "-o", temporaryPlistFilePath]);
+
+        if (p.wait() === 1)
             [CPException raise:ConverterConversionException reason:@"Could not convert to xml plist for file: " + aFilePath];
 
+        p.stdin.close();
```
aljungberg: Same here.

aljungberg:
Nice work finding this, at long last! I'd use `try {} finally {}` to be more 'correct' since we always want the streams to be closed even if the code between opening and closing fails.

Dogild:
What is the problem if this one ?

Dogild:
+#fixed
Works here ! Travis messed up probably.

Thanks !

PR #2353, opened 2015-05-09
New: added option --objc to objj to generate objc files from a .j file

Dogild:
The command objj has now the option -c or --objc to generate objective-c files. The first arg is the file and the second the destination folder.

The objective-c class generated are basic and only contains IBOutlet and IBAction. To do that, the objective-c-parser is used. This parser is also used in xCodeCapp.


Objective-J/CommonJS/lib/objective-j.js
```diff
@@ -110,6 +111,14 @@ exports.run = function(args)
...
                     multipleFiles = true;
                     break
+
+                case "-c":
+                case "--objc":
+                    argv.shift();
+                    argv.unshift(require("file").path(module.path).dirname().join("objective-j/objective-c-parser.j"));
+                    argv.unshift("objj");
+
+                    return OS.system(argv);
```
primalmotion: I find that weird to eat all arguments for a command, then stop everything and start another command because there is a special parameter

primalmotion: The more I think about this, the more I think this should be a new tool

Dogild: What about `objjToObjcSkeleton` ?

primalmotion: objj2objcskeleton

mrcarlberg:
Works great!

Thanks!

mrcarlberg:
+#fixed

PR #2358, opened 2015-05-19
Fixed: documentVisibleRect in CPScrollView does not return the expected result

### LINKED ISSUE #2356: CPScrollView documentVisibleRect returns wrong values if some scale is applied to the document view
Hi ! You can follow the full story (with example and full explanation) on the Google group : 

https://groups.google.com/d/msg/objectivej/6-gY5AR1Vj8/TecLcTra2jsJ

The short story : 

When applying a scale to the document view of a CPScrollView, [CPScrollView documentVisibleRect] returns a wrong answer. This is confirmed by Alexandre who tried in Cocoa and got correct values.

The origin of the problem is the way documentVisibleRect is computed : in Capp, this is done by calling documentVisibleRect on _contentView (the clip view), which returns [self convertRect:[self bounds] fromView:_documentView], thus using CPView convertRect.

But convertRect apply any transformation used on the "fromView".

And this is bad.

An example is in the Google discussion.

My thought : when a scale of 0.5 is applied to a documentView, the documentVisibleRect should be scaled by 1 / 0.5 (thus  x2 and not /2) because a clip of 100x100, which show 100x100 pixels of a document at scale 1.0, shows 200x200 at scale 0.5.

The solution could be to use a special convertRect in the case of a clipview applying the "reverse transform" (sorry for this, I'm not sure how to say it in English).

Didier.

Dogild:
Previously, the method documentVisibleRect did not return the expected result. An issue happened when using this method on a scaled view. The converted rect was converted in the wrong context.

Unit test in Tests/AppKit/CPScrollViewTest.j

Fixed #2356

### DIFF
--- a/AppKit/CPClipView.j
+++ b/AppKit/CPClipView.j
@@ -195,7 +195,7 @@
 
 - (CGRect)documentVisibleRect
 {
-    return [self convertRect:[self bounds] fromView:_documentView];
+    return [_documentView visibleRect];
 }
 
 @end
--- a/Tests/AppKit/CPScrollViewTest.j
+++ b/Tests/AppKit/CPScrollViewTest.j
@@ -334,10 +334,48 @@
     [self assert:[CPNotificationCenterHelper registeredNotificationsForObserver:scrollView] equals:[] message:@"Notications registered for the scrollView in the notification center are wrong"];
 }
 
+- (void)testDocumentVisibleRect
+{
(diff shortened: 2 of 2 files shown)


Dogild:
@didierkorthoudt does it work like this ?
@daboe01 does it break something on your side (as you were the initiator of this feature)

daboe01:
should be ok from my side.

didierkorthoudt:
I forgot to say it's OK for me... Sorry for this. (and thank you)

PR #2359, opened 2015-05-21
FIXED CPDatePicker AM/PM time format for GB locale

herbatnik:
CPDatePicker was setting american date/time format for GB locale as it was checking against the locale language.

I have removed not needed _isEnglishFormat method and replaced it with _isAmericanFormat.

Now for US locale format stays as it should (mm/dd/yyyy 12h clock with AM/PM)
and for GB locale format is in EU standard (dd/mm/yyyy 24h clock)


Dogild:
Thanks for that ! American people and the date... :D

Next time, it could be very great if you follow our guideline (https://github.com/cappuccino/cappuccino/blob/master/CONTRIBUTING.md#commit-messages) for the commit message :)

herbatnik:
All right. I'll have that in mind if I'll have some fix or improvement in future.

PR #2360, opened 2015-05-25
New: added the option --xml-output-format for the command objj

Dogild:
Previously, it was not possible to specify the desired output format of the command objj.
Now we can have either the default format or a xml format if --xml-output-format is added.


Dogild:
@mrcarlberg, @primalmotion any objections for this PR ?

Objective-J/CommonJS/lib/objective-j.js
```diff
@@ -94,6 +94,7 @@ exports.run = function(args)
             print("  -I, --objj-include-paths       include a specific framework paths")
             print("  -h, --help                     print this help");
             print("  -m, --multifiles               launch objj on several files")
+            print("  --xml-output-format            specify the output format in xml.")
```
primalmotion: can you add -x so all parameters have their short version? also I think the name should simply be --xml

Objective-J/ObjJAcornCompiler.js
```diff
@@ -442,14 +457,32 @@ ObjJAcornCompiler.prototype.compilePass2 = function()
     compile(this.tokens, new Scope(null ,{ compiler: this }), pass2);
     for (var i = 0; i < this.warnings.length; i++)
     {
-       var message = this.prettifyMessage(this.warnings[i], "WARNING");
+        var warning = this.warnings[i],
+            type = "WARNING";
+
+        var message = this.prettifyMessage(warning, type);
```
primalmotion: do you need to prettify the message when xml output is set to xml?

primalmotion:
I've added a few comments, otherwise, looks great!

Objective-J/ObjJAcornCompiler.js
Dogild: Yes, I use the message in the xml

PR #2362, opened 2015-06-13
FIXED: In CPView, the transformation matrix was set at each drawing pass, even if the matrix was the identity matrix.

cacaodev:
Now we check if the highDPI ratio is == 1 and skip the setTransform call in this case.


cacaodev:
@Dogild : I did some profiling on the CPDisplayServer and in my tests (a view filled and a few squares stroked), this method is eating half of the drawing duration with ratio 1.

@Dogild maybe there is a way to have one unique flag (computed earlier) that tells us if there is both a context and an effective scaling.

Dogild:
@cacaodev The idea looks better to me ! What about something like ```_needToSetTransformMatrix = _highDPIRatio !== 1;``` instead of ```_needToSetTransformMatrix = YES;```

We need anyway to setTransform each time the frame of the view is changing.

Dogild:
Thanks !

PR #2367, opened 2015-08-04
Xcodecapp4.0

primalmotion:
NEW: XcodeCapp 4.0
    
    XcodeCapp 4.0 is a major release of our beloved tool. In a nutshell it:
    
    - allows to manage multiple projects simultaneously
    - allows to follow operations and cancel them
    - has a per project Error and Warning reporting
    - supports capp_env as you can define additional paths and objj include path per project
    - much more things

Made with Love by @Dogild and myself


Dogild:
We use it here for 1 month with our project and libraries and it works pretty well ! \o/

PR #2371, opened 2015-08-14
CPVisualEffectView & CPAppearance

primalmotion:
Two things here : 

**CPAppearance**
This patch adds the `CPAppearance` class and uses it where needed. It introduces new theme state that are propagated in the view hierarchy, so it is possible for a control to have different theme attribute values according to the current appearance of the control or an ancestor view.

nib2cib support has also been added.

**CPVisualEffectView**
This patch contains a very naive implementation of the
NSVisualEffectView. This will only work on very recent unreleased
version of Safari, but should be supported by all at some point. This
allow to use the Yosemite/iOS blurry effect.

Not all options are supported (especially the
`CPVisualEffectBlendingModeBehindWindow` mode…). But hey! it’s a start
:)

Tests in Manual/CPVisualEffectViewTest

screen shot 2015-08-14 at 02 05 08


AppKit/CPAppearance.j
```diff
@@ -0,0 +1,156 @@
...
+}
+
+
+#pragma mark -
+#pragma mark Initialization
+
+/*! Creates a CPAppearance object initialized to the specified appearance file in the specified bundle
+    This method does actually nothing special. It just creates a default appearance object
+*/
+- initWithAppearanceNamed:(CPString)aName bundle:(CPBundle)bundle
```
Dogild: Miss -(id)

AppKit/CPVisualEffectView.j
```diff
@@ -0,0 +1,236 @@
...
+{
+    var prop       = CPBrowserStyleProperty("backdrop-filter"),
+        dark       = [[self appearance] isEqual:[CPAppearance appearanceNamed:CPAppearanceNameVibrantDark]],
+        darkColor  = [CPColor colorWithHexString:@"1e1e1e"],
+        lightColor = [CPColor whiteColor];
+
+    if (shouldEnable)
+    {
+        if (prop)
+            self._DOMElement.style[prop] = "blur(30px)";
```
Dogild: It should be in `#IF PLATFORM(DOM)`

Dogild:
Thanks :)

PR #2374, opened 2015-08-21
New: possibility to create a theme for every object of the Appkit

Dogild:
The purpose of this feature is to create theme for every object of the AppKit. Previously, we could not theme object like CPColor, however CPColor could be very interesting to theme, value like alternateSelectedControlColor were hard coded in the framework. With this PR, we can now theme a CPColor and take some values from the current theme of the application. Then, this PR offers the possibility to refactor the themeDescriptors, some classes contains theme attributes for other class (CPTabView and  CPTabViewItem for instance).

How doe it work ? A new category _CPObject+Theme.j has been added to the AppKit. This category contains every theme methods needed to theme an object (previously this category was in CPView). We add utils method like _encodeThemeObjectsWithCoder: and _decodeThemeObjectsWithCoder: in this category to be able to code and decode easily a coder for every object (this is generic).

This PR did not refactor the themeDescriptors, it only add this new mechanism for the methods alternateSelectedControlColor and secondarySelectedControlColor of CPColor. We will need to refactor that gradually

### DIFF
--- a/AppKit/AppKit.j
+++ b/AppKit/AppKit.j
@@ -20,6 +20,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
 
+@import "_CPObject+Theme.j"
 @import "_CPToolTip.j"
 @import "CALayer.j"
 @import "CGGeometry.j"
--- a/AppKit/CPColor.j
+++ b/AppKit/CPColor.j
@@ -25,6 +25,7 @@
 
 @import "CGColor.j"
 
+@import "_CPObject+Theme.j"
 @import "CPCompatibility.j"
 @import "CPImage.j"
 
@@ -65,7 +66,8 @@ var cachedBlackColor,
     cachedOrangeColor,
     cachedPurpleColor,
     cachedShadowColor,
-    cachedClearColor;
+    cachedClearColor,
+    cachedThemeColor;
 
 /// @endcond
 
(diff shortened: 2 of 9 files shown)


Dogild:
We have just integrated this branch on our master. We will know soon if there is a regression or not.

mrcarlberg:
This looks great!

Can it help to add a protocol for the classes that are themeable? It might help to make it more generic in blend task.j?

Dogild:
Can we put a protocol on a category ?

-- 
Alexandre Wilhelm

On September 29, 2015 at 12:58:29 AM, Martin Carlberg (notifications@github.com) wrote:

This looks great!

Can it help to add a protocol for the classes that are themeable? It might help to make it more generic in blend task.j?

mrcarlberg:
Yes, it should work.

Dogild:
I have just added the category CPTheme. This protocol is implemented by every CPObject class when loading the AppKit.

Let me know if it's good or if you were thinking about something else

mrcarlberg:
Ok, this could work but I was thinking it might be better if only the classes that are themeable, like CPColor and CPView, conforms to the CPTheme protocol. The methods can still be implemented on CPObject. This makes it easier to know if an object really can be themeable.

Is that a good way to do it? Maybe it should help to more generic in blendtask.j?

Dogild:
The point is to have everything themable in the AppKit. But I see your point ! What do you think @primalmotion ?
-- 
Alexandre Wilhelm

On September 29, 2015 at 1:28:52 PM, Martin Carlberg (notifications@github.com) wrote:

Ok, this could work but I was thinking it might be better if only the classes that are themeable, like CPColor and CPView, conforms to the CPTheme protocol. The methods can still be implemented on CPObject. This makes it easier to know if an object really can be themeable.

Is that a good way to do it? Maybe it should help to more generic in blendtask.j?

PR #2375, opened 2015-08-25
Fixed: crash when using a viewBased CPTableView with appearance

Dogild:
Previously, when using a viewBased CPTableView Cappuccino just crashed.
This was due to the new appearance feature. Indeed, at the end of the method initWithCoder: of CPView, we set the appearance of the current view and the themeStates of the view and its subviews. For an unknown reason some of these subviews were not completely initialized (see comment line 3742), this made crashed cappuccino because _themeState was null.

To fix this issue, we send the method setAppearance when the current stack was performed. (Magic setTimeout...).

This is for sure not the ultimate best fix, but it's prevent to have a broken master. We will need in the future to fix this issue in a better way.

### DIFF
--- a/AppKit/CPView.j
+++ b/AppKit/CPView.j
@@ -3737,7 +3737,15 @@ var CPViewAutoresizingMaskKey       = @"CPViewAutoresizingMask",
             _themeAttributes[attributeName] = CPThemeAttributeDecode(aCoder, attributeName, attributes[count], _theme, themeClass);
         }
 
-        [self setAppearance:[aCoder decodeObjectForKey:CPViewAppearanceKey]];
+        var appearance = [aCoder decodeObjectForKey:CPViewAppearanceKey];
+
+        // TODO : find a way to fix this horrible things
+        // We only set the appearance here when everything and every subviews were unarchived.
+        // Otherwise this causes a crash with some subviews not unarchived but already created (how ? init is not called, neither initWithFrame or initWithCoder) in the coder (I have no idea why these views are present in _objects of the coder)
+        // To reproduce, uncomment the timeout and open the test Tests/Manual/TableTests/ViewBasedCib
+        window.setTimeout(function() {
+            [self setAppearance:appearance];
+        },0);
 
         [self setNeedsDisplay:YES];
         [self setNeedsLayout];
--- a/AppKit/CPView.j
+++ b/AppKit/CPView.j
@@ -231,8 +231,6 @@ var CPViewHighDPIDrawingEnabled = YES;
     CPView              _nextKeyView;
     CPView              _previousKeyView;
 
-    unsigned            _viewClassFlags;
-
     // ToolTips
     CPString            _toolTip    @accessors(getter=toolTip);
     Function            _toolTipFunctionIn;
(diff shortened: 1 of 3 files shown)


AppKit/CPView.j
```diff
@@ -3737,7 +3737,15 @@ var CPViewAutoresizingMaskKey       = @"CPViewAutoresizingMask",
...
         }
 
-        [self setAppearance:[aCoder decodeObjectForKey:CPViewAppearanceKey]];
+        var appearance = [aCoder decodeObjectForKey:CPViewAppearanceKey];
+
+        // TODO : find a way to fix this horrible things
+        // We only set the appearance here when everything and every subviews were unarchived.
+        // Otherwise this causes a crash with some subviews not unarchived but already created (how ? init is not called, neither initWithFrame or initWithCoder) in the coder (I have no idea why these views are present in _objects of the coder)
+        // To reproduce, uncomment the timeout and open the test Tests/Manual/TableTests/ViewBasedCib
+        window.setTimeout(function() {
```
daboe01: in other places we use [[CPRunLoop currentRunLoop] performSelector:@selector(setAppearance:) target:self argument:appearance order:0 modes:[CPDefaultRunLoopMode]];

AppKit/CPView.j
```diff
@@ -3737,7 +3737,15 @@ var CPViewAutoresizingMaskKey       = @"CPViewAutoresizingMask",
             _themeAttributes[attributeName] = CPThemeAttributeDecode(aCoder, attributeName, attributes[count], _theme, themeClass);
         }
 
-        [self setAppearance:[aCoder decodeObjectForKey:CPViewAppearanceKey]];
+        var appearance = [aCoder decodeObjectForKey:CPViewAppearanceKey];
+
+        // TODO : find a way to fix this horrible things
```
daboe01: TODO : find a way to fix this kind of 'race condition'

Dogild:
I don't like this one at all. I will think to propose a better fix

Dogild:
@primalmotion, @aparajita , @aljungberg can you take a look to the last commit ! I removed an optimization method and I would like to know what you think about that.

Dogild:
Everything is slow with the pr

aparajita:
> @primalmotion, @aparajita , @aljungberg can you take a look to the last commit ! I removed an optimization method and I would like to know what you think about that.

You said things were slow after removing that method, right?

Regards,

  Aparajita

Dogild:
@aparajita Yep, I will try again when I'm back though

Dogild:
I just did few changes.

Now, we will create a canvas for the DOMElement if viewWillDraw or drawRect is implemented by the subclass (previously it was only if drawRect).

layoutSubviews, viewWillLayout and viewDidLayout are called each time the view has to layout whatever if the subclass implements these methods. Some computation are made in viewDidLayout of the CPView.

Speed impacts are not significant. Previously it was a bit slower because we had to create a canvas for each views, now it's not the case anymore.

@aparajita what do you think about that ?

aparajita:
Seems okay to me.

primalmotion:
thanks!

aljungberg:
>  Indeed, at the end of the method initWithCoder: of CPView, we set the appearance of the current view and the themeStates of the view and its subviews. For an unknown reason some of these subviews were not completely initialized (see comment line 3742), this made crashed cappuccino because _themeState was null.

That sounds normal to me? Since we have a view hierarchy the views have to be init'ed in some order, so for all but the last view there will be not yet fully loaded views in the tree. That's why we have `awakeFromCib`.

Still seems reasonable to lazy create the canvas.

Dogild:
The issue was that there were already initialized (I don't know how they were init actually) but not completely. The views have some ivars but not all of them as they should have with the method `init`

PR #2377, opened 2015-09-02
Fixed: underscores missing when accessing CPColor's components ivar

daboe01:
previously, the ivar was erroneously accessed by the property name using dot syntax.
this is corrected by adding the underscore appropriately.
credits: <https://groups.google.com/forum/?fromgroups#!topic/objectivej/DbHN1ffcjO0>

### DIFF
--- a/AppKit/CoreGraphics/CGContextCanvas.j
+++ b/AppKit/CoreGraphics/CGContextCanvas.j
@@ -589,7 +589,7 @@ function CGContextDrawImage(aContext, aRect, anImage)
 
 function to_string(aColor)
 {
-    return "rgba(" + ROUND(aColor.components[0] * 255) + ", " + ROUND(aColor.components[1] * 255) + ", " + ROUND(255 * aColor.components[2]) + ", " + aColor.components[3] + ")";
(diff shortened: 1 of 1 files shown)


aparajita:
Danke. You tested it?

daboe01:
no.

maybe michael can test it in his app?

@michael:
you can download the fixed capp framework here:
<https://github.com/daboe01/cappuccino/tree/fix-CGContextDrawRadialGradient>

best greetings,
daniel

Dogild:
More informations here : https://groups.google.com/forum/?fromgroups=#!topic/objectivej/DbHN1ffcjO0

The manual test CPGradientTest looks good with the current version ! This fix will work when using CPColor's array instead of using the method `CGGradientCreateWithColorComponents()` (which is how cocoa works)

Dogild:
According to the Cocoa doc (https://developer.apple.com/library/mac/documentation/GraphicsImaging/Reference/CGGradient/#//apple_ref/c/func/CGGradientCreateWithColors), when using the gradient color you need to pass CGColor and not CPColor. 

Everything looks good in the framework right now (manual test in CPGradientTest)

PR #2379, opened 2015-09-25
Fixed: The init method ’initWithObjects:count:’ for CPArray did not work

mrcarlberg:
The init method ’initWithObjects:count:’ for CPArray did not work on native CPJavaScriptArray when ’count’ is not the same as the length on the provided array.

A test case for this is also added in the test class CPArrayTest


Dogild:
Ok for me ! You can merge it when you want !

PR #2381, opened 2015-09-30
CPDictionary handles duplicate keys in correct way on init

mrcarlberg:
CPDictionary initWithObjectsAndKeys: was not compliant with Cocoa or the init method ’initWithObjects:forKeys:’ when duplicated keys was passed to the method.

Test cases are also added

### DIFF
--- a/Foundation/CPDictionary.j
+++ b/Foundation/CPDictionary.j
@@ -273,18 +273,16 @@ var CPDictionaryMaxDescriptionRecursion = 10;
     if (self)
     {
         // The arguments array contains self and _cmd, so the first object is at position 2.
-        var index = 2;
-
-        for (; index < argCount; index += 2)
+        while (argCount-- > 2)
         {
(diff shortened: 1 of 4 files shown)


Dogild:
Hi Martin,

I saw you have modified the tests for the timeZones. Not sure if this is the good solution. In cocoa, you will get the timeZone with or without the daylight saving (depends on the date of the system). I think we could do the same in CPTimeZone with the javascript date.

mrcarlberg:
Yeah, that is a solution. For this I just made the test cases work as CPTimezone now depends on how the keys are ordered in a CPDictionary. The order of the keys is not defined in a CPDictionary so that is, as you say, a bug in CPTimeZone.
Right now I don't have time to fix CPTimeZone but we should put it on the todo list.  :smiley:

Dogild:
Thanks

+#fixed

PR #2383, opened 2015-10-01
Use the faster objj_msgSend instead of the old slower objj_msgSend

mrcarlberg:
In some places the objj_msgSend function is called directly. Most of the times the old slower version is called. This commit will use the never faster version instead.


Dogild:
Looks good to me ! I'll let you merge it

PR #2384, opened 2015-10-06
Speedup of the Cappuccino framework by more efficient objj_msgSend

mrcarlberg:
This pull request will make Objective-J run faster by implementing a more efficient ```objj_msgSend``` function and ```objj_method``` implementation. The compiler can also optionally inline the function to increase the speed even more.

Initial tests show that the speed increase is more than 50% in special cases. For a general Objective-J application the speed increase is around 20%.

# The speed increase is possible by the following improvements:
1. More efficient ```objj_method``` implemention. That bad thing here is that the ```name``` property on ```objj_method``` has to change to ```method_name```.
2. ```objj_msgSend``` and ```objj_msgSendSuper``` functions has a more efficient implementation.
3. Array and Dictionary Literals generate more efficient code.
4. The ```objj_msgSend``` and ```objj_msgSendSuper``` functions are optionally inlined by the compiler.

# How to use:
• The ```Release``` version of the Cappuccino framework will have inlined ```objj_msgSend``` and ```objj_msgSendSuper``` functions. The ```Debug```version will use the old ```obj_msgSend``` functions. 
• To generate inlined code when compiling in the browser add the below line in the ```index[-debug].html``` file.
```
OBJJ_COMPILER_FLAGS = ["IncludeDebugSymbols", "IncludeTypeSignatures", "InlineMsgSend"];
```
• To generate inline code for the ```Release``` version of your project when building from command line change ```task.setCompilerFlags("-O");``` to ```task.setCompilerFlags("-O2");``` in the project Jakefile. This is needed for all your subprojects and frameworks too.

A new version of OJTest framework (https://github.com/cappuccino/OJTest) is needed to run the test cases. Copy it to the package folder in the Narwhal directory.

### DIFF
--- a/Objective-J/Bootstrap.js
+++ b/Objective-J/Bootstrap.js
@@ -45,6 +45,23 @@ if (DOMBaseElementsCount > 0)
         pageURL = new CFURL(DOMBaseElementHref, pageURL);
 }
 
+// Set compiler flags
+
+if (OBJJ_COMPILER_FLAGS)
+{
+    var flags = 0;
+    for (var i = 0; i < OBJJ_COMPILER_FLAGS.length; i++)
+    {
+        var flag = ObjJAcornCompiler.Flags[OBJJ_COMPILER_FLAGS[i]];
+
+        if (flag != null)
+        {
+            flags |= flag;
+        }
+    }
+    exports.setCurrentCompilerFlags(flags);
+}
+
 // Turn the main file into a URL.
 var mainFileURL = new CFURL(window.OBJJ_MAIN_FILE || "main.j"),
 
--- a/Objective-J/Bootstrap.js
+++ b/Objective-J/Bootstrap.js
@@ -47,7 +47,7 @@ if (DOMBaseElementsCount > 0)
 
 // Set compiler flags
 
-if (OBJJ_COMPILER_FLAGS)
+if (typeof OBJJ_COMPILER_FLAGS !== 'undefined')
 {
     var flags = 0;
     for (var i = 0; i < OBJJ_COMPILER_FLAGS.length; i++)
--- a/AppKit/Jakefile
+++ b/AppKit/Jakefile
@@ -34,7 +34,7 @@ appKitTask = framework ("AppKit", function(appKitTask)
     }).join(" ");
 
     if ($CONFIGURATION === "Release") {
-        appKitTask.setCompilerFlags(($INLINE_MSG_SEND ? "-O2 " : "-O ") + INCLUDES);
+        appKitTask.setCompilerFlags("-O2 " + INCLUDES);
     }
     else
         appKitTask.setCompilerFlags("-DDEBUG -g " + INCLUDES);
--- a/AppKit/Jakefile
+++ b/AppKit/Jakefile
@@ -33,8 +33,9 @@ appKitTask = framework ("AppKit", function(appKitTask)
         return "--include \"" + aFilename + "\"";
     }).join(" ");
 
-    if ($CONFIGURATION === "Release")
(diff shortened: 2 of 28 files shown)


aljungberg:
> The objj_msgSend and objj_msgSendSuper functions are optionally inlined by the compiler.

How does that handle method swizzling?

mrcarlberg:
The same code is just inlined so it works fine.
The thing you can't do is use decorators on obj_msgSend.
The compiler will only optionally inline the functions on the release version when building from command line.

aljungberg:
Ah I get it. Yeah the most common objj_msgSend decorators are all about debugging so enabling this option for release code makes sense.

Shouldn't this PR have some code generation unit tests?

mrcarlberg:
Yes, it definitively needs some code generation unit tests. I have plans to do the ```OutputTest``` in two different versions, one standard and one inlined. But I will add those when I know that the new code works for everyone.

boucher:
Very cool!

primalmotion:
That doesn't work on our built application either in release, release+press  or release+press+flatten

The page just hangs. If I make the browser to stop on all exceptions I get an exception, but no message is logged anywhere

screen shot 2015-10-06 at 10 40 01

primalmotion:
edit: in release, it doesn't load, but actually logs an exception:

```
TypeError: undefined is not an object (evaluating 'CPDictionary.isa.method_msgSend')
```

mrcarlberg:
Hmm, this is strange. Are you sure it is running against a the new Objective-J Runtime?

 It feels like you are running on an old Runtime as it gets undefined when evaluating ```CPDictionary.isa.method_msgSend``` All ```objj_class```objects should have the attribute ```method_msgSend```

primalmotion:
my bad actually :)

I did not realized I had to activate the -O2 flag for *all* libraries 

It's working great, we'll do some testing and report any issue.

Great Job!

mrcarlberg:
I have now added code generation unit test cases with inlined objj_msgSend functions for all tests in OutputTest

mrcarlberg:
I have now fixed a problem with compiler flags when compiling in the browser.

mrcarlberg:
I don't have any more outstanding issues. We are now using it in our project and are happy with it.

aljungberg:
Looks good so far. Some minor points we should be aware of:

- larger build output since each message send is more code
- `_objj_forward` becomes a keyword
- uses (very slightly) more memory to add `method_msgSend` to the isas after class initialisation

Did you compare the size of your build output for your app with and without this? What about memory usage in the browser? If it's 20-50% faster I imagine it can't be too significant.

mrcarlberg:
The code size does increase:
```Foundation```goes from 987,634 bytes to 1,151,583 bytes. (+17%)
```AppKit```goes from 3,587,337 bytes to 4,343,767 bytes. (+21%)
This is not compressed and not zipped.

I can't see any difference in memory usage in the Browser but as you say it should be a small increase.

File: Objective-J/Bootstrap.js [hunk 1]
```diff
@@ -45,6 +45,23 @@ if (DOMBaseElementsCount > 0)
         pageURL = new CFURL(DOMBaseElementHref, pageURL);
 }
 
+// Set compiler flags
+
+if (OBJJ_COMPILER_FLAGS)
```
zittix: This generated a Uncaught ReferenceError: OBJJ_COMPILER_FLAGS is not defined on Chrome. 

```js
typeof OBJJ_COMPILER_FLAGS !== 'undefined'
```

would be better.

mrcarlberg: Thanks for finding this problem. I'll make a fix for this.

Dogild:
Are we good with this one @mrcarlberg ?

PR #2386, opened 2015-10-21
Faster theme state handling

mrcarlberg:
ThemeState was changed from a bit mask to a Javascript object last year. This was done to handle more states as a bit mask is limited to 32 bits.
The performance impact of this was not good and made theme handling slower.
This pull request adds a cache mechanism to get back some of the lost performance.
Also fixed some other small performance improvements.

I tested to load a large cib file in Safari 9.0 and it took 1.80 seconds. With this change it will load in 1.60 seconds. Similar speed gains in Chrome too.

### DIFF
--- a/AppKit/CPButton.j
+++ b/AppKit/CPButton.j
@@ -280,7 +280,7 @@ CPButtonImageOffset   = 3.0;
             break;
 
         case CPOffState:
-            [self unsetThemeState:[CPThemeStateSelected, CPButtonStateMixed, CPThemeStateHighlighted]];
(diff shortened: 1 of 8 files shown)


aparajita:
Thanks. There are a few places where the formatting isn't quite right:

- `{` is always on a separate line
- `while` with a single statement gets no braces
- Always put a blank line after `}`

mrcarlberg:
We will never merge this as #2389 is built on top of this pull request. I will delete this as soon as #2389 is merged into master.

mrcarlberg:
I'm closing this as it is included in #2389

PR #2339, opened 2015-03-27
CPTabView improvements

cacaodev:
NEW: CPTabViewItem -setImage & nib2cib
FIXED: CPTabView -insertTabViewItem:atIndex:
NEW: CPTabView bindings: CPContentBinding, CPSelectedIndexBinding and CPSelectionIndexesBinding with no selection & multiple selection placeholders.
NEW: CPTabView -tabViewItemWithViewController:

With ojtest and manual tests. New test Manual/CPTabViewBindings/

PR #2391, opened 2015-10-26
Fixed: default selection of a CPTabView item

zittix:
With the lately merged PR #2339 it breaks the default CPTabView item selection. The item is correctly selected inside the internal state of the object but the tab view is not loaded at all.

This is due to the fact that the _selectedTabViewItem contains the selected item from the init call and the selectTabViewItemAtIndex: method is called within the awakeFromCib method to load the selected tab, which fails as this happens to be already selected.

This PR fixes that.

### DIFF
--- a/AppKit/CPTabView.j
+++ b/AppKit/CPTabView.j
@@ -789,7 +789,13 @@ var CPTabViewItemsKey               = "CPTabViewItemsKey",
         var idx = [self indexOfTabViewItem:_selectedTabViewItem];
 
         if (idx !== CPNotFound)
+        {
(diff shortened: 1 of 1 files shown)


Dogild:
Hi @zittix,

Thanks for the contribution, the only issue with the PR is the commit message. Can you please follow the  style of our commit message (more informations here https://github.com/cappuccino/cappuccino/blob/master/CONTRIBUTING.md#commit-messages)

zittix:
@Dogild sorry for that. This is now fixed.

cacaodev:
Merged, thanks zittix !

PR #2394, opened 2015-10-30
mouseEntered and mouseExited propagation fix proposal

didierkorthoudt:
In a hierarchy of views, mouseEntered and mouseExited events are propagated via two mechanisms : the CPWindow sendEvent and the CPResponder mouseEntered / mouseExited methods.

In the CPWindow sendEvent, the event was sent to all views concerned by mouse event, from the bottommost to the topmost.

We take the following example : view 1 is a subclass of CPView implementing a mouseEntered event and containing view 2, a standard CPView (thus not implementing a specific mouseEntered method), that exactly fills the same space than view 1.

In the CPWindow sendEvent, a mouseEntered event was sent to view 1 (thus catched by the view 1 mouseEntered method), then sent to view 2. As view 2 doesn't deal with this event, the standard CPResponder mouseEntered method was called, passing the event to the next responder of view 2, aka view 1, which then catched it again.

The proposed fix modifies this process : 

In the CPWindow sendEvent, instead of sending the event to all entered/exited views, it now only send it to the topmost view. Then, if the view doesn't cope with this event, the CPResponder sends it to the next responder (going down the hierarchy) but only if the event has to be sent, that is if the next responder didn't yet received the event in the past.

If a view deals with the event, it can, if needed, call the super method to try to propagate the event down the hierarchy.

If you try the following sample app without the fix, you can observe : 

> Left sample : 

- entering the red square (which implements the mouseEntered/Exited methods) fires a mouseEntered
- then entering the green square (standard CPView) also fires a mouseEntered by the red square
- leaving the green square fires a mouseExited by the red square (which is bad)
- leaving the red square fires a mouseExited (which is normal)

> Middle sample : 

Everything works as expected as the red square is above the green one

> Right sample : 

A little more complex as the blue rectangle covers half the red square.

If you try this with the proposed fix, everything should be fine.

```
@class SomeViewSubclass;

@implementation AppController : CPObject
{
    @outlet CPWindow    theWindow;
}

- (void)applicationDidFinishLaunching:(CPNotification)aNotification
{
    // This is called when the application is done loading.
}

- (void)awakeFromCib
{
    // This is called when the cib is done loading.
    // You can implement this method on any object instantiated from a Cib.
    // It's a useful hook for setting up current UI values, and other things.
    
    // In this case, we want the window from Cib to become our full browser window
    [theWindow setFullPlatformWindow:NO];
    
    
    // Simple test
    
    var aView = [[SomeViewSubclass alloc] initWithFrame:CGRectMake(10, 10, 50, 50)];
    
    [aView setBackgroundColor:[CPColor redColor]];
    
    [[theWindow contentView] addSubview:aView];
    
    var anotherView = [[CPView alloc] initWithFrame:CGRectMake(10, 10, 30, 30)];
    
    [anotherView setBackgroundColor:[CPColor greenColor]];
    
    [aView addSubview:anotherView];
    
    
    
    // Second simple (opposite)
    
    
    
    var aView = [[CPView alloc] initWithFrame:CGRectMake(100, 10, 50, 50)];
    
    [aView setBackgroundColor:[CPColor greenColor]];
    
    [[theWindow contentView] addSubview:aView];
    
    var anotherView = [[SomeViewSubclass alloc] initWithFrame:CGRectMake(10, 10, 30, 30)];
    
    [anotherView setBackgroundColor:[CPColor redColor]];
    
    [aView addSubview:anotherView];

    
    // Third
    
    
    var aView = [[SomeViewSubclass alloc] initWithFrame:CGRectMake(200, 10, 50, 50)];
    
    [aView setBackgroundColor:[CPColor redColor]];
    
    [[theWindow contentView] addSubview:aView];
    
    var halfView = [[CPView alloc] initWithFrame:CGRectMake(0, 0, 25, 50)];
    
    [halfView setBackgroundColor:[CPColor blueColor]];
    
    [aView addSubview:halfView];
    
    var anotherView = [[CPView alloc] initWithFrame:CGRectMake(10, 10, 30, 30)];
    
    [anotherView setBackgroundColor:[CPColor greenColor]];
    
    [aView addSubview:anotherView];
}

@end

@implementation SomeViewSubclass : CPView
{
}

- (void)mouseEntered:(CPEvent)anEvent
{
    CPLog.trace("######### mouseEntered in "+self);
}

- (void)mouseExited:(CPEvent)anEvent
{
    CPLog.trace("######### mouseExited from "+self);
}

@end
```

### DIFF
--- a/AppKit/CPResponder.j
+++ b/AppKit/CPResponder.j
@@ -188,7 +188,9 @@ CPDeleteForwardKeyCode  = 46;
 
 - (void)mouseEntered:(CPEvent)anEvent
 {
-    [_nextResponder performSelector:_cmd withObject:anEvent];
+    if ([[anEvent window] shouldPropagateMouseEnteredEventToView:_nextResponder])
+        
+        [_nextResponder performSelector:_cmd withObject:anEvent];
 }
 
 /*!
@@ -197,7 +199,9 @@ CPDeleteForwardKeyCode  = 46;
 */
 - (void)mouseExited:(CPEvent)anEvent
 {
-    [_nextResponder performSelector:_cmd withObject:anEvent];
+    if ([[anEvent window] shouldPropagateMouseExitedEventToView:_nextResponder])
+        
+        [_nextResponder performSelector:_cmd withObject:anEvent];
 }
 
 /*!
--- a/AppKit/CPResponder.j
+++ b/AppKit/CPResponder.j
@@ -189,7 +189,6 @@ CPDeleteForwardKeyCode  = 46;
 - (void)mouseEntered:(CPEvent)anEvent
 {
     if ([[anEvent window] shouldPropagateMouseEnteredEventToView:_nextResponder])
-        
         [_nextResponder performSelector:_cmd withObject:anEvent];
 }
 
@@ -200,7 +199,6 @@ CPDeleteForwardKeyCode  = 46;
 - (void)mouseExited:(CPEvent)anEvent
 {
     if ([[anEvent window] shouldPropagateMouseExitedEventToView:_nextResponder])
-        
         [_nextResponder performSelector:_cmd withObject:anEvent];
 }
 
--- a/AppKit/CPControl.j
+++ b/AppKit/CPControl.j
@@ -511,6 +511,8 @@ var CPControlBlackColor = [CPColor blackColor];
 
 - (void)mouseEntered:(CPEvent)anEvent
 {
+    [super mouseEntered:anEvent];
+
     if (![self isEnabled])
         return;
 
@@ -519,6 +521,8 @@ var CPControlBlackColor = [CPColor blackColor];
 
 - (void)mouseExited:(CPEvent)anEvent
 {
+    [super mouseExited:anEvent];
+
     var currentLocation = [self convertPoint:[anEvent locationInWindow] fromView:nil],
         isWithinFrame = [self tracksMouseOutsideOfFrame] || CGRectContainsPoint([self bounds], currentLocation);
 
--- a/AppKit/CPWindow/CPWindow.j
+++ b/AppKit/CPWindow/CPWindow.j
@@ -198,6 +198,7 @@ var CPWindowActionMessageKeys = [
     CPView                              _toolbarView;
 
     CPArray                             _mouseEnteredStack;
+    CPArray                             mouseEnteredStack;
     CPView                              _leftMouseDownView;
     CPView                              _rightMouseDownView;
 
@@ -1985,13 +1986,16 @@ CPTexturedBackgroundWindowMask
             if (!_mouseEnteredStack)
                 _mouseEnteredStack = [];
 
+            if (!mouseEnteredStack)
+                mouseEnteredStack = [];
+            
             var hitTestView = [_windowView hitTest:point];
 
             if ([_mouseEnteredStack count] && [_mouseEnteredStack lastObject] === hitTestView)
                 return [hitTestView mouseMoved:anEvent];
 
-            var view = hitTestView,
-                mouseEnteredStack = [];
+            var view = hitTestView;
+            mouseEnteredStack = [];
 
             while (view)
             {
@@ -2013,8 +2017,7 @@ CPTexturedBackgroundWindowMask
             {
                 var event = [CPEvent mouseEventWithType:CPMouseExited location:point modifierFlags:[anEvent modifierFlags] timestamp:[anEvent timestamp] windowNumber:_windowNumber context:nil eventNumber:-1 clickCount:1 pressure:0];
 
-                for (; index < count; ++index)
-                    [_mouseEnteredStack[index] mouseExited:event];
+                [_mouseEnteredStack[count-1] mouseExited:event];
             }
 
             index = deviation + 1;
@@ -2024,14 +2027,25 @@ CPTexturedBackgroundWindowMask
             {
                 var event = [CPEvent mouseEventWithType:CPMouseEntered location:point modifierFlags:[anEvent modifierFlags] timestamp:[anEvent timestamp] windowNumber:_windowNumber context:nil eventNumber:-1 clickCount:1 pressure:0];
 
-                for (; index < count; ++index)
-                    [mouseEnteredStack[index] mouseEntered:event];
+                [mouseEnteredStack[count-1] mouseEntered:event];
             }
 
             _mouseEnteredStack = mouseEnteredStack;
 
             [hitTestView mouseMoved:anEvent];
     }
+}
+
+// The two following methods tell the sender (CPResponder) if it can or not propagate the current event deeper in the view hierarchy
+
+- (BOOL)shouldPropagateMouseEnteredEventToView:(CPView)aView
+{
+    return (![_mouseEnteredStack containsObject:aView]);
+}
+
+- (BOOL)shouldPropagateMouseExitedEventToView:(CPView)aView
+{
+    return (![mouseEnteredStack containsObject:aView]);
 }
 
 /*!


aparajita:
@primalmotion @Dogild Can you take a look at this? I'm a little worried about messing with mouse tracking, a lot of AppKit classes depend on getting mouseEntered/mouseExited correctly. 

The larger issue is that Cappuccino's mouse tracking is broken. In Cocoa, you don't get mouseEntered/mouseExited events unless you specifically add a tracking rect, whereas Cappuccino sends the events to every view under the mouse. Clearly this is a bug; the window should not be sending the mouse event to every view under the pointer if the responder chain then passing the event up the view hierarchy.

If this code fixes that problem without breaking all of the classes that override mouseEntered/mouseExited, then we can accept it for now (although there are a bunch of stylistic and formatting issues that need to be fixed), but ultimately we really need to implement the correct solution.

AppKit/CPResponder.j
```diff
@@ -188,7 +188,9 @@ CPDeleteForwardKeyCode  = 46;
 
 - (void)mouseEntered:(CPEvent)anEvent
 {
-    [_nextResponder performSelector:_cmd withObject:anEvent];
+    if ([[anEvent window] shouldPropagateMouseEnteredEventToView:_nextResponder])
+
```
Dogild: Style here, you don't need a line break

AppKit/CPResponder.j
```diff
@@ -197,7 +199,9 @@ CPDeleteForwardKeyCode  = 46;
 */
 - (void)mouseExited:(CPEvent)anEvent
 {
-    [_nextResponder performSelector:_cmd withObject:anEvent];
+    if ([[anEvent window] shouldPropagateMouseExitedEventToView:_nextResponder])
+
```
Dogild: Same,  you don't need a line break

didierkorthoudt:
OK, I'll fix code styling ASAP.
I'll do more tests this weekend to be sure everything is OK with this proposal.
As Aparajita said, some capp UI items may need to propagate the event if they rely on the "bad" events flow.

didierkorthoudt:
I've found another problem with the current system (with ou without the fix) : it's base solely on the view hierarchy (a view inside another view). But (and it's the case in my app) if you have a view over another but not a subview of this other, event propagation is broken.

Moreover, current implementation costs a lot of processing time passing events to view that doesn't need these.

I first try to fix the present issue and will afterwards try to look at the more general issue.

didierkorthoudt:
I've fixed the 2 style errors.

I've also tested the proposed fix with all capp objects having mouseEntered / Exited : 
- CPControl
- CPScroller (does call super by itself)
- CPSplitView
- CPTableHeaderView
- CPTokenField
and found nothing strange.

Menus seem to work correctly.

I'll use the fix on my own app to check it in real life.

aparajita:
The current system really is broken. What we really need is an implementation of NSTrackingArea.

aparajita:
> I've also tested the proposed fix with all capp objects having mouseEntered / Exited :
> 
> 	• CPControl
> 	• CPScroller (does call super by itself)
> 	• CPSplitView
> 	• CPTableHeaderView
> 	• CPTokenField and found nothing strange.
> Menus seem to work correctly.

Thanks for checking that.

didierkorthoudt:
I'm currently looking at NSTrackingArea.

If you're interested, I could try to propose an implementation...

aparajita:
> I'm currently looking at NSTrackingArea.
> 
> If you're interested, I could try to propose an implementation…

Please do!

didierkorthoudt:
My tests are no more fully functional. I've made changes in my test app and must find why it stopped working (I hope this is only due to some mistake in my code reorganization).

I'll continue investigation this evening.

@aparajita I've begun... but it could take a while...  ;-)

Dogild:
Sorry @didierkorthoudt, I didn't finish to review this PR. I'll let you finish the work you just started !

What you can do as well is to create unit tests for the issue/feature. It's not funny funny to do it, but once you have every cases covered by tests it will be way more easy to fix/create the new feature ;)

didierkorthoudt:
OK, found !

I didn't remember that I had modified the CPControl implementation...

CPControl was blocking the CPResponder event propagation, thus relying on the CPWindow sendEvent. I've pushed a fix and everything now seems functional.

This two way event propagation schema is, as @aparajita said, way too problematic. My proposed fix should clean a little the situation but it's not a long term solution (as some other problems remains, as I mentioned earlier).

So I'm going to dive in CPTrackingArea...

@Dogild , is there somewhere some guidelines to do unit test ?

Dogild:
Well, we don't have any guidelines or something. It's something I would like to do (specially I talk about that for the next conf).

The idea for this test is to create a view hierarchy in your setup method, and then in a test send an event to the CPApp. The good way to see if the method mouseExisted and mouseEntered are called would be to use the mock of OJTest. You should start with that ;)

Exemple of mock (from another lib) :

```
@import <AppKit/AppKit.j>
@import <OJMoq/OJMoq.j>

@import "../GraphCappuccino.j"
@import "TestModel.j"

var data;

@implementation NUTreeViewSelectionTest : OJTestCase
{
    CPWindow    _window;
    NUTreeView  _treeView;

    OJMoqSpy dataSourceSpy
}

+ (void)setUp
{
    var person0 = [Person initWithFirstName:@"Person0"],
        person1 = [Person initWithFirstName:@"Person1"],
        person2 = [Person initWithFirstName:@"Person2"],
        person3 = [Person initWithFirstName:@"Person3"],
        person4 = [Person initWithFirstName:@"Person4"],
        person5 = [Person initWithFirstName:@"Person5"],
        person6 = [Person initWithFirstName:@"Person6"],
        person7 = [Person initWithFirstName:@"Person7"];

    [person0 setChildren:[person1]];
    [person1 setChildren:[person2, person3, person4]];
    [person2 setChildren:[person5, person6]];
    [person3 setChildren:[person7]];

    data = [person0, person1, person2, person3, person4, person5, person6, person7];
}

+ (void)tearDown
{
    data = [];
}

- (void)setUp
{
    [[CPApplication alloc] init];

    _window = [[CPWindow alloc] initWithContentRect:CGRectMake(0.0, 0.0, 10000.0, 10000.0) styleMask:CPWindowNotSizable];

    _treeView = [[NUTreeView alloc] initWithFrame:CGRectMake(0, 0, 5000, 5000)];
    [_treeView setAllowsMultipleSelection:NO];
    [_treeView setSavePrimaryKeyPath:@"firstName"];
    [_treeView setAlignment:CPMinXEdge];
    [_treeView setAutoNestTreeView:YES];
    [_treeView setDelegate:self];
    [_treeView setDataSource:self];

    [[_window contentView] addSubview:_treeView];

    [_treeView reloadData];
    dataSourceSpy = spy(self);
}

- (void)tearDown
{
    [dataSourceSpy reset];
}

- (void)testSelectItemAndDeselectItemWithMultipleSelection
{
    [_treeView setAllowsMultipleSelection:YES];

    [dataSourceSpy reset];
    [dataSourceSpy selector:@selector(selectionShouldChangeInTreeView:pendingItem:) times:1 arguments:[_treeView, data[0]]];
    [dataSourceSpy selector:@selector(treeView:shouldSelectItem:) times:1 arguments:[_treeView, data[0]]];
    [dataSourceSpy selector:@selector(treeView:willSelectItem:) times:1 arguments:[_treeView, data[0]]];
    [dataSourceSpy selector:@selector(treeView:didSelectItem:) times:1 arguments:[_treeView, data[0]]];
    [dataSourceSpy selector:@selector(treeView:shouldDeselectItem:) times:0];
    [dataSourceSpy selector:@selector(treeView:willDeselectItem:) times:0];
    [dataSourceSpy selector:@selector(treeView:didDeselectItem:) times:0];
    [_treeView selectItem:data[0]];
    [dataSourceSpy verifyThatAllExpectationsHaveBeenMetInOrder];
    [self assert:[_treeView selectedItems] equals:[data[0]]];


    [_treeView selectItem:data[0]];
    [dataSourceSpy reset];
    [dataSourceSpy selector:@selector(selectionShouldChangeInTreeView:pendingItem:) times:1 arguments:[_treeView, nil]];
    [dataSourceSpy selector:@selector(treeView:shouldDeselectItem:) times:1 arguments:[_treeView, data[0]]];
    [dataSourceSpy selector:@selector(treeView:willDeselectItem:) times:1 arguments:[_treeView, data[0]]];
    [dataSourceSpy selector:@selector(treeView:didDeselectItem:) times:1 arguments:[_treeView, data[0]]];
    [dataSourceSpy selector:@selector(treeView:shouldSelectItem:) times:0];
    [dataSourceSpy selector:@selector(treeView:willSelectItem:) times:0];
    [dataSourceSpy selector:@selector(treeView:didSelectItem:) times:0];
    [_treeView deselectItem:data[0]];
    [dataSourceSpy verifyThatAllExpectationsHaveBeenMetInOrder];
    [self assert:[] equals:[]];

@end
```
Unfortunately, this feature is not unavailable with our current cappuccino-base (the one used by travis to build cappuccino). It means that the OJTest version installed will not be the last one but the one when 0.9.8 was released, and this one does not provide the mock :/

@aparajita, @aljungberg , @ahankinson should we update the script of travis to fetch the last version of OJTest ?

didierkorthoudt:
Maybe this fix proposal could be forgotten if you accept the new PR related to CPTrackingArea...

didierkorthoudt:
I'm closing this one as this problematic is solved by CPTrackingArea introduction.

PR #2396, opened 2015-11-10
Objective-J runtime method type accessor functions

mrcarlberg:
Added runtime functions to get return type and parameter types on methods.

The old function ```method_getTypes```was not compliant to how Objective-C runtime works. This function is now deprecated.

### DIFF
--- a/Objective-J/Runtime.js
+++ b/Objective-J/Runtime.js
@@ -1020,6 +1020,7 @@ GLOBAL(method_getName) = function(/*Method*/ aMethod)
     return aMethod.method_name;
 }
 
+// This will not return correct values if the compiler does not have the option 'IncludeTypeSignatures'
 GLOBAL(method_copyReturnType) = function(/*Method*/ aMethod)
 {
     var types = aMethod.method_types;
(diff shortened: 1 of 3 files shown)


Objective-J/Runtime.js
```diff
@@ -1020,11 +1020,26 @@ GLOBAL(method_getName) = function(/*Method*/ aMethod)
     return aMethod.method_name;
 }
 
+// FIXME: This function is deprecated and should be removed in a future release
+// Please use 'method_copyReturnType' or 'method_copyArgumentType'
 GLOBAL(method_getTypes) = function(/*Method*/ aMethod)
```
Dogild: We have jus added this method, do we need to set it as deprecated ? I'm pretty sure that not a lot of people are using it !

Objective-J/Runtime.js
```diff
@@ -1020,11 +1020,26 @@ GLOBAL(method_getName) = function(/*Method*/ aMethod)
...
     return aMethod.method_types;
 }
 
+GLOBAL(method_copyReturnType) = function(/*Method*/ aMethod)
+{
+    var types = aMethod.method_types;
+    return types ? types[0] : NULL;
+}
+
+GLOBAL(method_copyArgumentType) = function(/*Method*/ aMethod, /*unsigned int*/ index)
```
Dogild: Is it compliant with cocoa ? I believe in cocoa the first and second argument are types of the receiver and the selector

Dogild:
Any possibility to add `method_getNumberOfArguments` ? You just need to call `length` on `method.method_imp`.

Objective-J/Runtime.js
mrcarlberg: True, I will just remove it!

mrcarlberg:
I have now fixed some bugs and added test cases etc.

Objective-J/Runtime.js
```diff
@@ -1020,9 +1020,39 @@ GLOBAL(method_getName) = function(/*Method*/ aMethod)
...
+    {
+        var argType = types[0];
+
+        return argType != NULL ? argType : NULL;
+    }
+    else
+        return NULL;
+}
+
+GLOBAL(method_copyArgumentType) = function(/*Method*/ aMethod, /*unsigned int*/ index)
```
Dogild: Just wondering, do we should to follow cocoa here ? Cocoa the first index and second index are the _cmd and self.

mrcarlberg: Yes, you are right. I think we should do as in Cocoa. Both in number of arguments and the types. I'll make a change.

mrcarlberg:
It will now comply to how Objective-C runtime works

Dogild:
Thanks !

Issue #2398, opened 2015-11-12
method_getTypes does not work with Release build.


zyukin:
The same code works fine when compiling in the browser.

cappbot:
**Milestone:** Someday.  **Label:** #new.  **What's next?** A reviewer should examine this issue.

mrcarlberg:
I can confirm that this is a problem. The release version has not any type signatures for method.
We will change back this.

mrcarlberg:
A workaround that might work should be to add the option ```-T``` in your Jakefile for the release build.
Change something like ```setCompilerFlags("-O2")``` to ```setCompilerFlags("-O2 -T")```

zyukin:
With `-T` return error:
```
Error on line 11893 of file [unknown]
Error: Could not load file at file:AppController.j when compiling file:/usr/local/narwhal/packages/cappuccino/lib/cappuccino/jake/blendtask.j
```

mrcarlberg:
Can you please provide the line you changed in your Jakefile. Both how it was before and how it is now after the change.

zyukin:
I changed `task.setCompilerFlags("-O2")` to `task.setCompilerFlags("-O2 -T")`. This error reproduce if clean `Build` folder before build.

cappbot:
**Milestone:** Someday.  **Labels:** #accepted, Objective-J, bug.  **What's next?** A reviewer should examine this issue.

mrcarlberg:
The workaround does not work. I will make a PR that fixes type signatures for method definitions.

Dogild:
For people who has issues with that, there are two possibilities to have temporary fixes : 

- `setCompilerFlags("-O2 -g")` when building
- remove line 2328 in `Objective-J/ObjJAcornCompiler.js` `if (compiler.flags & ObjJAcornCompiler.Flags.IncludeDebugSymbols)`

Dogild:
@mrcarlberg How cocoa works when using the runtime methods in a release application ? Should we include typed of the method all the time in our compiled application ?

mrcarlberg:
I have been busy fixing som other important matters in out current project. I have most of it working and it will be ready in a day or two.
Yes, the default will be that type information will always be included. There will be an option to turn it off if needed.

PR #2399, opened 2015-11-20
Default include type signatures for all kind of builds

mrcarlberg:
This will make sure method and ivar type information is provided to the Objective-J runtime for all types of builds. It is done by providing the Objective-J compiler with option ```IncludeTypeSignatures```. It can optionally be turned off.

This will fix: #2398

A fix is also included that make sure that type signatures are copied to KVO method implementation for a object that is observed with KVO.

Some test cases are added for this.

### DIFF
--- a/Objective-J/CommonJS/lib/objective-j/compiler.js
+++ b/Objective-J/CommonJS/lib/objective-j/compiler.js
@@ -187,7 +187,7 @@ function resolveFlags(args)
             objjcFlags &= ~ObjectiveJ.ObjJAcornCompiler.Flags.CheckSyntax;
 
         else if (argument.indexOf("-T") === 0)
-            objjcFlags |= ObjectiveJ.ObjJAcornCompiler.Flags.IncludeTypeSignatures;
+            objjcFlags &= ~ObjectiveJ.ObjJAcornCompiler.Flags.IncludeTypeSignatures;
 
         else if (argument.indexOf("-g") === 0)
             objjcFlags |= ObjectiveJ.ObjJAcornCompiler.Flags.IncludeDebugSymbols;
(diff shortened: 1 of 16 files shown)


Objective-J/CommonJS/lib/objective-j/compiler.js
```diff
@@ -259,12 +259,16 @@ exports.main = function(args)
         if (argv[0] === "--help" || argv[0].substr(0, 1) == '-')
         {
             print("Usage (objjc 2.0): " + args[0] + " [options] [--] file...");
-            print("  -p, --print                    print the output directly to stdout");
-            print("  --unmarked                     don't tag the output with @STATIC header");
+            print("  -p, --print                         print the output directly to stdout");
+            print("  --unmarked                          don't tag the output with @STATIC header");
             print("");
-            print("  -T, --includeTypeSignatures    include type signatures in the compiled output");
+            print("  -T, --dont-include-type-signatures  include type signatures in the compiled output");
```
Dogild: Do we really want that ? What is the use case to not have type signatures ?

Dogild:
Looks good to me ! I'll test it later in the day

Objective-J/CommonJS/lib/objective-j/compiler.js
mrcarlberg: The use case is if you want to hide how your application works. Usually you minify/obfuscate the Javascript code, remove debug symbols and remove the type information. Sometimes your application need the type information and you have to leave it there.

Dogild:
Tested here and it works !

primalmotion:
Merged, Thanks!

primalmotion:
+#fixed

PR #2404, opened 2015-12-27
New: CPArray -arrayByApplyingBlock:

cacaodev:
This new method maps the receiver using the provided function.
Similar to javascript or swift map function.

There are tens of places in the fmw where using this method can reduce the amount of code.

With test.

### DIFF
--- a/Foundation/CPArray/_CPArray.j
+++ b/Foundation/CPArray/_CPArray.j
@@ -880,7 +880,11 @@ Returns a hash for the object. Unlike Cocoa, the hash value does not take conten
     return join.call([self _javaScriptArrayCopy], aString);
 }
 
-/* @ignore */
+/*!
+    Returns an Array formed by applying a function to the objects in the receiver.
+    @param aFunction a function taking two arguments: (element, index).
+    @return an Array containing the transformed elements.
+*/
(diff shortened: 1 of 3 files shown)


cacaodev:
It appears that Array.map is a lot slower than for loops. An example here : https://jsperf.com/map-vs-native-for-loop/7

As a consequence, I will remove the native map but I still believe cappuccino needs a map API because it is very common in most languages.
The naming could also be better.
What about ```-arrayByApplyingBlock:``` ? More consistent with other CPArray methods naming and express better the fact that it is an immutable method. And the array prefix tells clearly what it returns.
Of course, in js Block should be Function but we already use Block for other array methods that take a function arg.
What do you think ?

aparajita:
I agree with -arrayByApplyingBlock: as the name.

aljungberg:
I agree, `arrayByApplyingBlock` is more consistent. Methods that return something tend to be named by what they return, like `arrayByAddingObject`.

(Often we say `UsingBlock` -- `arrayByUsingBlock` doesn't sound as clear though. I like your naming, `arrayByApplyingBlock`, better.)

+#needs-improvement

aparajita:
Merci!

cacaodev:
Merged, thanks for your input.

PR #2405, opened 2015-12-29
Fixed: nib2cib on a NSBox fails when having a CPTableView in it

### LINKED ISSUE #2400: Possible nib2cib bug: null is not an object (evaluating 'self._superview._frame')
Steps to reproduce:
In the attached project
```
nib2cib Resources/CSGSecurity.xib 
```
As it stands, it should generate the `null is not an object (evaluating 'self._superview._frame')` error, and moving "bad table" out of the box should "fix" the issue and moving the table back inside the box should cause the issue to resurface.
The "bad table" is originally a clone (ALT-drag) of the "good table".

Discussion at https://groups.google.com/forum/m/#!topic/objectivej/rklSvOloYUQ

Minimal example: [IsolateBug.zip](https://github.com/cappuccino/cappuccino/files/52861/IsolateBug.zip)

Dogild:
Previously, when having a CPTableView in a CPBox, nib2cib failed due to a superview not defined yet. This was raised because in nib2cib we used the method setFrame on the CPBox to modify the frame of the CPBox (we need to modify this frame because the sizing difference between cappuccino and cocoa). We now directly modify the attribute _frame from the object and update the bounds in the same time.

This PR fixed another bug as well. Previously the contentView of the CPBox was not encoded, we now encode it. This allows us to get a full CPView object for the contentView, previously we got a weird (I have no idea how this object was created though...) object CPView with some missing attributes. This raised a crash because _trackingAreas or _themeState were not defined in this object.

Fixed #2400

### DIFF
--- a/AppKit/CPBox.j
+++ b/AppKit/CPBox.j
@@ -598,7 +598,8 @@ var CPBoxTypeKey          = @"CPBoxTypeKey",
     CPBoxBorderTypeKey    = @"CPBoxBorderTypeKey",
     CPBoxTitle            = @"CPBoxTitle",
     CPBoxTitlePosition    = @"CPBoxTitlePosition",
-    CPBoxTitleView        = @"CPBoxTitleView";
+    CPBoxTitleView        = @"CPBoxTitleView",
+    CPBoxContentView      = @"CPBoxContentView";
 
 @implementation CPBox (CPCoding)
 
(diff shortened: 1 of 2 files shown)


Dogild:
@aparajita and @primalmotion can you take a look to this one ?

primalmotion:
Looks all good

primalmotion:
+#fixed

Thanks

Issue #2406, opened 2016-01-06
CPSplitView with one view should resize its view and removes its divider


t00f:
Given a `CPSplitView` with two `CPView`, say `viewA` and `viewB`, I should be able to remove and re-add `viewB` as a subview depending on a certain context.

Cocoa removes `viewB`, resizes `viewA` so it takes the whole space available and removes the divider.

Cappuccino currently fails while trying to position the divider in method `setPosition:ofDividerAtIndex:` at line `733` because variable `frameB` is null.

``` javascript
733     preSize = frameB.size[_sizeComponent];
```

Here is a link to a complete [Cocoa example](https://dl.dropboxusercontent.com/u/19246769/TestSplitView.zip)

cappbot:
**Milestone:** Someday.  **Label:** #new.  **What's next?** A reviewer should examine this issue.

cappbot:
**Milestone:** Someday.  **Labels:** #accepted, #needs-patch, #needs-reduction, AppKit, bug.  **What's next?** 

 * This issue needs a volunteer to write and submit code to address it.
 * A minimal test app should be created which demonstrates the concern of this issue in isolation.

t00f:
Added a test that verify that this issue should no longer exist.

PR #2407, opened 2016-01-12
Test: Added a test that verifies CPSplitView behavior

t00f:
Verify `CPSplitView`  behavior when having only 1 subview.

Issue #2406 reported a behavior different from Cocoa Framework. This test validates that Cappuccino `CPSplitView` is working well when adding and removing a subview. A `CPSplitView` that has only one subview should resize it to take the whole space available.

This PR adds a new test to Tests/Manual


primalmotion:
Thanks!

+#fixed

Dogild:
This should have been an unit tests.

PR #2409, opened 2016-01-14
FIXED: CPPredicate's predicateFormat with a CPNull value as right exp…

primalmotion:
…ression wasn't working

Previously a CPPredicate created with format `value == nil` was converted back to `value == <CPNull @ xxxx>`

This patch ensures `predicateFormat` returns `value == nil`

Test added in CPPredicateTest.j


primalmotion:
I have no idea why this patch went away, and why it is still available when I access it by its URL. I'm just readding it

primalmotion:
+#fixed

PR #2410, opened 2016-01-14
FIXED: crash when opening a CPPopover opened on platform window A on …

primalmotion:
…platform window B

Previously, if a child window was opened on platform window A, then the user opened it on different platfrom window, the application would crash.

This patch contains a fix that ensure the window is correctly removed from the good CPDOMWindowLayer

### DIFF
--- a/AppKit/Platform/DOM/CPDOMWindowLayer.j
+++ b/AppKit/Platform/DOM/CPDOMWindowLayer.j
@@ -84,23 +84,35 @@
     // We will have to adjust the z-index of all windows starting at this index.
     var count = [_windows count],
         zIndex = (anIndex === CPNotFound ? count : anIndex),
-        isVisible = aWindow._isVisible;
+        isVisible = aWindow._isVisible,
+        isLocal = [_windows containsObject:aWindow];
 
-    // If the window is already a resident of this layer, remove it.
+    // if the window is visible
     if (isVisible)
     {
-        // Adjust the z-index to start at the window being inserted
-        zIndex = MIN(zIndex, aWindow._index);
+        // If the window is already a resident of this layer, remove it.
+        if (isLocal)
+        {
+            // Adjust the z-index to start at the window being inserted
+            zIndex = MIN(zIndex, aWindow._index);
 
-        // If the window being inserted is below the insertion index,
-        // the index will be one less after we remove the window below.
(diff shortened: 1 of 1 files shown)


primalmotion:
@aparajita hey, can you take a look at that?

primalmotion:
Ok so that doesn't work well, but I leave that open because it gives some clues about the problem.

This patch fixes the bug, but creates a new one when the popover closes when we click on it

Dogild:
Can you provide a reduction ! I can take a look

AppKit/Platform/DOM/CPDOMWindowLayer.j
```diff
@@ -84,23 +84,35 @@
     // We will have to adjust the z-index of all windows starting at this index.
     var count = [_windows count],
         zIndex = (anIndex === CPNotFound ? count : anIndex),
-        isVisible = aWindow._isVisible;
+        isVisible = aWindow._isVisible,
+        isLocal = [_windows containsObject:aWindow];
 
-    // If the window is already a resident of this layer, remove it.
+    // if the window is visible
```
aparajita: Redundant comment, please remove it

AppKit/Platform/DOM/CPDOMWindowLayer.j
```diff
@@ -84,23 +84,35 @@
...
+            // Adjust the z-index to start at the window being inserted
+            zIndex = MIN(zIndex, aWindow._index);
+
+            // If the window being inserted is below the insertion index,
+            // the index will be one less after we remove the window below.
+            if (aWindow._index < anIndex)
+                --anIndex;
+
+            [_windows removeObjectAtIndex:aWindow._index];
+
```
aparajita: Extra blank line, please remove it

aparajita:
@primalmotion Do you have a reduction?

mrcarlberg:
@primalmotion @Dogild What can we do to get this in a finished state?

Dogild:
I would close the PR, since it's not up to date and it did not work and open an issue to remember this problem

mrcarlberg:
@Dogild Can you please do this as I'm not familiar with this problem and will probably write some nonsense that nobody will understand.

daboe01:
+#wont-fix

Issue #236, opened 2009-08-07
mouseExited gets sent to subviews that have been removed


saikat:
To recreate this issue:
1. Add view A as a subview to view B
2. Put mouse over view A
3. Remove view A from its superview
4. When you move your mouse now, view A still gets a mouseExited event

daboe01:
probably fixed by #2395

daboe01:
@didierkorthoudt was this fixed by #2395?

didierkorthoudt:
Hi @daboe01 ,

Sorry for the delay...

I've just checked and nope... viewA received mouseExited.

I check in CPTrackingArea mechanism and get back to you...

Sorry for this.

Didier.

daboe01:
Thank you for checking. Best greetings Daniel 

Von meinem iPhone gesendet

didierkorthoudt:
I've found the problem (in the CPTrackingArea). It's caused by a running condition but only (it seems) when you use the mouseEntered event to do the view removal. In other conditions, CPTrackingArea handles this correctly.

I have to adapt my code to cope with this situation. I'll do it ASAP.

PR #2411, opened 2016-01-16
CPTrackingArea fixes

didierkorthoudt:
This PR should solve issue #236 and improves management of view hierarchy modifications done in mouse events methods (that is while processing those events). + Some other smaller updates.


didierkorthoudt:
In summary, all events that should be sent are now queued until the end of processing in order to avoid running conditions when a mouse related method is called and modifies the view hierarchy (and tracking area hierarchy) while processing it.

When adding or removing a tracking area, management stacks are modified to allow or avoid a view to receive mouse related events.
This should conform to Cocoa.

When removing a view with a tracking area, this view won't receive mouseExited.
When adding a view with a tracking area, mouse events will be sent as if mouse has entered the tracking area BUT only if CPTrackingAssumeInside is not specified.

AppKit/CPWindow/CPWindow.j
```diff
@@ -3844,59 +3854,140 @@ var interpolate = function(fromValue, toValue, progress)
...
+
+    var point    = [anEvent locationInWindow],
+        dragging = ([anEvent type] !== CPMouseMoved);
+
+    do
     {
-        // Here, we are outsite the window content view tracking area, so let _windowView set the cursor (resize cursor, ...)
+        // Initialize this run
+        
+        _restartHandlingTrackingAreaEvent = NO;
```
aparajita: Don't need this blank line.

AppKit/CPWindow/CPWindow.j
```diff
@@ -3983,132 +4076,127 @@ var interpolate = function(fromValue, toValue, progress)
...
-            }
+        // aView is not a subview of frontmostView
+        // Search in view hierarchy which one will be over the other
+        // (this is done by comparing their draw order)
 
-            // aView is not a subview of frontmostView
-            // Search in view hierarchy which one will be over the other
-            // (this is done by comparing their draw order)
+        var firstView      = frontmostView,
+        firstSuperview = [firstView superview];
```
aparajita: Missing indent

AppKit/CPWindow/CPWindow.j
```diff
@@ -3983,132 +4076,127 @@ var interpolate = function(fromValue, toValue, progress)
...
 
-            var firstView      = frontmostView,
-                firstSuperview = [firstView superview];
+        while (firstView !== _contentView)
+        {
+            var secondView      = aView,
+            secondSuperview = [secondView superview];
```
aparajita: Missing indent

aparajita:
Looks good, just need a few small formatting fixes.

didierkorthoudt:
Thank you @aparajita ! (and thank you Xcode for those missing indents...)

I have to check with @Dogild about the nib2cib matter and then I think we could be OK.

AppKit/CPWindow/CPWindow.j
```diff
@@ -3844,59 +3854,139 @@ var interpolate = function(fromValue, toValue, progress)
...
+        [_mouseEnteredStack removeObjectIdenticalTo:trackingArea];
 
-    // Cursor update
+        for (var i = 0; i < _queuedEvents.length;)
+        {
+            if ([_queuedEvents[i] trackingArea] === trackingArea)
+                [_queuedEvents removeObjectAtIndex:i];
+            else
+                i++;
+        }
```
cacaodev: You should enumerate from the end if you want to remove objects while you enumerate.

didierkorthoudt: Hi @cacaodev ! Yes, indeed (and that was my first version), but I have a bunch of (good and bad) reasons to do this... For example, I have to send events in the produced order so that mouseEntered are processed first, then mouseExited and finally cursorUpdate. This way, a mouseEntered method removing the related view will be processed in a way that next cursorUpdate event will be deleted.

cacaodev: Mhm, ok then.

didierkorthoudt: @cacaodev Oops ! Sorry for this, I didn't check where this loop was. My comment is not pertinent here. So you're right. I'll adapt. I think I had a good reason doing so, then modified the code and the remaining code doesn't need this construct anymore.

Dogild:
Can you please follow the guideline for the commit message format.

You will need to create a new pr with good commit messages for this pr. You can use the option --squash to merge every commits in one.

didierkorthoudt:
Closing this one, replaced by PR #2414

PR #2415, opened 2016-01-20
NEW: CPViewController -loadViewAsynchronously:(Function/*view, error*/)

cacaodev:
with manual test CPViewControllerTest

This is an addition compared to cocoa where resources are loaded synchronously. On the web, the preferred way is to load resources asynchronously.


AppKit/CPViewController.j
```diff
@@ -176,6 +176,58 @@ var CPViewControllerCachedCibs;
         _view = [CPView new];
 }
 
+- (void)loadViewAsynchronously:(Function/*(view, error)*/)aHandler
```
Dogild: What about `loadViewAsynchronouslyWithBlock:` ?

cacaodev: In this case, I would prefer loadViewAsynchronouslyUsingBlock: which is more in line with our current naming pattern for closures.

Dogild: Good for me !

mrcarlberg: Here I think we should using a name pattern on a bigger scale. I have used a pattern like ```loadViewWithCompletionBlock:(Function)completionBlock```. Some places uses a pattern like ```loadViewWithCompletionHandler:(Function)aHandler```. It should be good if we can come to some consensus about this. I think the word Asynchronously feels a little out of place but maybe I'm wrong...

kerusan: Semantic is everything and defines a framework. Maybe this comment is not the best place to widen the discussion, but since it is a name issue here, I'll try to explain my view. Also it is something that I am working with right now so I'll be happy to have some discussion around it.

We are using in Cappuccino a rule - that we should mimic Cocoa functionality and name standard as long as it is possible. The rule of thumb is to make the API's clear, distinct and obvious, but not to overload the semantic. And I think NeXT and Apple has excelled in this in the NS-frameworks.

But the two frameworks differs in some very basic ways and therefor we have to do some exceptions once in a while. 
In this special case the method is loading a view asynchronously and that is one of the times blocks excels in an method API. The NSViewController's `loadView` the method is not asynchronous since, and here the two framework differs. In Cocoa/C/ObjC we can use threads but since JavaScript is single-threaded, it uses asynchronous technics to achieve throughput in the execution flow.
The NSViewController's `loadView` is synchronous and does not have to have a callback but in Cappuccino we have to make it asynchronous and need a callback. If we look at the cases in Cocoa where blocks is used there is when the block is called inside the method and  when it is called after as a completion handler. (The word block is used with care since it has more than one meaning aka. also 'to block' something.)

An example of the first case is in, for instance, NSArray's `enumerateObjectsUsingBlock:` and the second case is in asynchronous cases like NSURLSession's `downloadTaskWithURL:completionHandler:`. The 'loadView' method is more of the second case and if using cocoa name standard it maybe would be nice to call it `loadViewWithCompletionHandler:`. This is clear and distinct.
To explicitly point out that it is an asynchronous call is in my opinion to overload the method name. This may advantageously be pointed out in the documentation. My view here is based on the basic differences in the two frameworks and since JavaScript relies on asynchronicity it is redundant. 

If we look forward in the Cappuccino framework's future we will soon see that node.js heavily relies on asynchronicity in it's API, here the normal case is not synchronous but asynchronous. An example is the module FileSystem where functions like `fs.exists`, `fs.link`, `fs.open`, `fs.chmod` etc. are asynchronous and the normal case, but there are other functions like `fs.existsSync`, `fs.linkSync`, `fs.openSync` and `fs.chmodSync` if a synchronous call is explicitly preferred. 

I am sorry to overload this thread with the following and it should be continued elsewhere but it shed some lights on the problem but...
I am now implementing some more classes to Foundation to be used under node.js like CPFileManager and others and there I have to extend the NS API with asynchronous methods that are equivalent, like the method: 
`- (BOOL)moveItemAtURL:(CPURL)srcURL toURL:(CPURL)dstURL error:(CPError)error`
who will have a equivalent that may be called
`- (BOOL)moveItemAtURL:(CPURL)srcURL toURL:(CPURL)dstURL errorHandler:(Function)anErrorHandler completionHandler:(Function)aCompletionHandler`
The first method uses delegation and the other function blocks to handle errors and completion.

Seeing forward to continue this discussion.

cacaodev: `loadViewWithCompletionHandler:(Function)aHandler` seems to be a winner because it is the most expressive, simple and consistent with cocoa.
If there is no objection I will rename the method and merge this PR.

mrcarlberg: Yes, I think this is how we should name it too. I will now start using it in my own code.

cacaodev:
There is something that need to be decided : what overridable methods or delegate methods the completionHandler replaces or not.
In cappuccino, we have -viewControllerWillLoadCib , -viewControllerDidLoadCib and -viewDidLoad.
In cocoa there is also:
-viewWiil/DidAppear (when the view is added to the hierarchy or unhidden)
-viewWill/DidDisappear (when the view is removed from the hierarchy or hidden)

It seems reasonable to skip -viewDidLoad because the handler can replace it. However, the async loading may also be done internally. For example when a CPTabViewItem +tabViewItemWithViewController: loads its view async. In this case the dev user may want to have access to the view after the loading ended.
For this case, I see 3 options: 

1. Add a delegate method to the API indicating that the view has loaded (example: tabView:tabViewItemDidLoadView:)
2. Keep the call to -viewDidLoad in CPViewController -loadViewWithCompletionHandler:. The dev user would need to subclass the view controller and override the -viewDidLoad method.
3. Ignore -viewDidLoad for async loading but add the 10.10 methods viewWill/DidAppear: for both loading kinds.

kerusan:
From the Apple docs: 'A Short Practical Guide to Blocks'
"Completion handlers are callbacks that allow a client to perform some action when a framework method or function completes its task. Often the client uses a completion handler to free state or update the user interface. Several framework methods let you implement completion handlers as blocks (instead of, say, delegation methods or notification handlers)."

There is a few places i Cocoa AppKit/UIKit where completionHandlers are used, and there are some different types of methods. 
Those that are framework ones and are executed by the framework but can be overridden are not applicable in this context.
Those that are both syncronous and have an async equivalent: Like NSURLSession's methods `downloadTaskWithURL:` / `downloadTaskWithURL:completionHandler:`. Here if the delegate is omitted you have to use the method with a completion handler. If `downloadTaskWithURL:` is used then the delegate is called. 
Also when NSTask have a termination handler the notification is not called.

So the rule of thumb in Cocoais, like the doc mentions, to use either delegate/notification or completions. But there are also more complex situations where some delegate methods are omitted but others are called when competion handlers are used. There are some delegate protocol/methods used in NSURLSession that will be executed in secondary threads.

In your case with loadView and its following delegate methods where there are delegate messages in Capp but not in Cocoa I think that the third alternative is the best, and then we will mimic Cocoa more. It will also help getting control when the view is displayed / shown witch we now miss.

Dogild:
Thanks

PR #2416, opened 2016-01-20
FIXED: CPBox contentView incorrect initWithCoder

didierkorthoudt:
In the initWithCoder method of CPBox, the replaceSubview was done with bad parameters order. 

The result was that the [aBox contentView] view was not in the subviews of aBox.

Fixed.


primalmotion:
Thanks for catching this! I'll merge it in as soon as the test completes

didierkorthoudt:
Please wait before merging. There's another problem in original code. I'm currently checking. I think the replaceSubview is not enough and it causes the loss of content in my own app.

didierkorthoudt:
The view hierarchy is not preserved by this. Let me some minutes to find a fix.

primalmotion:
sure let me know

primalmotion:
It would be nice to add a comment to describe what is is malformed view

And also to add a unit test for coding/decoding a CPBox and ensure it's all fine (meaning this patch is not an issue due ti Xcode only)

primalmotion:
+#fixed

Thanks!

PR #2418, opened 2016-02-03
Fixed: Allow build in directory with space in path

mrcarlberg:
Can now build Cappuccino in directory with space in the path


Dogild:
Thanks Martin

PR #2420, opened 2016-02-19
New: added first Cucumber test in Tests/CucumberTests/CPTextFieldTest/

Dogild:
This is a very small cucumber tests, it only enter few characters in a CPTextField and then fix its value


Dogild:
@primalmotion what was the trick we talked about for launching a first web server in before_script ?

daboe01:
+#wont-fix

daboe01:
i do not think that we have the bandwidth to go into that direction for the foreseeable future.

PR #2424, opened 2016-03-01
Does not throw exception if plist is not valid. It will return nil instead

mrcarlberg:
If parsing of a plist is invalid it will return nil instead of throwing an 'undefined is not an object' exception


Dogild:
Thanks :)

PR #2426, opened 2016-03-07
Fixed: //@ sourceURL=... is deprecated since many years ago.

mrcarlberg:
We have not changed it for support of older browsers.
Chrome version 51 and above writes a warning about it
in the console so it is time now.

### DIFF
--- a/Objective-J/Executable.js
+++ b/Objective-J/Executable.js
@@ -209,16 +209,16 @@ Executable.prototype.setCode = function(code)
     {
 #endif
 #if DEBUG
-    // "//@ sourceURL=" at the end lets us name our eval'd files for debuggers, etc.
+    // "//# sourceURL=" at the end lets us name our eval'd files for debuggers, etc.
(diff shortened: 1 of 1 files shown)


Dogild:
Can you give more informations about this things, I don't really know why there was this ?

Thanks

mrcarlberg:
The comment //# sourceURL=... will tell the browsers where the original file is. This helps when debugging in a browser and using a compiled language as Objective-J or CoffeeScript. We have used the old, now deprecated, syntax //@ sourceURL=... to support very old browsers.
Chrome version 51 and above will print a warning in the console if the old syntax is used. This will generate a lot of warnings. To be specific one line for each compiled file.

The very old browsers will still run the cappuccino code as it is only a comment, but it will be harder to debug with them.

You can read more about //#sourceURL at the very bottom of this page: https://developer.chrome.com/devtools/docs/javascript-debugging

Dogild:
Thanks for the informations !

Dogild:
Thanks

PR #2427, opened 2016-03-12
NEW: nib2cib support for CPButton -imageScaling.

### LINKED ISSUE #2425: CPButton Image Scaling is Not Propagated From Xib to Instantiation
Trivially reproducible. Setting the property at run-time always works.

cacaodev:
With manual test CPButtonTest.
Fixes #2425.


Tools/nib2cib/NSButton.j
```diff
@@ -439,6 +440,7 @@ var NSButtonIsBorderedMask = 0x00800000,
 
         _isBordered = (buttonFlags & NSButtonIsBorderedMask) ? YES : NO;
         _bezelStyle = (buttonFlags2 & 0x7) | ((buttonFlags2 & 0x20) >> 2);
+        _imageScaling = (((buttonFlags2 >> 6) & 0x3) + 2) % 4;
```
Dogild: Wow, are you sure about this one ? It looks like very complicated

cacaodev: I'm not happy with it neither. I know for sure that the OS X impl is ((buttonFlags2 >> 6) & 0x3) but with this I get the values 0,1,2,3 instead of 2,3,0,1. The solution is certainly simpler than that but currently i'm out  of ideas.

Dogild:
Thanks

PR #2428, opened 2016-03-13
CPSearchField: use CPMenuItem -setIndentationLevel: 

cacaodev:
… now that it is available. This removes a dirty hack.


Dogild:
Looks good to me.
Thanks

PR #2431, opened 2016-03-23
Dependent key paths use a relationship

mrcarlberg:
Only the first added observer will observe the attribute for the relationship object on a dependent relationship keypath. This PR fixes that problem.

Also, an optimisation on when the willChange… chain is running, there is no need to replace the observers as no new values are set yet. That will be done when the didChange… chain is running.

Test case is provided

### DIFF
--- a/Foundation/CPKeyValueObserving.j
+++ b/Foundation/CPKeyValueObserving.j
@@ -399,299 +399,298 @@ var kvoNewAndOld        = CPKeyValueObservingOptionNew | CPKeyValueObservingOpti
 
 - (void)_replaceModifiersForKey:(CPString)aKey
 {
-    if ([_replacedKeys containsObject:aKey] || ![_nativeClass automaticallyNotifiesObserversForKey:aKey])
-        return;
+    if (![_replacedKeys containsObject:aKey] && [_nativeClass automaticallyNotifiesObserversForKey:aKey]) {
+        [_replacedKeys addObject:aKey];
 
-    [_replacedKeys addObject:aKey];
+        var theClass = _nativeClass,
+            KVOClass = _targetObject.isa,
+            capitalizedKey = aKey.charAt(0).toUpperCase() + aKey.substring(1);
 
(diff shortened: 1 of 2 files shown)


mrcarlberg:
@Dogild You have added the label ```feature``` but I think it should be a ```bug``` as only the first key value observer will receive observations for a dependent key with a relationship (for example ```organization.name```). The others will never receive anything.

Foundation/CPKeyValueObserving.j
```diff
@@ -1300,6 +1299,10 @@ var kvoNewAndOld        = CPKeyValueObservingOptionNew | CPKeyValueObservingOpti
     if (aKeyPath === _firstPart)
```
cacaodev: In what we are seeing here, is there any significant change **_before_** this line  ?

mrcarlberg: The only change is that this part is inlined one level except for the first if statement.
Before this pull request the first if statement just returned from this method if the key was already replace. Now it will only skip the first part in the method.

Foundation/CPKeyValueObserving.j
```diff
@@ -399,299 +399,298 @@ var kvoNewAndOld        = CPKeyValueObservingOptionNew | CPKeyValueObservingOpti
 
 - (void)_replaceModifiersForKey:(CPString)aKey
 {
-    if ([_replacedKeys containsObject:aKey] || ![_nativeClass automaticallyNotifiesObserversForKey:aKey])
-        return;
-
-    [_replacedKeys addObject:aKey];
+    if (![_replacedKeys containsObject:aKey] && [_nativeClass automaticallyNotifiesObserversForKey:aKey]) {
```
cacaodev: Style: new line before the opening brace.

mrcarlberg: Thanks for finding this style error. I have now fixed it.

cacaodev:
Merged, thanks a lot !

PR #2434, opened 2016-04-04
Fixed: crash when having a CPTableViewHeader without a CPTableView

Dogild:
Previously, when having a CPTableViewHeader without a CPTableView, cappuccino just crashed.
We now check if _tableView is set before iterating on the tableview's columns.


mrcarlberg:
Works for me!

mrcarlberg:
Merged! Thanks!

PR #2436, opened 2016-04-05
Fixed: Merged in the parser and compiler that is used in the node objj-runtime

mrcarlberg:
The life will be easier to have the same in both environments.
Warnings and errors now don’t use read only properties for line and column information. So the test cases in Toolstest.j now work in modern JS engines.
It is also easier to set options for the compiler.


Dogild:
@t00f can you try this PR with the Nuage Networks UI ?

t00f:
@Dogild @mrcarlberg it looks good to me !

PR #2437, opened 2016-04-05
Fixed: Setting entity name from code does not work

mrcarlberg:
When setting entity name from code the ```_managedProxy``` will be null and the name will not be stored


Dogild:
Thanks

PR #2438, opened 2016-04-05
Fixed: First popup item is always selected if selected binding value is set before content binding value

mrcarlberg:
Test case is provided.


Dogild:
Thanks

PR #2439, opened 2016-04-08
NEW: CPViewController -viewWill/DidAppear & viewWill/didDisappear

cacaodev:
With a new ojtest CPViewControllerTest


AppKit/CPView.j
```diff
@@ -246,6 +248,9 @@ var CPViewHighDPIDrawingEnabled = YES;
 
     id                  _animator;
     CPDictionary        _animationsDictionary;
+
+    // CPViewController support
+    CPViewController _viewController @accessors(property=_viewController);
```
Dogild: I'm not convinced by this mechanism. Would it be better to have an internal mechanism (notification for instance) to notify that a view will be removed, added or hidden ?

cacaodev: I agree with you. Suprisingly cocoa has this ivar in NSView but it's better to avoid it.

AppKit/CPView.j
```diff
@@ -114,6 +115,10 @@ CPViewMaxYMargin    = 32;
 
 CPViewBoundsDidChangeNotification   = @"CPViewBoundsDidChangeNotification";
 CPViewFrameDidChangeNotification    = @"CPViewFrameDidChangeNotification";
+CPViewWillAppearNotification        = @"CPViewWillAppearNotification";
+CPViewDidAppearNotification         = @"CPViewDidAppearNotification";
+CPViewWillDisappearNotification     = @"CPViewWillDisappearNotification";
+CPViewDidDisappearNotification      = @"CPViewDidDisappearNotification";
```
Dogild: This one should be "private" like `_CPViewWillAppearNotification`

AppKit/CPView.j
```diff
@@ -1622,33 +1623,89 @@ var CPViewHighDPIDrawingEnabled = YES;
...
     }
+
+    _isHidden = aFlag;
 }
 
-- (void)_notifyViewDidHide
+- (void)_postViewWillAppearNotification
 {
-    [self viewDidHide];
+    [[CPNotificationCenter defaultCenter] postNotificationName:_CPViewWillAppearNotification object:self userInfo:nil];
```
Dogild: I'm wondering if we should loop on the subviews as well

cacaodev: It doesn't in cocoa. Makes sense to me because in -viewWillAppear, the subviews are already here. If you want to animate the subviews when the superview is added, it is still possible. You can do some setup (for example hide the subviews) and then start a animation in viewDidAppear.

AppKit/CPViewController.j
```diff
@@ -293,6 +293,8 @@ var CPViewControllerCachedCibs;
 
 - (void)_viewDidLoadWithCompletionHandler:(Function)aHandler
 {
+    [self _registerNotifications];
```
Dogild: Should you unregister for these notifications one day ? :)

cacaodev:
Splitted this PR. The first commit is now #2457.

AppKit/CPViewController.j
cacaodev: Done. Notifs are removed when the view is explicitly set, with ojtest.

Dogild:
Travis failed because the job was cancelled. Can we relaunch it ?

cacaodev:
Yes please.

Dogild:
Close and reopen the PR to relaunch the tests

PR #2443, opened 2016-04-15
NEW: CPCollectionView animations

cacaodev:
This PR adds built-in animations when adding/removing/reordering CPCollectionView items.

The animation duration for both fade in/out and moving items is currently 0.6s. Your feeling about what the right duration should be for these animations (can be different) is welcome.

The animated behavior is not currently opt-in but it could be. Perhaps a flag in CPControl ?

### DIFF
--- a/AppKit/CPCollectionView.j
+++ b/AppKit/CPCollectionView.j
@@ -31,6 +31,9 @@
 @import "CPDragServer_Constants.j"
 @import "CPPasteboard.j"
 @import "CPView.j"
+@import "CPAnimationContext.j"
+@import "CPKeyValueBinding.j"
+@import "CPDragServer_Constants.j"
 
 @class _CPCollectionViewDropIndicator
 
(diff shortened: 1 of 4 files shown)


Dogild:
Hi Cacaodev, it looks like the PR has more than `NEW: CPCollectionView animations`. Could you explain in details what other changes you did in the `CPCollectionView` ?

Thanks,

cacaodev:
Hi @Dogild , in the first commit there are two changes:

- Animating a collection based control means that we need to transform a model change into insertions/deletions/rearrangement because this is what is animated. The setContent: method have been rewritten for this purpose: we compare new/old content and extract info about items that need to be added or deleted. Then, the layout pass (displayItems: method) takes that info and runs different animations depending on the operation.

- While reviewing this code (I wrote some time ago), i noticed that there is also a change in the caching system , the -newItemForRepresentedObject: method. I don't think it's related or needed for animations. If not, I will take out the code from this PR and test it separately.

Dogild:
Yep, it's better with two pr

cacaodev:
Closing this one for now. I will reopen when some needed changes in collection view and CPAnimationContext are merged.

PR #2447, opened 2016-05-06
CPAnimationContext fixes

cacaodev:
Various fixes for CPAnimationContext and CPView animator.

With cucapp test, testing commits #1,#2,#3 of this PR.


AppKit/CPView.j
```diff
@@ -558,7 +560,7 @@ var CPViewHighDPIDrawingEnabled = YES;
 #endif
 
     // Notify the subview that it will be moving.
-    [aSubview viewWillMoveToSuperview:self];
+    [aSubviewOrViewAnimator viewWillMoveToSuperview:self];
```
Dogild: Are you sure about that ?

AppKit/CoreAnimation/CPViewAnimator.j
```diff
@@ -8,11 +8,14 @@
...
-
-    if (orderInAnim && [orderInAnim isKindOfClass:[CAPropertyAnimation class]])
+    if (aSuperview !== nil)
     {
-        [_target setValue:[orderInAnim fromValue] forKeyPath:[orderInAnim keyPath]];
+        var orderInAnim = [self animationForKey:@"CPAnimationTriggerOrderIn"];
+
+        if (orderInAnim && [orderInAnim isKindOfClass:[CAPropertyAnimation class]])
+        {
+            [_target setValue:[orderInAnim fromValue] forKeyPath:[orderInAnim keyPath]];
```
Dogild: Style here, you don't need the brackets

AppKit/CPView.j
cacaodev: You mean sure it should be supported to add an animator, or the way it is implemented. For me it seems logical that the special event keypath CPAnimationTriggerOrderIn is invoked when the animator is added to the superview : [view addSubview:animator] . Although, to be honest, this is not working in cocoa (it crashes) but I can't say what is the right way to reach this key path because it seems it has been silently deprecated. Or I missed something.

For the implementation of this feature, I can think of a more elegant way to do it: the animator could register for viewWill/DidAppear notifications sent by its target (the actual view) and perform the animation there.

Dogild:
@cacaodev Conflicts need to be fixed :)

cacaodev:
I'm i favor of removing the CPAnimationTriggerOrderIn feature (appearing animation) and put it back when I figure out how it works in cocoa, why if it has been deprecated, and what replaces it there. Don't you think ?

Tests/CucumberTests/CPAnimationContextTest/AppController.j
```diff
@@ -0,0 +1,233 @@
...
+    [fadeIn setFromValue:0];
+    [fadeIn setToValue:1];
+    [fadeIn setDuration:ANIMATION_DURATION];
+
+    var fadeOut = [CABasicAnimation animationWithKeyPath:@"alphaValue"];
+    [fadeOut setFromValue:1];
+    [fadeOut setToValue:0];
+    [fadeOut setDuration:ANIMATION_DURATION];
+
+    [vanillaView setAnimations:@{@"CPAnimationTriggerOrderIn":fadeIn, @"CPAnimationTriggerOrderOut":fadeOut}];
```
Dogild: Can you update the test, since you put CPAnimationTriggerOrderIn as deprecated

Dogild:
If you feel ready to merge it @cacaodev, do it ;)

PR #2448, opened 2016-05-09
Fixed: CPViewController isViewLoaded property in async mode

cacaodev:
The isViewLoaded property was incorrect when using the -loadViewWithCompletionHandler: method.
With cucapp test.


Dogild:
Thanks, I really need to finish the integration of cucapp and travis !

PR #2451, opened 2016-05-16
Fixed: edits in cell based tableviews can get lost sometimes

### LINKED ISSUE #2450: edits in cell based tableviews get lost when clicking on a dataview
you can see this e.g. in Manual TableTest/EditingControls (see attachment).
bildschirmfoto 2016-05-14 um 17 03 12

daboe01:
previously, the ivars _editingColumn and _editingRow were changed during editing a text cell inside cell based tableviews when a non-text dataView was clicked on. This usually resulted in lost or corrupt data.

This PR makes sure that the edit is committed before the event for the other edit is processed.

fixes #2450

### DIFF
--- a/AppKit/CPTableView.j
+++ b/AppKit/CPTableView.j
@@ -5089,6 +5089,17 @@ Your delegate can implement this method to avoid subclassing the tableview to ad
     }
     else if (!_isViewBased && [aView isKindOfClass:[CPControl class]] && ![aView isKindOfClass:[CPTextField class]])
     {
+        // we are currently in an edit session but the user clicked on a non-text control
+        if (_editingColumn !== -1)
+        {
+            var currentFirstResponder = [[CPApp keyWindow] firstResponder];
+            // make sure the currently edited control is commited
+            [self _commitDataViewObjectValue:currentFirstResponder];
+            // resend the same event in next run loop iteration
+            [[CPRunLoop currentRunLoop] performSelector:@selector(sendEvent:) target:CPApp argument:[CPApp currentEvent] order:0 modes:[CPDefaultRunLoopMode]];
+            return currentFirstResponder;
+        }
+
         [self getColumn:@ref(column) row:@ref(row) forView:aView];
 
         _editingColumn = column;


AppKit/CPTableView.j
```diff
@@ -5098,6 +5098,17 @@ Your delegate can implement this method to avoid subclassing the tableview to ad
     }
     else if (!_isViewBased && [aView isKindOfClass:[CPControl class]] && ![aView isKindOfClass:[CPTextField class]])
     {
+        // we are currently in an edit session but the user clicked on a non-text control
+        if (_editingColumn >= 0)
```
cacaodev: _editingColumn != -1 is equivalent but would me more explicit.

AppKit/CPTableView.j
```diff
@@ -5098,6 +5098,17 @@ Your delegate can implement this method to avoid subclassing the tableview to ad
...
+        {
+            var fr = [[CPApp keyWindow] firstResponder];
+            // make sure the currently edited control is commited
+            [self _commitDataViewObjectValue:fr];
+            // resend the same event in next run loop iteration
+            [[CPRunLoop currentRunLoop] performSelector:@selector(sendEvent:) target:CPApp argument:[CPApp currentEvent] order:0 modes:[CPDefaultRunLoopMode]];
```
cacaodev: If you want to perform all selectors currently trapped in the run loop (is sendEvent:  one of them ?) , you can do `[currentLoop performSelectors]`

Dogild:
Hi @daboe01, can you change the commits message and follow the cappuccino commit guideline please

AppKit/CPTableView.j
daboe01: @cacaodev i just want to resent the current event a second time _after_ the currently edited data are committed. I use [CPApp sendEvent:] for this.

AppKit/CPTableView.j
```diff
@@ -5089,6 +5089,17 @@ Your delegate can implement this method to avoid subclassing the tableview to ad
...
     else if (!_isViewBased && [aView isKindOfClass:[CPControl class]] && ![aView isKindOfClass:[CPTextField class]])
     {
+        // we are currently in an edit session but the user clicked on a non-text control
+        if (_editingColumn !== -1)
+        {
+            var currentFirstResponder = [[CPApp keyWindow] firstResponder];
+            // make sure the currently edited control is commited
+            [self _commitDataViewObjectValue:currentFirstResponder];
+            // resend the same event in next run loop iteration
+            [[CPRunLoop currentRunLoop] performSelector:@selector(sendEvent:) target:CPApp argument:[CPApp currentEvent] order:0 modes:[CPDefaultRunLoopMode]];
```
Dogild: @cacaodev Should we send that to the entire CPApp ? I would only send that to the next responder, what do you think ?

daboe01: @Dogild @cacaodev  Sending to the tableView's next responder would not work. IMHO re-sending the event in the most transparent fashion is the way to go.

Dogild: @daboe01 what is the reason that it would not work ?

daboe01:
@Dogild from my testing, it will not work to just send the event to the nextResponder. This is probably because _editingColumn and and _editingRow are not correctly initialized yet at this point in time.  The easiest way for me to make it work, was to resend the same event in the next iteration of the runloop.

daboe01:
-#needs-improvement

daboe01:
@Dogild any objections still?

Dogild:
I disagree about sending the event to the `CPApp` again. IMO we should send the event manually to the hit view got in parameter.

daboe01:
Ok but it has to happen in the next iteration of the run loop to work.

Von meinem iPhone gesendet

daboe01:
@Dogild 
this does not work (just tried):
```objj
           // [[CPRunLoop currentRunLoop] performSelector:@selector(sendEvent:) target:CPApp argument:[CPApp currentEvent] order:0 modes:[CPDefaultRunLoopMode]]; // this works
            [[CPRunLoop currentRunLoop] performSelector:@selector(_hitTest:) target:self argument:aView order:0 modes:[CPDefaultRunLoopMode]]; // this does not work
```
the reason is that event has to be sent to the _tableview_ because of the side effects that happen from the mouse event (i.e. updating _editingColumn and so on).

daboe01:
@Dogild the current master corrupts your model if you use cell based tableviews with dataviews other than textfields. _your data ultimately ends up in the wrong columns and even rows._

i see that my hotfix is a bit awkward. however, the issue is so serious that i would recommend to merge this in right now and wait for a cleaner solution in the long run.

mrcarlberg:
@daboe01: I have a question why you try to use the selector `hitTest:` above? You wrote that it does not work but from what I understand it will just find the view?

If you check how it is done in `sendEvent:` in CPWindow.j. Here is the code part where it handles a mouse down event:
``` objective-c
        case CPLeftMouseDown:
        case CPRightMouseDown:
            // This will return _windowView if it is within a resize region
            _leftMouseDownView = [_windowView hitTest:point];

            if (_leftMouseDownView !== _firstResponder && [_leftMouseDownView acceptsFirstResponder])
                [self makeFirstResponder:_leftMouseDownView];

            var keyWindow = [CPApp keyWindow];

            // This is only when we move from a platform to another one
            if ([keyWindow platformWindow] != [self platformWindow])
                [self makeKeyAndOrderFront:self];

            [CPApp activateIgnoringOtherApps:YES];

            var theWindow = [anEvent window],
                selector = type == CPRightMouseDown ? @selector(rightMouseDown:) : @selector(mouseDown:);

            if ([theWindow isKeyWindow] || ([theWindow becomesKeyOnlyIfNeeded] && ![_leftMouseDownView needsPanelToBecomeKey]))
                return [_leftMouseDownView performSelector:selector withObject:anEvent];
            else
            {
                // FIXME: delayed ordering?
                [self makeKeyAndOrderFront:self];

                if ([_leftMouseDownView acceptsFirstMouse:anEvent])
                    return [_leftMouseDownView performSelector:selector withObject:anEvent];
            }
            break;
```
As you can see above it first finds the view with `hitTest:` then performs the selector `mouseDown:` on the view. There are a lot of "if this" and "if that" but that is basically what happens. When you tried with `hitTest:` above it will not do the `mouseDown:` part.

Why do you need to do it in the next run loop? Can't it work if you just commit the changes and then allow it continue down in the `_hitTest:` method? The `_editingColumn` and `_editingRow` properties will be set at the end of the method to the new values and it will return the hit view. After that in the `sendEvent:` method it should just perform the `mouseDown:` selector on the hit view.

Should look like something like this:
``` objective-c
    else if (!_isViewBased && [aView isKindOfClass:[CPControl class]] && ![aView isKindOfClass:[CPTextField class]])
    {
        // we are currently in an edit session but the user clicked on a non-text control
        if (_editingColumn !== -1)
        {
            var currentFirstResponder = [[CPApp keyWindow] firstResponder];
            // make sure the currently edited control is commited
            [self _commitDataViewObjectValue:currentFirstResponder];
        }

        [self getColumn:@ref(column) row:@ref(row) forView:aView];

        _editingColumn = column;
        _editingRow = row;

        [aView addObserver:self forKeyPath:@"objectValue" options:CPKeyValueObservingOptionOld | CPKeyValueObservingOptionNew context:"editing"];
    }
```

I have not tried this myself so I can be totally wrong 😄 

Also maybe @cacaodev has some input on this problem?

daboe01:
@mrcarlberg this is what i tried in the first place.  this does not work (i just tried it again). this is most likely because of side effects from _commitDataViewObjectValue:

in my testing we really need to resend the event.
alternatively, we would possibly need a major refactoring of CPTableView.

mrcarlberg:
I just did a quick look how this works and I feel that the main problem is that there are side effects in the `hitTest:` method for the `CPTableView`. The best should be if we didn't do anything in there except returning the correct view.

We can't set `_editingColumn` and `_editingRow` in that method. When the firstResponder is later resigned it will commit the value to the wrong place.

Your fix work as you commit the value and return the firstResponder. As your return the firstResponder the framework thinks we clicked the on the firstResponder and will not resign it (and not commit anything).

I think we need to fix this problem but we can't do it like this. We must get info from @cacaodev why this behaviour exists.

Why can't all this be handled in `_firstResponderDidChange` or somewhere else?

cacaodev:
The _hitTest: method is a little confusing because it handles 2 different behaviors: the first part is for the "edit data view on first click if the row is selected" feature in view-based tv, instead of double click in cell-based. I believe hitTest is the right place to do that.
The second part, after the else, is for cell-based table-views and it seems to be the problematic part. It handles editing (in the table view context) on controls that can change their value but do not accept first responder like checkboxes. Obviously a guard is missing here because if you are already in an edit session in a text field, the _editingColumn corresponding to the textfield will get the value of the clicked control.
To answer the question, If we can handle cell-based buttons elsewhere I think we should do it. I just don't have any idea right now.
And just to be clear, we can't have both the text field edit session staying live (even when scrolling) AND the button changing its value on first click. We need to choose to either commit the text field value or discard the button value change.

daboe01:
@cacaodev do you think my patch is acceptable as a temporary 'hotfix'?

At least it has a perfect user experience and no risk of introducing new bugs from major code refactoring.

cacaodev:
We need to fix the wrong values for _editingColumn/Row. Have you try @mrcarlberg proposal ?

daboe01:
@cacaodev fixing the wrong values for _editingColumn/Row is only half of the problem.
we also need to commit the text and at the same time forward the event to the dataview.
otherwise, the user would need to click twice. this would be annoying and unlike in cocoa.

and to answer your question: @mrcarlberg code from his post yesterday does not work.

mrcarlberg:
I have just tested how it works in Cocoa with one cell with a textfield and one cell with a checkbox. If I start edit the textfield and click on the checkbox it changes state and the textfield is still the first responder and I can continue edit the textfield as before I clicked on the checkbox.

@cacaodev This must be the behaviour we want to have but above you say that "We need to choose to either commit the text field value or discard the button value change". Are there any special reasons for this?

@daboe01 Does your hot fix have this behaviour?

cacaodev:
@mrcarlberg No special reason, I just consider it is preferable to fix the bug (the data scrambling) now even if the behavior doesn't match exactly cocoa.

daboe01:
@mrcarlberg my hotfix makes the following when the user clicks a non-text dataview:
1. the edit session commits
2. this automatically blurs the text field
3. the click on the dataview is executed

This is what i intuitively would expect. i know that this was the behaviour in previous macos versions. i am surprised that this is not the default behaviour of the current cocoa implementation anymore. i feel that keeping the edit session is confusing. when i click on a separate field i am normally finished with the editing and want the data committed. this is not the case when the editing session is ongoing.

daboe01:
one further thought on the blur behaviour: keeping the edit session in case of clicking on a control would be completely alien to the web platform. I do not know any web-application with such behaviour.
we should keep this in mind because otherwise cappuccino apps could not be used in the uncontrolled web with uncontrolled and untrained users.

mrcarlberg:
This has been discussed before and Cappuccino should mimic the Cocoa behaviour. Outside a table view we have the behaviour as Cocoa. I think it should be the same inside a table view too.

mrcarlberg:
Can the solution be that we change the method `- (void)_commitDataViewObjectValue:(id)aDataView` to something like `- (void)_commitDataViewObjectValue:(id)aDataView forColumn:(CPInteger)column andRow:(CPInteger)row`

This method is only used from two places in `CPTabelView` and we need to pass the row and column to the commit method. We have to look up the row and column in the `observeValueForKeyPath:` method but that could maybe be done with the `getColumn:row:forView:` method. The second method should just use `_editingRow` and `_editingColumn`

Then we don't need to change the `_editingRow` and `_editingColumn` in the `_hitTest:` method. The hot fix is not needed and there are no corrupted data.

This separate the `_commitDataViewObjectValue:` method from `_editingRow` and `_editingColumn` and we can commit data on cells separate from the one that is edited at the moment.

@cacaodev @daboe01 Can this work? I have of course not tested it so I might be wrong.... 😄

daboe01:
i am convinced. i will try your approach this weekend.

daboe01:
it looks like @mrcarlberg 's approach is actually working.
i am closing this one. i will open a new one that follows martin's approach in the future.

daboe01:
the new PR is #2467

PR #2455, opened 2016-05-18
Fixed: CPColorPanel always opens in main platform window.

### LINKED ISSUE #2387: CPColorWell in another platform window
When putting a CPColorWell in another platform window, the color panel opens in the main window instead of in the platform window.

This can be observed in the following screenshot :

cpcolorwell_bug

didierkorthoudt:
When using a CPColorWell on a platform window other than the main platform window, the CPColorPanel was always opened on the main platform window.

This PR forces the CPColorWell to set the platform window of the CPColorPanel to its own platform window.

fixes #2387


Dogild:
Thanks

PR #2456, opened 2016-05-22
Fixed: right mouse clic during a drag operation cancels it

### LINKED ISSUE #2378: Right mouse clic during a drag operation problem
In a CPControl subclass, I've implemented dragging via startTrackingAt, continueTracking and stopTracking.

Everything works well except that if, during a drag, I also click on the right mouse button (ok, that's a silly idea but on a two buttons mouse this can be done by mistake), the drag stops without any notification (no call, for example, to stopTracking).

To block this, I've tested many things like overwriting the trackMouse method and add CPRightMouseDownMask to the event catcher ; or add a new method like disableRightMouse that catch (with dequeue) all right mouse events (down, up, dragged).

But the problem remains : I get events, that's OK, but they always stop my drag operation, even if they are dequeued.

Here is a simple sample app showing the problem : just drag the red square or the window and, during this, also right clic and baam! it stops.

This has been test under Safari, Chrome, Firefox, Chrome/win, Firefox/win, IE/win

/*
 * AppController.j
 * RightMouseClicCatch
 *
 * Created by You on September 20, 2015.
 * Copyright 2015, Your Company All rights reserved.
 */

@import <AppKit/AppKit.j>

@class DKDragView;

@implementation AppController : CPObject
{
    CPWindow    myWindow;
    DKDragView  dragView;
}

- (void)applicationDidFinishLaunching:(CPNotification)aNotification
{
    // This is called when the application is done loading.
}

- (void)awakeFromCib
{
    // This is called when the cib is done loading.
    // You can implement this method on any object instantiated from a Cib.
    // It's a useful hook for setting up current UI values, and other things.

    // In this case, we want the window from Cib to become our full browser window
//    [theWindow setFullPlatformWindow:NO];
    
    myWindow = [[CPWindow alloc] initWithContentRect:CGRectMake(50, 50, 500, 500) styleMask:CPTitledWindowMask | CPResizableWindowMask | CPClosableWindowMask];
    
    dragView = [[DKDragView alloc] initWithFrame:CGRectMake(10, 10, 200, 200)];
    
    [[myWindow contentView] addSubview:dragView];

    [myWindow orderFront:self];
}

@end

@implementation DKDragView : CPControl
{
    BOOL    _rightMouseIsDisabled   @accessors(getter=rightMouseIsDisabled);
}

- (id)initWithFrame:(CGRect)aRect
{
    self = [super initWithFrame:aRect];
    
    if (self)
    {
        [self setBackgroundColor:[CPColor redColor]];
    }
    
    return self;
}

- (BOOL)startTrackingAt:(CGPoint)aPoint
{
    [self disableRightMouse];
    
    return YES;
}

- (BOOL)continueTracking:(CGPoint)lastPoint at:(CGPoint)aPoint
{
    [self setCenter:[[self superview] convertPoint:aPoint fromView:self]];
    
    return YES;
}

- (void)stopTracking:(CGPoint)lastPoint at:(CGPoint)aPoint mouseIsUp:(BOOL)mouseIsUp
{
    [self enableRightMouse];
}

// Right mouse catcher

- (void)disableRightMouse
{
    CPLog.trace("disableRightMouse");
    
    _rightMouseIsDisabled = YES;
    
    [self _trapNextRightMouse];
}

- (void)enableRightMouse
{
    CPLog.trace("enableRightMouse");
    
    _rightMouseIsDisabled = NO;
}

- (void)_trapNextRightMouse
{
    [CPApp setTarget:self selector:@selector(_rightMouseWasClicked:) forNextEventMatchingMask:CPRightMouseDownMask | CPRightMouseUpMask | CPRightMouseDraggedMask untilDate:nil inMode:CPDefaultRunLoopMode dequeue:YES];
}

- (void)_rightMouseWasClicked:(CPEvent)anEvent
{
    CPLog.trace("_rightMouseWasClicked (disabled:"+_rightMouseIsDisabled+")");
    
    if (_rightMouseIsDisabled)
    {
        [[[anEvent window] platformWindow] _propagateContextMenuDOMEvent:NO];
        
        [self _trapNextRightMouse];
    }
    else
        
        // As we catched the right mouse event, we must post it again
        
        [CPApp sendEvent:anEvent];
}

@end

This fails even if I add :

        [[[anEvent window] platformWindow] _propagateCurrentDOMEvent:NO];

didierkorthoudt:
While dragging a window or a custom view, if the user also pushes on the right mouse button (by mistake), the drag operation is stopped (without any notification to the application).

This PR modifies the DOM event processing so that the behavior is now compatible with the Cocoa way.

fixes #2378

### DIFF
--- a/AppKit/Platform/CPPlatformWindow.j
+++ b/AppKit/Platform/CPPlatformWindow.j
@@ -60,6 +60,7 @@ var PrimaryPlatformWindow   = NULL;
 
     BOOL                    _mouseIsDown;
     BOOL                    _mouseDownIsRightClick;
+    int                     _firstMouseDownButton;
     CGPoint                 _lastMouseEventLocation;
     CPWindow                _mouseDownWindow;
     CPTimeInterval          _lastMouseUp;
--- a/AppKit/Platform/DOM/CPPlatformWindow+DOM.j
+++ b/AppKit/Platform/DOM/CPPlatformWindow+DOM.j
@@ -1245,12 +1245,16 @@ _CPPlatformWindowWillCloseNotification = @"_CPPlatformWindowWillCloseNotificatio
     {
         if (_mouseIsDown)
         {
+            if (aDOMEvent.button !== _firstMouseDownButton)
+                return;
+
             event = _CPEventFromNativeMouseEvent(aDOMEvent, _mouseDownIsRightClick ? CPRightMouseUp : CPLeftMouseUp, location, modifierFlags, timestamp, windowNumber, nil, -1, CPDOMEventGetClickCount(_lastMouseUp, timestamp, location), 0, nil);
 
             _mouseIsDown = NO;
             _lastMouseUp = event;
             _mouseDownWindow = nil;
             _mouseDownIsRightClick = NO;
+            _firstMouseDownButton = -1;
         }
 
         if (_DOMEventMode)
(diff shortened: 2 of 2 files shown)


Dogild:
Looks good to me and it its conform to cocoa. Thanks

PR #2457, opened 2016-05-23
FIXED: CPView received multiple -viewDidHide/Unhide messages

cacaodev:
Also Fixed: when a view is hidden by a hidden superview and added to a non
hidden superview, the -viewDidUnhide message is sent.

With test : CPViewTest.j

### DIFF
--- a/AppKit/CPView.j
+++ b/AppKit/CPView.j
@@ -172,6 +172,7 @@ var CPViewHighDPIDrawingEnabled = YES;
     CPArray             _registeredDraggedTypesArray;
 
     BOOL                _isHidden;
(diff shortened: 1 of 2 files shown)


Dogild:
Thanks

enquora:
Merge f1145ef for this patch breaks scroller functionality in cptableview and cppopubutton/cpmenu on my installation. Scrollview not resizing correctly

daboe01:
@enquora @Dogild @cacaodev  i also see this.

Dogild:
@cacaodev do you have a fix for the regression ? or should we revert the PR ?

PR #2459, opened 2016-05-26
CPCollectionView setContent: and layout changes

cacaodev:
Before this commit, changing the content from ["A","B"] to ["B","C"] would remove the views "A" and "B" and then add "B" and "C".
After this commit, such change will result in removing A, moving B, adding C.

The perf is very similar to the previous implementation (see testSetContentPerfTest). This change is a necessary for collection view items animation.


daboe01:
i do not see any issues with this PR.

-#new
+#ready-to-commit

Dogild:
Hi @cacaodev, I'm wondering if we should put some effort to be more compliant with the `NSCollectionView`. It looks like they have added a dataSource for this class since 10.11 (https://developer.apple.com/library/mac/documentation/AppKit/Reference/NSCollectionViewDataSource_protocol/index.html#//apple_ref/occ/intf/NSCollectionViewDataSource).

What do you think ?

cacaodev:
@Dogild This new API/model is definitely interesting in terms of usability and performance. We can do it later, starting with a) the "reusable items" feature (& data source) and then maybe b) the new "section" stuff where items are referenced by their IndexPath.

The main challenge for a) will be to enqueue items at the appropriate moment, ie when they become invisibles.

cacaodev:
It appears that the new 10.11 features are incompatible (or very difficult to implement) with the optimization contained in this PR (commit 3).
The 10.11 features have a new data source model and reusable items by identifier. It means lazy content loading and lazy view display which is very interesting for perf and usability.
I will extract some commits from this one and close it.

PR #2461, opened 2016-06-18
FIXED: CPView received multiple -viewDidHide/Unhide messages

cacaodev:
Also Fixed: when a view is hidden by a hidden superview and added to a non hidden superview, the -viewDidUnhide message is now sent.

With test in CPViewTest.j


daboe01:
-#new
+#needs-review
+AppKit

Dogild:
Thanks

PR #2462, opened 2016-06-21
fixed: cappuccino pasteboad is always clobbered by system pasteboard

### LINKED ISSUE #2062: Pasteboard types other than CPStringPboardType are effectively unsupported
It seems that cappuccino always uses the native pasteboard when available (chrome/ff) and then only return the CPStringPboardType

daboe01:
previously, every native nativePasteEvent: compared the pasted content to [pasteboard _stateUID], which is always undefined in the current implementation of CPPasteboard. This resulted in a call to pasteboard declareTypes:[CPStringPboardType] owner:self, effectively removing all other representations during each paste.

This PR makes capp to only call [CPStringPboardType] owner:self if the system and capp CPStringPboardType actually differ. This makes the system and capp pasteboards cooperate. This is a prerequisite for proper copy/paste support of the new text system.

Fixes #2062 and
https://github.com/daboe01/CPTextView/issues/109

### DIFF
--- a/AppKit/Platform/DOM/CPPlatformPasteboard.j
+++ b/AppKit/Platform/DOM/CPPlatformPasteboard.j
@@ -450,9 +450,10 @@ Return true if the event may be a copy and paste event, but the target is not an
 
     if ([value length])
     {
-        var pasteboard = [CPPasteboard generalPasteboard];
+        var pasteboard = [CPPasteboard generalPasteboard],
+            cappString = [pasteboard stringForType:CPStringPboardType];
 
-        if ([pasteboard _stateUID] != value)
+        if (cappString !== value)
         {
             [pasteboard declareTypes:[CPStringPboardType] owner:self];
             [pasteboard setString:value forType:CPStringPboardType];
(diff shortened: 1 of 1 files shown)


daboe01:
-#new
+#needs-review

daboe01:
+AppKit

daboe01:
@aljungberg can you take a look please?

daboe01:
@aljungberg are you still around?

aljungberg:
Hmm but if nothing was pasted into `_DOMPasteboardElement`, then `value.length` would be 0 and nothing would be written to the pasteboard.

And it the user did paste a string, wouldn't we expect the pasteboard to only have a string in it (and remove all other representations)?

daboe01:
@aljungberg please consider this use case:
the user has plain and rich representations of the same text on the pasteboard (this can happen with the new CPTextView, see discussion in this issue: https://github.com/daboe01/CPTextView/issues/109).

the user decides to paste twice, starting with a plain field and then into a rich text view. Without this patch, the first paste will remove the rich representation of the pasteboard and the user will never get the rich text pasted in.
With this patch, the rich representation is pasted into the rich text view as expected.

daboe01:
from looking into _generateStateUID from CPPasteboard.j, comparing _stateUID and value does not make sense anyway.

mrcarlberg:
Merged! Thanks!

If this PR creates problems we have to fix them later. It is needed for #2068.

PR #2464, opened 2016-07-08
Remove nativehost

daboe01:
previously, the repository contained nativehost, an obsoleted technology that we decided to remove.
Nativehost is removed by this PR


daboe01:
i recommend this if you really need to get native on mac:
http://sveinbjorn.org/platypus

Dogild:
Thanks

daboe01:
if you need to go native cross plattform, look at http://electron.atom.io/

PR #2467, opened 2016-07-24
Fixed: tableview corruption when clicking on a cell-based control during an editing session

### LINKED ISSUE #2450: edits in cell based tableviews get lost when clicking on a dataview

daboe01:
previously, the ivars _editingColumn and _editingRow were changed as a side effect of hitTest:. This resulted in model corruption when clicking on a cell-based non-text control during an editing session.

This PR makes sure that edited textfields stay in focus even after committing a click on a cell-based control.

fixes #2450


Dogild:
Looks good to me. @cacaodev @mrcarlberg can you confirm ?

Dogild:
Can we get some unit tests for this one ? :)

daboe01:
@Dogild how can we unit-test this without the DOM? Keep in mind that CPTextFields are involved.

mrcarlberg:
I think it should be possible to create unit test cases for this. We don't need the DOM for mosts test. It should be possible to create a table view with for example a checkbox and a text field. Place the views inside a ```CPWindow```. Make the text field first responder and simulate a click on the checkbox by sending an event to the ```CPWindow```. Check that the checkbox has changed and the first responder is still the text field.

It might take some time to code the setup of all the views but I think it is possible. It should really help if you have the time to do it.

mrcarlberg:
Other than the need for test cases I think this looks ok. @cacaodev Do you have any comments on this as you are very familiar with the ```CPTextView```

Dogild:
+1 @mrcarlberg. Except of sending the event to the `CPWindow` you should send it to the `CPApp`.

daboe01:
@mrcarlberg thank you for the hints towards writing a unit test. unfortunately, i will not have time for this in the foreseeable future.

mrcarlberg:
@daboe01 I understand and I think we can merge this pull request without a test case like that as we have the manual test in TableTest/EditingControls.
I'm ready to merge this but I would like to get feedback from @cacaodev about if this looks ok as he has a lot of knowledge about ```CPTableView```.

cacaodev:
Sorry for the delay, I wasn't around these days. I am completely ok with this commit and I thank @daboe01 for fixing my code.

mrcarlberg:
Merge! Thanks!

PR #2469, opened 2016-08-04
Fixed: jerky scrolling by momentum in Chrome > 51

### LINKED ISSUE #2468: Scrolling by momentum is broken in Chrome 52 on slower maschines
you can see this e.g. in the manual DragAndDrop TableTest

daboe01:
Previously, the scrolling events were generated in a setTimeout with a delay of 0ms.
These were not reliably executed in recent chrome versions which caused freezes during scrolling with momentum on older systems such as a MBP late 2011.
This PR removes the setTimeout and uncomments the limitDateForMode:CPDefaultRunLoopMode to enforce the visual updates.
Tested on FF, Safari and of course chrome.
Fixes #2468


AppKit/Platform/DOM/CPPlatformWindow+DOM.j
```diff
@@ -952,50 +952,44 @@ _CPPlatformWindowWillCloseNotification = @"_CPPlatformWindowWillCloseNotificatio
                                   timestamp:timestamp windowNumber:windowNumber context:nil eventNumber:-1 clickCount:1 pressure:0];
     event._DOMEvent = aDOMEvent;
 
-    // We lag 1 event behind without this timeout.
-    setTimeout(function()
```
Dogild: Would it not work with `[CPRunLoop -performBlock]` ?

daboe01:
@Dogild just tried performBlock. same issue as with setTimeout

AppKit/Platform/DOM/CPPlatformWindow+DOM.j
```diff
@@ -952,50 +952,44 @@ _CPPlatformWindowWillCloseNotification = @"_CPPlatformWindowWillCloseNotificatio
                                   timestamp:timestamp windowNumber:windowNumber context:nil eventNumber:-1 clickCount:1 pressure:0];
     event._DOMEvent = aDOMEvent;
 
-    // We lag 1 event behind without this timeout.
```
Dogild: Does someone have an idea why we did this like this ?

daboe01:
i dont know but i do not see any event lagging with this PR

daboe01:
@Dogild would't it make sense to assign @aljungberg?
i remember that he reworked scroll-wheel scrolling some time ago.

mrcarlberg:
I have a vague memory about a "1 event behind" discussion some years ago. Can someone ping @aljungberg on all possible channels as he don't check emails. 😄 He is probably our best chance on figuring out why it works like this today.

aljungberg:
It's been 5 years so I might be a little foggy on this. But this change, as it stands, does in fact cause a 1 event lag, just like the comment says, doesn't it? 

Starting from a point where the `_DOMScrollingElement` is invisible, we make it visible. Note that no amount of scrolling could have happened to it yet because it wasn't visible before. When we check the deltas on `_DOMScrollingElement.scrollLeft` and `_DOMScrollingElement.scrollTop` they are 0. As a result we send no `CPEvent` (unless the `deltaX` and `deltaY` fallback kicks in).

Next if a second scrolling event is received, `_DOMScrollingElement` has been scrolled. It has been scrolled by the *previous* event. But we are now, indeed, 1 event behind. We are showing the results of the previous scrolling event. When we return to the browser it will scroll some more.

In addition to being 1 event behind, we'll also drop the final event entirely.

I'm guessing the real issue is related to [chromium 570845](https://bugs.chromium.org/p/chromium/issues/detail?id=570845). Maybe we need to use `requestAnimationFrame` instead to check on the movements of the scroller. This could actually be a performance benefit too since we could limit scroll updates to 60 per section even if we get a flood of them during touch scrolling.

daboe01:
@aljungberg thank you for hinting towards  https://bugs.chromium.org/p/chromium/issues/detail?id=570845
I am pretty sure this is the problem, albeit i started seeing this issue starting from chrome 52.
51 was definitely ok on my maschines.

daboe01:
i agree with @aljungberg that we should fix this with requestAnimationFrame.
I will try this and eventually open a new PR as time permits.

daboe01:
one final note: given the abundance of events during scrolling, i feel that lagging one behind could be safely neglected.

aljungberg:
> one final note: given the abundance of events during scrolling, i feel that lagging one behind could be safely neglected.

Yes in my cases I imagine that's true, although I can also imagine situations where a quick scroll only gives us a single event and missing it would be bad. I guess the only way to know for sure is to run some experiments on modern browsers on different platforms, but that might be more effort than simply switching to requestAnimationFrame (which could have other benefits too).

PR #2470, opened 2016-08-05
Fixed: obsolete file in root directory

daboe01:
previously, the root directory contained a so called rakefile.
this does not make sense anymore these days and is removed with this PR


Dogild:
Thanks

PR #2475, opened 2016-08-12
Fixed: missing requirements for developers...

daboe01:
... in README.markdown

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -22,6 +22,8 @@ System Requirements
 To run Cappuccino applications, all you need is a web browser that understands
 JavaScript.
 
+To develop Cappuccino applications, all you need is a simple text editor. However, Cappuccino's integrated Xcode support leverages the eases of Cocoa development for web development. 
+
 To build Cappuccino itself, please read below. More information is available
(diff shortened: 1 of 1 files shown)


daboe01:
-#new
+documentation

mrcarlberg:
We should add information that we need ```gcc``` or ```clang``` to compile Objective-J code from the command line. It is used to preprocess all macros like ```#include```, ```#define``` and ```#if``` etc...

daboe01:
@aparajita will this hold for cappuccino 2 (beeing able to code with only a simple editor?)

mrcarlberg:
We want everything to be javascript so the need for ```gcc``` or ```clang``` will disappear in the future.

We don't know exactly when or how at the moment but this is an important issue.

daboe01:
from a marketing perspective i would like to be minimalistic here.
shouldn't we leave it as in https://github.com/cappuccino/cappuccino/pull/2475/commits/a4d374ecb6aad9ad5a904529090613b14095628d?

mrcarlberg:
Ok, I do agree with you. Please revert!

mrcarlberg:
Thanks!

PR #2477, opened 2016-08-16
fixed: scrolling events are missed on slower maschines in chrome 52

### LINKED ISSUE #2468: Scrolling by momentum is broken in Chrome 52 on slower maschines

daboe01:
Previously, the scrolling events were generated in a setTimeout with a delay of 0ms.
These were not reliably executed in recent chrome versions which caused freezes during scrolling with momentum on older systems such as a MBP late 2011.
This PR replaces the setTimeout with requestAnimationFrame and uncomments the limitDateForMode:CPDefaultRunLoopMode to enforce the visual updates.
Tested on FF, Safari and of course chrome.
Fixes #2468

### DIFF
--- a/AppKit/Platform/DOM/CPPlatformWindow+DOM.j
+++ b/AppKit/Platform/DOM/CPPlatformWindow+DOM.j
@@ -952,8 +952,8 @@ _CPPlatformWindowWillCloseNotification = @"_CPPlatformWindowWillCloseNotificatio
                                   timestamp:timestamp windowNumber:windowNumber context:nil eventNumber:-1 clickCount:1 pressure:0];
     event._DOMEvent = aDOMEvent;
 
-    // We lag 1 event behind without this timeout.
-    setTimeout(function()
+    // We lag 1 event behind without this approach
+    window.requestAnimationFrame(function()
     {
         if (aDOMEvent.deltaMode !== undefined && aDOMEvent.deltaMode !== 0)
         {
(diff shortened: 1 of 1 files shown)


daboe01:
this PR uses the requestAnimationFrame that @aljungberg suggested in #2469

daboe01:
-#new
+regression
+AppKit
+#needs-review

daboe01:
+Chrome

mrcarlberg:
I have tested this on major browsers on both Mac and PC and it sure fixes the problems with scrolling on Chrome version > 51.

mrcarlberg:
Thanks

aljungberg:
`requestAnimationFrame` is not supported in IE 8 or 9. We might want to fall back on regular `setTimeout` in that case. I don't know how many users care about those browsers anymore.

daboe01:
interestingly, the chrome team seems to have addressed this issue upstream. i do not see it anymore in chrome version 52.0.2743.116 (64-bit).

daboe01:
unfortunately, we are still not done with this.
i see poor scrolling experience on chrome again, this time with retina display. It looks like that chrome again randomly skips our function when it takes longer than one display cycle.
@aljungberg  @mrcarlberg shouldn't we we accept the 'lag 1 event behind' and just execute synchronously? in my testing on a mac book pro touchpad, this lag is completely unnoticable.

PR #2478, opened 2016-08-19
Fix mobile text entry

aljungberg:
Fixes iOS virtual keyboard "done" button not working and tapping other text fields while editing (often) not working.

(Also some unrelated typo fixes. Sorry.)


peterbaral:
This problem reappears with iOS/iPadOS 14 and 15 and Safari 14 and 15.

PR #2479, opened 2016-09-06
CPTabView: load tab view items asynchronously when possible

cacaodev:
= when the item view is managed by a view controller as a result of +tabViewItemWithViewController:

Manual test CPTabViewNib have been modified to show a tab view item managed by a view controller.
The appropriate place to do some setup on the view is now -viewDidAppear: because we cannot use any more the tabView:didSelectTabViewItem: delegate method which is synchronous.


AppKit/CPTabView.j
```diff
@@ -335,13 +335,36 @@ var CPTabViewDidSelectTabViewItemSelector           = 1 << 1,
...
+    {
+        [controller loadViewWithCompletionHandler:function(view, error)
+        {
+            if (view !== nil && error == nil)
+            {
+                [aTabViewItem setView:view];
+
+                if ([self selectedTabViewItem] == aTabViewItem)
+                    [self _displayItemView:view];
+            }
```
Dogild: Should we put a warning when we get an error here ?

cacaodev: Done

Dogild:
Looks good to me! You can merge it ;)

PR #2485, opened 2016-09-26
nib2cib: discard warnings in the nib file.

cacaodev:
ibtool optionally generates warnings in the nib file that are wrapped into a connection. Now we print the warning when nib2cibing but ignore this entry when converting to cib.

### DIFF
--- a/Tools/nib2cib/NSIBObjectData.j
+++ b/Tools/nib2cib/NSIBObjectData.j
@@ -50,7 +50,12 @@
         _classesKeys = [aCoder decodeObjectForKey:@"NSClassesKeys"];
         _classesValues = [aCoder decodeObjectForKey:@"NSClassesValues"];
 
-        _connections = [aCoder decodeObjectForKey:@"NSConnections"];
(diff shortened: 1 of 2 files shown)


mrcarlberg:
@cacaodev I want to test this. How can I create a connection that will generate a warning?

cacaodev:
You can't use nib2cib because CPStackView is not supported yet.
You can create a xib file in Xcode 8 (tested with the 10.11 sdk). Add a NSStackView to a window. Then compile the xib using` ibtool --compile path.plist `and the `--output-format xml1` option to have something readable. The connection will appear at the end of the plist and show the warning. Maybe it's an Apple bug to have the warning here. The only workaround I found is this one.

mrcarlberg:
Looks good!

mrcarlberg:
Merged! Thanks!

cacaodev:
@mrcarlberg I reverted this because of a nib2cib error. I will submit a new and complete one when the fix becomes really useful.

PR #2486, opened 2016-09-27
Fixed:  textsizing was wrong in safari

daboe01:
due to the color panel being created during init of the font panel


daboe01:
sorry, this fix does not work yet. i will keep you updated here.

daboe01:
now it works for me. as it turns out, you have to always set the font before sizing in safari. setting the is, however, a very expensive operation in chrome. therefore i introduced some browser sniffing to do this only in safari.

AppKit/CPStringDrawing.j
```diff
@@ -74,6 +75,12 @@ CPStringSizeCachingEnabled = YES;
         CPStringSizeMeasuringContext.font = [aFont cssString];
         var teststring = "0123456879abcdefghiklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ,.-()";
         CPStringSizeIsCanvasSizingInvalid = ABS([CPPlatformString sizeOfString:teststring withFont:aFont forWidth:0].width - CPStringSizeMeasuringContext.measureText(teststring).width) > 2;
+
+        CPStringSizeNeedsAlwaysSetFont = NO;
+        var ua = navigator.userAgent.toLowerCase();
+        
+        if (ua.indexOf('safari') !== -1 && ua.indexOf('chrome') === -1)
+            CPStringSizeNeedsAlwaysSetFont = YES; // Safari
```
mrcarlberg: We want you to check things like this in ```CPCompatibility.j```. Add a feature (please document it well) and use the provided engine checks.

Ask if the feature is present from this method.

daboe01:
the problem with my case is that the engines need to be mutually exclusive. We already have a webkit engine. Do you think it makes sense to split this into webkit-safari and webkit-chrome?

mrcarlberg:
Yes, as Chrome has left the Webkit and created there own engine called Blink. The Cappuccino project has not yet recognised this. You should add a new engine call Blink and create a new feature that checks against the new engine.

daboe01:
+#needs-review

AppKit/CPCompatibility.j
```diff
@@ -80,6 +81,8 @@ CPFileAPIFeature                        = 31;
 
 CPAltEnterTextAreaFeature               = 32;
 
+//  Safari calculates incorrect text size unless you set the canvas font even if it is already set
+CPTextSizingAlwaysNeedsSetFontFeature  = 33;
```
cacaodev: If it's a bug, it should go in PLATFORM_BUGS.

AppKit/CPCompatibility.j
```diff
@@ -179,7 +182,10 @@ else if (USER_AGENT.indexOf("AppleWebKit/") != -1)
         PLATFORM_BUGS |= CPJavaScriptPasteRequiresEditableTarget;
         // https://bugs.webkit.org/show_bug.cgi?id=39689
         PLATFORM_BUGS |= CPJavaScriptPasteCantRefocus;
+        PLATFORM_FEATURES[CPTextSizingAlwaysNeedsSetFontFeature] = YES;
     }
+	else
+        PLATFORM_ENGINE = CPBlinkBrowserEngine;
```
cacaodev: Some browser version check is needed here. The named Blink engine is not present in all chrome versions.

AppKit/CPCompatibility.j
```diff
@@ -30,7 +30,8 @@ CPGeckoBrowserEngine                    = 1;
 CPInternetExplorerBrowserEngine         = 2;
 CPKHTMLBrowserEngine                    = 3;
 CPOperaBrowserEngine                    = 4;
-CPWebKitBrowserEngine                   = 5;
+CPWebKitBrowserEngine                   = 5;  // Safari
+CPBlinkBrowserEngine                    = 6;  // Chrome
```
cacaodev: We need to be careful because this means that the engine checks against  CPWebKitBrowserEngine will now fail with recent versions of Chrome. I don't think this is what we want.

AppKit/CPCompatibility.j
```diff
@@ -179,7 +182,10 @@ else if (USER_AGENT.indexOf("AppleWebKit/") != -1)
         PLATFORM_BUGS |= CPJavaScriptPasteRequiresEditableTarget;
         // https://bugs.webkit.org/show_bug.cgi?id=39689
         PLATFORM_BUGS |= CPJavaScriptPasteCantRefocus;
+        PLATFORM_FEATURES[CPTextSizingAlwaysNeedsSetFontFeature] = YES;
```
cacaodev: If there is a way to detect this directly, it's always preferable to a simple engine/browser detection.

AppKit/CPCompatibility.j
cacaodev: I've seen that and it's a good thing because internally we are not supposed to use engine detection. On the other hand we still have this CPBrowserIsEngine() public method and we don't want apps that use it to fail when they check the WebKit engine in Chrome.
The question behind that is : does Blink support all the features the Webkit engine supported at the time Blink was introduced ?. If yes (i suspect the answer is yes), then the current chrome  supports both engines. In this case, we should allow that by turning these constants into masks.

AppKit/CPCompatibility.j
cacaodev: You just need to check the version. The info i got from the internets is that Blink started at Chrome 28.

daboe01: @cacaodev i made Blink detection functional: http://stackoverflow.com/questions/20655470/how-to-detect-blink-in-chrome

AppKit/CPCompatibility.j
daboe01: i agree 100%. i converted the engine variable into a mask

cacaodev:
@daboe01 Great, thanks. Can you add a comment with a reference to this bug (WebKit bug tracker or else) if you have one. If not, a test would be helpful in order to check when it is eventually fixed in Webkit.

daboe01:
@cacaodev luckly, the workaround seems to be very cheap performance wise.

daboe01:
issue: https://bugs.webkit.org/show_bug.cgi?id=150224

mrcarlberg:
I have tested this and it looks good.

mrcarlberg:
@daboe01 Do you have anything more you want to do on this PR before I merge it?

mrcarlberg:
@cacaodev Are all issues you had on this PR resolved now?

daboe01:
I am finished. 

Von meinem iPhone gesendet

AppKit/CPStringDrawing.j
```diff
@@ -59,22 +57,12 @@ CPStringSizeCachingEnabled = YES;
     return [self sizeWithFont:aFont inWidth:NULL];
 }
 
-- (void) _initializeStringSizing
++ (void) initialize
```
cacaodev: -  +initialize should not  be in a category but in the main class implementation (for clarity).
-  You don't need to call super because +initialize is already sent to the superclass.
-  If you don't need +initialize to be sent to super, you need to block it in the subclass, otherwise it will be sent to both the subclass and  the superclass with the subclass impl (this one). So wrap this into a check  `if ([self class] == [CPString class) { }`

cacaodev: With these changes, it's ready to commit for me.

daboe01: hi @cacaodev i am not sure how i can move this from the CPStringDrawing category the main class implementation (CPString.j). How could i access file-private globals such as _CPStringSizeIsCanvasSizingInvalid_ from within CPString.j?

PR #2491, opened 2016-10-05
Fixed: CPView setFrameSize was calling _updateTrackingAreas too often

didierkorthoudt:


didierkorthoudt:
All unit tests on CPView still work.
Manuel test (TrackingArea) still work.

mrcarlberg:
I have tested this and it is much faster 😃 

But, after a while playing around in a table view. Changing window size and column width with the mouse I get a crash in ```[CPWindow _flushTrackingEventQueueForMouseAt:]```.

As you can see in the attached screen dump the trackingArea for the first ```_queuedTrackingEvents``` is ```undefined```.

I get this problem every time after a few changes on some columns width. I don't have this problem when running on an older version of Cappuccino.

skarmavbild 2016-10-06 kl 13 02 47

didierkorthoudt:
@mrcarlberg Thank you for testing ! And I'm happy that you found this speedier...

:-)

Well, the _flushTrackingEventQueue bug is strange as I didn't touch the CPWindow class.

I'll check this ASAP. This must be a side effect. Thank you also for reporting this.

didierkorthoudt:
Going to PR #2492

PR #2498, opened 2016-10-26
Fixed:  font can be nil sometimes

daboe01:
previously, _font was nil sometimes in cib-generated textfields. This caused crashes as reported here: https://groups.google.com/forum/?fromgroups#!topic/objectivej/-FeO-hNbN3w

This PR ensures that the font property is never nil.


Dogild:
Thanks!

PR #2500, opened 2016-10-29
Fixed: CPViewAnimator was not giving correct frame information

didierkorthoudt:
First work

### DIFF
--- a/AppKit/CoreAnimation/CPAnimationContext.j
+++ b/AppKit/CoreAnimation/CPAnimationContext.j
@@ -153,7 +153,7 @@ CPLog.debug(_cmd + "context stack =" + _CPAnimationContextStack);
     duration = [animation duration] || [self duration];
 
     var needsFrameTimer = (aKeyPath === @"frame" || aKeyPath === @"frameSize" || aKeyPath === @"frameOrigin") &&
-                          ([anObject hasCustomLayoutSubviews] || [anObject hasCustomDrawRect]) &&
+                          [[anObject animator] wantPeriodicFrameUpdates] &&
                           (objectId = [anObject UID]);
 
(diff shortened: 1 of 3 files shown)


didierkorthoudt:
Now, correct frame is computed during the view animation.
@cacaodev : as you will see, I've tried to make this working also for frameOrigin but without success. Any idea ? Also, a final setFrame (or setFrameOrigin, or setFrameSize according to animation going on) should be fired without the _setInhibitDOMUpdate but I didn't find where until now.

AppKit/CoreAnimation/CPAnimationContext.j
```diff
@@ -677,31 +677,44 @@ FrameUpdater.prototype.addTarget = function(target, keyPath, duration)
...
                 return ROUND(parseFloat(style.getPropertyValue(prop)));
-          };
+          },
 
-    var updateFrame = function(timestamp)
+        initialFrame  = [aView frame],
+        initialX      = initialFrame.origin.x,
+        initialY      = initialFrame.origin.y,
+        initialWidth  = initialFrame.size.width,
+        initialHeight = initialFrame.size.height,
```
Dogild: Just keep the initialFrame. Make a copy of it, so it will not be updated with the animation

cacaodev:
Bravo for getting the frame values from the matrix, the toughest part. 
I will have a look to the frameOrigin: not getting called later. About the final step: for all animations, there is an internal completion handler (CompletionHandlerAgent) that syncs css values and cappuccino ivars by calling the cappuccino method related to the keyPath. If I remember well (not tested) it was disabled for animations getting frame updates because redundant.

didierkorthoudt:
Replaced by #2502

PR #2502, opened 2016-10-31
Fixed: CPViewAnimator was not giving correct frame information

didierkorthoudt:
Replaces #2500.

With this PR, correct frame/frameSize/frameOrigin are available when animating a view.

If your animated view doesn’t implement a custom layoutSubviews or drawRect, simply set the new flag : ```[[aView animator] setWantPeriodicFrameUpdates:YES]``` (this is automatically done for you if you have at least one of the 2 custom methods implemented).

This will call setFrame / setFrameOrigin / setFrameSize accordingly to the property you're animating.


AppKit/CPView.j
```diff
@@ -252,6 +252,8 @@ var CPViewHighDPIDrawingEnabled = YES;
 
     id                  _animator;
     CPDictionary        _animationsDictionary;
+    BOOL                _inhibitDOMUpdates      @accessors(setter=_setInhibitDOMUpdates);
+    BOOL                _forceDOMUpdates        @accessors(setter=_setForceDOMUpdates);
```
cacaodev: This flag forces everything, not just DOM updates. Should be renamed to reflect that _forceUpdates or just _force for example.

AppKit/CoreAnimation/CPAnimationContext.j
```diff
@@ -162,14 +162,19 @@ CPLog.debug(_cmd + "context stack =" + _CPAnimationContextStack);
     var completionFunction = function()
     {
         if (needsFrameTimer)
+        {
             [self stopFrameUpdaterWithIdentifier:objectId];
+            [anObject _setForceDOMUpdates:YES];
```
cacaodev: This should be set and unset in each of the view animator proxy methods using the convenience method:
```[self _setTargetValue:aTargetValue withKeyPath:aKeyPath fallback:handler completion:handler];```
Be aware that the completion handler needs this flag to be on, but not the fallback handler (the handler being executed when animations are unavailable).

File: AppKit/CoreAnimation/CPAnimationContext.j [hunk 3]
```diff
@@ -263,10 +268,8 @@ CPLog.debug(_cmd + "context stack =" + _CPAnimationContextStack);
 - (void)getAnimations:(CPArray)cssAnimations getTimers:(CPArray)timers forView:(CPView)aTargetView usingAction:(Object)anAction rootView:(CPView)rootView cssAnimate:(BOOL)needsCSSAnimation
 {
     var keyPath = anAction.keypath,
-        isFrameKeyPath = (keyPath == @"frame" || keyPath == @"frameSize"),
-        customLayout = [aTargetView hasCustomLayoutSubviews],
-        customDrawing = [aTargetView hasCustomDrawRect],
-        needsFrameTimer = isFrameKeyPath && (customLayout || customDrawing);
+        isFrameKeyPath  = (keyPath === @"frame" || keyPath === @"frameSize" || keyPath === @"frameOrigin"),
+        needsFrameTimer = isFrameKeyPath && [[aTargetView animator] wantPeriodicFrameUpdates];
```
cacaodev: Can you rename needsFrameTimer to needsPeriodicFrameUpdates because it's no longer a timer. Thanks.

AppKit/CoreAnimation/CPAnimationContext.j
```diff
@@ -674,31 +677,38 @@ FrameUpdater.prototype.addTarget = function(target, keyPath, duration)
...
-    var updateFrame = function(timestamp)
+        getCSSPropertyValue = function(prop) {
+            return ROUND(parseFloat(style.getPropertyValue(prop)));
+        },
+        initialFrame = CGRectMakeCopy([aView frame]),
+        updateFrame  = function(timestamp)
     {
-        var width  = getCSSPropertyValue("width"),
-            height = getCSSPropertyValue("height");
+        var matrix = style["transform"].split('(')[1].split(')')[0].split(','),
```
cacaodev: Use the CPBrowserCSSProperty function to get the transform property.

AppKit/CoreAnimation/CPViewAnimator.j
```diff
@@ -1,9 +1,20 @@
 
 @import "_CPObjectAnimator.j"
 @import "CPView.j"
+@import "CPCompatibility.j"
 
 @implementation CPViewAnimator : _CPObjectAnimator
 {
+    BOOL    wantPeriodicFrameUpdates    @accessors;
```
cacaodev: ivar must be underscored.

AppKit/CoreAnimation/CPViewAnimator.j
```diff
@@ -76,34 +87,42 @@
 
 @end
 
-var transformOrigin = function(start, current)
+var CSSStringFromCGAffineTransform = function(anAffineTransform)
 {
-    return "translate(" + (current.x - start.x) + "px," + (current.y - start.y) + "px)";
+    // Firefox : add px to the translate values.
+    if (CPBrowserIsEngine(CPGeckoBrowserEngine))
+        return "matrix(" + anAffineTransform.a + ", " + anAffineTransform.b + ", " + anAffineTransform.c + ", " + anAffineTransform.d + ", " + anAffineTransform.tx + "px, " + anAffineTransform.ty + "px)";
```
cacaodev: Same here, use the CPBrowserCSSProperty please.

AppKit/CoreAnimation/CPViewAnimator.j
```diff
@@ -76,34 +87,42 @@
...
+    else
+        return "matrix(" + anAffineTransform.a + ", " + anAffineTransform.b + ", " + anAffineTransform.c + ", " + anAffineTransform.d + ", " + anAffineTransform.tx + ", " + anAffineTransform.ty + ")";
```
cacaodev: You can just make this more compact by appending "px"  | "".

AppKit/CoreAnimation/CPAnimationContext.j
didierkorthoudt: Ignored as discussed.

AppKit/CoreAnimation/CPAnimationContext.j
didierkorthoudt: Done.

AppKit/CPView.j
didierkorthoudt: Done.

AppKit/CoreAnimation/CPViewAnimator.j
didierkorthoudt: Done.

AppKit/CoreAnimation/CPAnimationContext.j
didierkorthoudt: Sorry but I don't understand where else you want me to force updates. This should only be set when the animation used "live" setFrameXXX with _inhibitDOMUpdates set as frame rect is, at the end of the animation, set to the final value but without having impacted the DOM (so, simply calling setFrameXXX has no effect as each method begins with ```if (CGRectEqualToRect(_frame, aFrame))``` (or similar for other methods). Maybe I'm missing something somewhere...  ;-)  Do you have any clue for me ?

didierkorthoudt:
CPAnimatablePropertyContainerTest does not behave correctly with this... Custom drawing and subviews animations are not correct. Digging into this...

cacaodev:
The problem is that the transform property is inherited by subviews and by the canvas. Unless all our controls support the transform property (not tested but improbable), we need a solution for that. Either a css way to block inheritance, or invert the transform matrix. Otherwise we will have to go back to the previous code.

didierkorthoudt:
I've just tested without subviews own animations and this is must better

didierkorthoudt:
Do you have an idea of such a control that wouldn't support this so I can test ?

AppKit/CoreAnimation/CPAnimationContext.j
```diff
@@ -152,24 +152,29 @@ CPLog.debug(_cmd + "context stack =" + _CPAnimationContextStack);
 
     duration = [animation duration] || [self duration];
 
-    var needsFrameTimer = (aKeyPath == @"frame" || aKeyPath == @"frameSize") &&
-                          ([anObject hasCustomLayoutSubviews] || [anObject hasCustomDrawRect]) &&
-                          (objectId = [anObject UID]);
+    var needsPeriodicFrameUpdates = (aKeyPath === @"frame" || aKeyPath === @"frameSize" || aKeyPath === @"frameOrigin") &&
+                                    [[anObject animator] wantPeriodicFrameUpdates] &&
```
cacaodev: The result here seems wrong if we have a custom layout and a frameOrigin keyPath. I'm afraid you will have to compute needsPeriodicFrameUpdates with the custom layout|drawing info here.
It's (isFrameOrFrameSizeKeyPath && isCustomLayoutOrDrawing) || wantPeriodicFrameUpdates.

cacaodev:
Try with a button, I'm pretty sure it will be ugly. But even if it worked, how could we handle custom layout ? We wouldn't know how to apply the transform because custom layout is a black box for us.
We could infer an animation if we had the end values (that's what I did with autosize) but unfortunately we don't have that info.

didierkorthoudt:
I'll try with a button.

Maybe it's a little bit too simple but, for what I understand, with the new organisation (using _inhibitDOMUpdates and so on), the animation acts on the view itself which then calls it's setFrameXXX and/or layoutSubviews / drawRect, using all "natural" Cappuccino system (the only thing we don't do is letting cappuccino act on the animated view DOM to avoid interfering with the CSS animation). So, naively, everything should work as usual...

No ?

didierkorthoudt:
OK, forget about my (too) naive approach... I've tested with a button and that's horrible !..  oO

I'll adapt code from your comments and search for a solution...

AppKit/CoreAnimation/CPAnimationContext.j
didierkorthoudt: Done.

AppKit/CoreAnimation/CPAnimationContext.j
```diff
@@ -152,24 +152,29 @@ CPLog.debug(_cmd + "context stack =" + _CPAnimationContextStack);
...
+        }
 
         if (animationCompletion)
             animationCompletion();
 
-        if (needsFrameTimer || animationCompletion)
+        if (needsPeriodicFrameUpdates || animationCompletion)
             [[CPRunLoop currentRunLoop] performSelectors];
 
+        [anObject _setForceUpdates:NO];
```
didierkorthoudt: Done (added a -respondsToSelector test).

AppKit/CoreAnimation/CPViewAnimator.j
```diff
@@ -1,9 +1,20 @@
 
 @import "_CPObjectAnimator.j"
 @import "CPView.j"
+@import "CPCompatibility.j"
 
 @implementation CPViewAnimator : _CPObjectAnimator
 {
+    BOOL    _wantPeriodicFrameUpdates   @accessors(property=wantPeriodicFrameUpdates);
```
didierkorthoudt: Done (I though this was some king of habit in Cappuccino...)

AppKit/CoreAnimation/CPAnimationContext.j
cacaodev: Because -layoutSubviews needs the receiver frameSize but not the frameOrigin. If you show me an example where the subviews are displayed based on the superview frameOrigin I will admit I'm wrong. As I said, frameOrigin updates should be opt-in. They are needed in your case because you need to draw from the superview but it's not the common case.

AppKit/CoreAnimation/CPAnimationContext.j
cacaodev: It's really more consistent to have this in the keyPath related completion handler that is called in any case (and there is a reason for that). If there are no frame updates i agree the flag is useless (just opening an open door as we say in my country) but it's completely harmless.

didierkorthoudt: Well, in fact there are plenty of places where it's implicit that we're dealing with views...

AppKit/CoreAnimation/CPAnimationContext.j
didierkorthoudt: Done.

cacaodev:
There are now too different bug fixes/features in this PR. I have to ask you to keep in this one the original bug fix described in the title. This way I will be able to test it and push it quickly.

You can open a new PR for the others items : 
- inhibitDomUpdates
- request frames updates
-  sub animations removal

Thanks.

didierkorthoudt:
Hi @cacaodev Well, I've checked how I could split the two but the fact of fixing the live information in FrameUpdater (top and left values, corrected by using the matrix for the translation) makes the setFrameXXX methods actually work, that is they modify the position of the view, thus interfering with the DOM CSS animation. This is the origin of the inhibitDOMUpdates flag.
Indeed the sub animations removal could be considered as a separate update but is a direct consequence of the fix as, everything set, the usual layout system can work as expected and is thus redondant with sub animations.
OK, I could remove the wantsPeriodicFrameUpdates flag and redo it in a new PR but is it worth the work ?
Just let me know and I'll adapt.

didierkorthoudt:
A more simple solution could be to rename the PR...  ;-)

cacaodev:
It seems I won't convince you of the benefits of incremental development although they are real (faster to commit, safer code, easier to test) 
... anyway in the demo, the frame keyPath is not animated correctly. I'm seeing the animation going too far, then jumping back.

Also, the debate css animations vs. requestAnimationFrame is interesting but it definitly needs to be opened and discussed in another PR.

didierkorthoudt:
Well, I'm already convince but we should have split code modifications sooner... For the beauty of the art, I'll try to split as you mentioned.

The behavior you observe is strange as it works here. I'll make more tests to be sure. If it jumps back, that means the CSS animation is going too far and then that translation parameters are not correct.

didierkorthoudt:
Also, I've got performances concerns when using this in my app. The FrameUpdater is called only twice during the animation, whereas, in tests, it's called many-many times...

didierkorthoudt:
Re-tested and no problem with the frame animation. It stops just touching the bottom of the white area. I've checked subviews + custom layout + custom drawing. FYI, layoutSubviews is called more than 70 times during the animation.

didierkorthoudt:
Well, found something else. Strange. It worked but no more. Stay tuned.

cacaodev:
I'm not seeing anymore the jumping back problem, sorry for this false alert, maybe a cache issue. But I see frames not updated in the following configurations:
- frameSize + customDrawing
- frameSize + has Subviews + custom layout
- frameSize + has Subviews + auto layout (means autosize)
- frame + has subviews + autosize

On Safari 10.0.1

didierkorthoudt:
Yes, that's what I see now also (same browser). Something must have interfere... I dig.

didierkorthoudt:
OK, done some cleaning in my env.

So : for frame+has subviews+auto layout : we must set wantsPeridicFrameUpdates to YES (that's fine according to our new policy).

For all frameSize, I forgot to impact a fix and an error message should be in the console arguing that undefined blah blah blah (that is because when only animating frame size, there's no matrix, so we can't split nothing...). Fixed. I push a commit for this.

I continue to test (also for my performances problem).

didierkorthoudt:
One problem remains : frameSize + custom drawing. Strange.

didierkorthoudt:
+ frame & custom drawing. Size changes are no more done.

didierkorthoudt:
OK. Found. As when animating frame size (width & height) I reverted to your property animation (because of the side effects of using scaling within the matrix), I can't anymore inhibit DOM updates on setFrameSize (but still on setFrameOrigin). I go this way, I push you a commit (so you can test that everything is working) and try to split all those adaptations into 2 separate PR's.

cacaodev:
With your commit the first 2 items i mentioned are ok.
Autosize (checkbox "has Subviews" + "Auto Layout") is still not working for frame &frameSize keyPaths.
There is also an issue if you select backgroundColor + frame keypaths + custom drawing.

AppKit/CPView.j
```diff
@@ -252,6 +252,8 @@ var CPViewHighDPIDrawingEnabled = YES;
 
     id                  _animator;
     CPDictionary        _animationsDictionary;
+    BOOL                _inhibitDOMUpdates      @accessors(setter=_setInhibitDOMUpdates);
+    BOOL                _forceUpdates           @accessors(setter=_setForceUpdates);
```
cacaodev: These ivars need to be intialized with default values.

AppKit/CPView.j
```diff
@@ -1084,7 +1086,7 @@ var CPViewHighDPIDrawingEnabled = YES;
 {
     var origin = _frame.origin;
 
-    if (!aPoint || CGPointEqualToPoint(origin, aPoint))
+    if ((!aPoint || CGPointEqualToPoint(origin, aPoint)) && !_forceUpdates)
```
cacaodev: If aPoint is nil, it should return.

AppKit/CoreAnimation/CPAnimationContext.j
```diff
@@ -152,24 +152,32 @@ CPLog.debug(_cmd + "context stack =" + _CPAnimationContextStack);
...
 
         if (animationCompletion)
             animationCompletion();
 
-        if (needsFrameTimer || animationCompletion)
+        if (needsPeriodicFrameUpdates || animationCompletion)
             [[CPRunLoop currentRunLoop] performSelectors];
 
+        if ([anObject respondsToSelector:@selector(_setForceUpdates:)])
+            [anObject _setForceUpdates:NO];
```
cacaodev: I would prefer _setForceUpdate to be inside animationCompletion. If not it should be balanced with the same condition as the previous call.

didierkorthoudt:
For frame+has subviews+auto layout : we must set wantsPeridicFrameUpdates to YES (that's fine according to our new policy). I've tested this and it works.

```
    if (leftView)
    {
        [leftView setFrame:CGRectMake(0, 0, 200, 200)];
        [animationSandbox addSubview:leftView];
        [[leftView animator] setWantsPeriodicFrameUpdates:YES];
    }
```

didierkorthoudt:
Regarding the backgroundColor + frame keypaths + custom drawing problem (no background color animation), when I test on your version, it doesn't animate background color either... (no need to check frame keypaths, this is also wrong with only background color animation + custom drawing). I wonder if this is not "normal" (or logical) as custom drawing should be responsible to set the background, no ?

AppKit/CPView.j
didierkorthoudt: Oops ! Indeed ! I should have seen this myself...

AppKit/CPView.j
```diff
@@ -1116,7 +1121,7 @@ var CPViewHighDPIDrawingEnabled = YES;
 {
     var size = _frame.size;
 
-    if (!aSize || CGSizeEqualToSize(size, aSize))
+    if ((!aSize || CGSizeEqualToSize(size, aSize)) && !_forceUpdates)
```
didierkorthoudt: Oops ! Indeed ! I should have seen this myself...

AppKit/CPView.j
didierkorthoudt: OK, no problem. Done.

cacaodev:
I'm not seeing a lack of backgroundColor animation; it seems the two animations have different frame sizes. If you say there is no regression, we'll see that later.

cacaodev:
> For frame+has subviews+auto layout : we must set wantsPeridicFrameUpdates to YES (that's fine according to our new policy). I've tested this and it works.

It may work like this but it's a deviation from cocoa where you can animate a control with one line
``` [[control animator] setFrame:aFrame]```
This won't work in cappuccino if the control contains autosized view (most of them do).

It also break current code and creates an inconsistency with views governed by custom layout where this call is not needed.

**Comment by didierkorthoudt** [comment]
I was also wondering if  we shouldn't fire FrameUpdate in any case... This will be consistent accross various view types... What do you think ? In fact this resumes in removal of wantsPeriodicFrameUpdates and always assume yes...

**Comment by didierkorthoudt** [comment]
> I'm not seeing a lack of backgroundColor animation; it seems the two animations have different frame sizes. If you say there is no regression, we'll see that later.

I've redone the test and selecting background color animation + custom drawing doesn't animate the color.

I'm pretty sure (but won't bet my head on this) I'm on the latest build. Could you please check this ?

If it's working for you, I'll create a fresh install of Cappuccino.

**Comment by cacaodev** [comment]
I could not reproduce it, maybe I was on the wrong build.

**Comment by didierkorthoudt** [comment]
OK, I'll split this PR. I'll adapt this one and reinsert the code for subviews.

One question remains : can I remove the wantsPeriodicFrameUpdates and assume it's always the case if this is a frameXXX key path ?

**Comment by cacaodev** [comment]
We discussed that before. IMO, the default frameOrigin animation should not be potentially degraded  just to get an information it may not use.

**Comment by didierkorthoudt** [comment]
OK, that was hard but here we go : I've put back subviews animations (plus the fixes requested).

I think I've tested everything with the CPAnimatablePropertyContainerTest (adding ```[[leftView animator] setWantsPeriodicFrameUpdates:YES]``` only to test if frameOrigin can provide live info.

I now open a new PR.

**Comment by cacaodev** [comment]
Great, one last thing: can you try to re-enable _inhibitDOMUpdates just for the DOM element sizing but not for the canvas sizing. I think the lack of canvas sizing was causing the problem in the configuration frameSize + custom drawing.
It would mean adding the condition on CPView L1377.

**Comment by didierkorthoudt** [comment]
Sorry but I don't understand what you want me to do... (maybe because it's Sunday morning...  ;-)  )
Also, I don't see any problem with frameSize + custom drawing...

**Comment by cacaodev** [comment]
Just to change and test -_setDisplayServerSetStyleSize: with this implementation:
```objective-j
- (void)_setDisplayServerSetStyleSize:(CGSize)aSize
{
#if PLATFORM(DOM)
    var scale = [self scaleSize];
    
    if (!_inihibitDOMUpdates)
        CPDOMDisplayServerSetStyleSize(_DOMElement, aSize.width * 1 / scale.width, aSize.height * 1 / scale.height);

    if (_DOMContentsElement)
    {
        CPDOMDisplayServerSetSize(_DOMContentsElement, aSize.width * _highDPIRatio * 1 / scale.width, aSize.height * _highDPIRatio * 1 / scale.height);
        CPDOMDisplayServerSetStyleSize(_DOMContentsElement, aSize.width * 1 / scale.width, aSize.height * 1 / scale.height);

        _needToSetTransformMatrix = YES;
    }
#endif
}

```

cacaodev:
Very last thing: setWantsPeriodicFrameUpdates:NO does not make sense. The naming is not consistent with the current behavior and we don't want to. For frame&frameSize keypath, frame updates are needed in order to work normally.
If you agree, we an rename it to ```requestPeriodicFrameUpdates``` with no argument. I can do this after your commit if you want.

didierkorthoudt:
Regarding the setWantsPeriodicFrameUpdates:NO, I don't agree with you (for once...  ;-)  ). You may have a view that needs to get x/y info on setFrameOrigin animation in some circumstances and don't in other (for performance).
I have such a view in my own app : when the view is "linked" to its "parent" (with a drawn link), I need x/y coordinates while animating to update the link. But this view can be "detached" from its parent and then, I don't need to update the link while animating.
What do you think ?

cacaodev:
OK then. It would be great to update the demo to show this: a checkbox that enables/disables a line between the 2 views and a periodic frames request.

cacaodev:
Merged, thanks !

PR #2503, opened 2016-10-31
Fix CPBox drawing artifacts when used in CPSeparator mode

enquora:
Suppress title positioning when CPBox used as a line or separator - by definition the title is null.


Tools/nib2cib/NSBox.j
```diff
@@ -71,9 +71,8 @@
         }
 
         _title         = [[aCoder decodeObjectForKey:@"NSTitleCell"] objectValue] || @"";
-        _titlePosition = [aCoder decodeObjectForKey:@"NSTitlePosition"];
 
-        if (_titlePosition === undefined)
+        if (_titlePosition === undefined && _boxType !== CPBoxSeparator)
```
cacaodev: Hi enquora,
Is there a reason why titlePosition should not be decoded ? if you need to test for the presence of a key in the CPKeyedUnarchiver you can use ```-containsValueForKey``` and then set the value, including a default value.

enquora: I used mrcarlberg's suggestion from [this thread](https://groups.google.com/forum/#!topic/objectivej/_1FNQDcD7_4)
I haven't reviewed the full implementation of CPBox and was hesitant about using this code without a full understanding.

The problem with the current implementation is that when CPBox is used, it should have no title, so there should be no titlePosition. What would there be to decode?

I am still somewhat concerned there might be implications I haven't fully grokked.

mrcarlberg: @cacaodev If you look at the history of this line in the git repository there is some kind of fix for when the ```_titlePosition``` is not present. I suggested this fix as Xcode now has removed the title position from the .xib file for line separators, starting from I think Xcode 8. Maybe it is better to check if the attribute is missing with the ```containsValueForKey``` method but I think it is not in the scope of this pull request as @enquora says. This pull request is just to fix the new behaviour of Xcode.

cacaodev: Ok but that doesn't answer the question: why titlePosition decoding have been removed in this pull request ?

mrcarlberg: I don't think that line 74 should be removed. What we need is just the extra test ```_boxType !== CPBoxSeparator```. @enquora can you please put the line back again and I think we should be ok with this.

enquora: I somehow convinced myself that it would not work without removing line 74. This left me uneasy. And yet, a quick run just now with line 74 in seems to give the desired result. I need to have a broader look at the class and see what visual/manual tests we have in place. I will try to fit it in over the next day or so.

mrcarlberg:
Ok, I have looked at this again and I think we should attack this in a different manner. I think we should alter as little as possible in nib2cib (NSBox.j) and make the adjustment in CPBox.j when we initialize from the .cib file.

I have added two lines with a 'else' in the ```initWithCoder:``` method in CPBox.j. It solves the problem for me. Can you please check if this work for you?

The end of the method looks now like this:
``` objective-c
        }
        else
            _titlePosition = CPNoTitle;

        [self setAutoresizesSubviews:YES];
        [_contentView setAutoresizingMask:CPViewWidthSizable | CPViewHeightSizable];

        [self _manageTitlePositioning];
    }

    return self;
}
```

enquora:
I've had time today to look at this approach. It both works and is logically consistent with the design of the class. I'll prepare an updated PR.

mrcarlberg:
@enquora Any progress on this?

enquora:
Just pushed the fix to my own fork. Thought I had done this previously but was having trouble with the Github workflow. I presume the original PR will now be updated automatically.

mrcarlberg:
You should be able to remove the changes you have made in NSBox.j as the problem now is solved in CPBox.j ?

mrcarlberg:
Thanks! Merged!

Issue #2499, opened 2016-10-28
Linux: Unterminated regular expression


HTPC-Schrauber:
Hi,

I am trying to build cappuccion under Linux.

Ich checked out from github. The executed the bootstrap.sh.
So far all went fine. Then I run jake in cappuccino folder and get the following:

```
(in /home/maik/cappuccino)
(in /home/maik/cappuccino/Objective-J)
Copying... CommonJS/objj-executable
Copying... CommonJS/package.json
Copying... CommonJS/bin/cplutil
Copying... CommonJS/bin/objj
Copying... CommonJS/bin/objjc
Building... /home/maik/cappuccino/Build/Release/CommonJS/objective-j/lib/objective-j.js
Building... /home/maik/cappuccino/Build/Release/CommonJS/objective-j/lib/objective-j/cache-manifest.js
Building... /home/maik/cappuccino/Build/Release/CommonJS/objective-j/lib/objective-j/compiler.js
Building... /home/maik/cappuccino/Build/Release/CommonJS/objective-j/lib/objective-j/jake.js
Building... /home/maik/cappuccino/Build/Release/CommonJS/objective-j/lib/objective-j/loader.js
Building... /home/maik/cappuccino/Build/Release/CommonJS/objective-j/lib/objective-j/jake/applicationtask.js
Building... /home/maik/cappuccino/Build/Release/CommonJS/objective-j/lib/objective-j/jake/bundletask.js
Building... /home/maik/cappuccino/Build/Release/CommonJS/objective-j/lib/objective-j/jake/environment.js
Building... /home/maik/cappuccino/Build/Release/CommonJS/objective-j/lib/objective-j/jake/frameworktask.js
Copying... CommonJS/lib/objective-j/jake/LICENSES/LGPL-v2.1
Copying... CommonJS/lib/objective-j/jake/LICENSES/MIT
Copying... CommonJS/lib/objective-j/jake/RESOURCES/MHTMLTest.txt
Building... /home/maik/cappuccino/Build/Release/CommonJS/objective-j/lib/objective-j/rhino/regexp-rhino-patch.js
Building... /home/maik/cappuccino/Build/Release/Objective-J/Objective-J.js
    Version:   0.9.9
    Revision:  dda81b12da98955d55b8805b29ad3c695fc0cb91
    Timestamp: 1477660302130
Build took 1 second
(in /home/maik/cappuccino/CommonJS)
    Version:   0.9.9
[...]

PR #2504, opened 2016-11-01
fixed RegEx with newer gcc versions

HTPC-Schrauber:
According to [this](https://github.com/cappuccino/cappuccino/issues/2499) issue, cappuccino fails to build on linux with newer gcc versions (from 5.x ongoing).

This is because of the used RegEx in source. I changed all RegEx to the form of `new RegExp("xxxx")`.
Now tbuild works again. Tested on Fedora 24 with gcc 6.2.1

### DIFF
--- a/Tools/nib2cib/Converter.j
+++ b/Tools/nib2cib/Converter.j
@@ -165,7 +165,7 @@ ConverterConversionException = @"ConverterConversionException";
         if (system.engine === "rhino")
             plistContents = String(java.lang.String(plistContents).replaceAll("\\<key\\>\\s*CF\\$UID\\s*\\<\/key\\>", "<key>CP\\$UID</key>"));
         else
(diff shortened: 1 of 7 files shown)


Tools/nib2cib/Converter.j
```diff
@@ -165,7 +165,7 @@ ConverterConversionException = @"ConverterConversionException";
         if (system.engine === "rhino")
             plistContents = String(java.lang.String(plistContents).replaceAll("\\<key\\>\\s*CF\\$UID\\s*\\<\/key\\>", "<key>CP\\$UID</key>"));
         else
-            plistContents = plistContents.replace(/\<key\>\s*CF\$UID\s*\<\/key\>/g, "<key>CP$UID</key>");
+            plistContents = plistContents.replace("/\<key\>\s*CF\$UID\s*\<\/key\>/g", "<key>CP$UID</key>");
```
daboe01: is this really equivalent?
nothing mentioned on https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/replace

PR #2505, opened 2016-11-02
Fix RegEx for gcc > 4.9

HTPC-Schrauber:
On linux distributions with gcc > 4.9 cappuccino does not build because of the format of the RegEx.
I replaced all RegEx by a construct like new RegExp("xxx")
In the RegEx itself the \ had to be escaped for this to work.

This Pull Request refers to this [issue](https://github.com/cappuccino/cappuccino/issues/2499)

### DIFF
--- a/AppKit/CPTextView/CPTextView.j
+++ b/AppKit/CPTextView/CPTextView.j
@@ -1853,13 +1853,13 @@ var kDelegateRespondsTo_textShouldBeginEditing                                  
 
 + (JSObject)_paragraphBoundaryRegex
 {
-    return new RegExp(/^.[^\n\r]/m);
+    return new RegExp("^.[^\\n\\r]", "m");
 }
 
 + (JSObject)_whitespaceRegex
 {
     // do not include \n here or we will get cross paragraph selections
-    return new RegExp(/^.[ \t]+/m);
+    return new RegExp("^.[ \\t]+", "m");
 }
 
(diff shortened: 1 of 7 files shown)


AppKit/CPTextView/CPTextView.j
```diff
@@ -1848,18 +1848,18 @@ var kDelegateRespondsTo_textShouldBeginEditing
 
 + (JSObject)_wordBoundaryRegex
 {
-    return /(^[0-9][\.,])|(^.[^-\.,+#'"!§$%&/\(<\[\]>\)=?`´*\s{}\|¶])/m;
+    return new RegExp("(^[0-9][\\.,])|(^.[^-\\.,+#'\\\"!§$%&/\\(<\\[\\]>\\)=?`´*\\s{}\\|¶])", "m");
```
daboe01: i am a bit uneasy about the \\\" (triple backslash)
should't this be \" (one backslash for the JS parser, not the regex parser)?
not 100% sure as i did not test this, though.

HTPC-Schrauber: You are right. The triple backslashes are mistaken. No idea why I did this.
A single backslash is needed to escape the doublequote.

daboe01:
-#new
Tools
+#needs-review

daboe01:
+tools

cacaodev:
@mrcarlberg do you think we can merge this ?

mrcarlberg:
Merged! Thanks!

PR #2506, opened 2016-11-05
Fixed: CSSAnimations subviews animations removal

didierkorthoudt:
As PR #2502 fixed the x/y (or top/left) values computed during an animation (plus the introduction of _inhibitDOMUpdates), legacy Cappuccino layout methods can be used alone to get subviews correctly animated. Firing CSS animations for subviews in addition to CSS animation for the view, is no more necessary.


didierkorthoudt:
Replaced by #2507.

PR #2507, opened 2016-11-08
Fixed: CSSAnimations subviews animations removal

didierkorthoudt:
As PR #2502 fixed the x/y (or top/left) values computed during an animation (plus the introduction of _inhibitDOMUpdates), legacy Cappuccino layout methods can be used alone to get subviews correctly animated. Firing CSS animations for subviews in addition to CSS animation for the view, is no more necessary.


didierkorthoudt:
This PR replaces #2506.

cacaodev:
This PR replaces css animation with requestAnimationFrame() based animation for subviews.
We will need a test that proves that the performance is not degraded.
Also, one of the advantages of css animations is that they occur in a different thread, meaning they can be  concurrent with any process occurring on the main js thread.

didierkorthoudt:
Hi @cacaodev ! Well, I've search for such a test but it's not easy... As an animation has a fixed duration, how to measure this ?

BUT... I've found that, before performance, animation reliability could be a more accurate measure.

So I've got 2 little test apps, one without #2507, the other with. 

Performances are almost the same, the second (with #2507) being a little, little bit speedier. BUT, the true argument are subviews animations fidelity.

Without #2507, you animate each subview with a separate CSS animation, not truly bound to the main view animation. Result can be special as you can see here : 

http://www.intranet-ulg-ac.be/CA/without2507/

With #2507, only the main view is animated with CSS, subviews being animated with native cappuccino methods. You can see that this is more accurate than the previous one : 

http://www.intranet-ulg-ac.be/CA/with2507/

If you have any idea on how to compare performances, please tell me. I've tried (you can see a little light grey square sliding on top of the screen) but, as said before, it's very very similar on both versions.

cacaodev:
Thanks for the test. If you want to, they can be unified and included in the repository.

The tests show that the subviews intermediate frames are incorrect. Can we fix that ?

I would prefer because the WithPR test is shaky, the other is smooth, on my 6 years old macmini.

didierkorthoudt:
Well, I don't know how I could unify those 2 versions because they each use a different version of the framework...

I'll search for a fix but I'm not confident in finding one as animations are all independent... But I may have an idea. Let me check.

daboe01:
-#new
+AppKit
+#needs-improvement

mrcarlberg:
@didierkorthoudt What is the status on this?

didierkorthoudt:
@mrcarlberg I'll come back to Cappuccino in about 3 weeks and will continue on this. Sorry for the delay.

daboe01:
moved to #3038
+#duplicate

Issue #1506, opened 2012-04-16
Inconsistent Tooltips display across browsers


BlairDuncan:
It looks like the width is only being calculated correctly in Safari. 
Screen caps from the 3 different browsers:

http://www.uhoo.us/TooltipIssue/Safari.png
http://www.uhoo.us/TooltipIssue/FF.png
http://www.uhoo.us/TooltipIssue/Chrome.png

PR #2508, opened 2016-11-10
Fixed: text-sizing by canvas and via DOM sizing differed by 1px

daboe01:
previously, text was wrapped sometimes because the sizing from CPStringDrawing.j differed from the internal sizing of the DOM by 1px in some circumstances. This caused wrapping e.g. in segmented controls. See also https://github.com/cappuccino/cappuccino/issues/1506

This PR fixes the issue by adding 1px to the CPStringDrawing.j  textsizing.
This approach has also been taken in the predecessor implementation of textsizing (https://github.com/cappuccino/cappuccino/commit/48276f989fe1a3f6f6f66cb4cfdda35bd295f318)


cacaodev:
Naive questions:
1) Are you able to reproduce this bug in the browser with pure javascript/html code ?
2) Do you see a difference when comparing the result in different browsers ?
3) What happens when you CEIL the value, does it solve the problem ?

daboe01:
@cacaodev you are 100% right. This fix is flawed as i am having issues with the CPTextView now. Will definitely try CEIL this weekend and report back.

daboe01:
@cacaodev  ad 3): CEIL indeed solves the prob better than adding 1px.

daboe01:
@cacaodev ad 2): CEIL gives same result in FF, Chrome and Safari

daboe01:
the ceil solution does not seem to have issues with CPTextView

cacaodev:
This will be merged soon but I would feel more comfortable with a manual test that shows the bug & the fix so we can see the result in various browsers. And a pure JS reduction in the same test to check when / if this is fixed in the browser engine.

daboe01:
you can see the issue in the segmented control manual test (the segments wrap without this fix).
the pure JS reduction is here: http://jsfiddle.net/hzyw6b4f/1/

see also: http://stackoverflow.com/questions/4308989/are-the-decimal-places-in-a-css-width-respected

Dogild:
@cacaodev there is no point to keep `_sizeWithFont`, it will just confused future developer and user to have a private method who is only showing an issue in the framework ;)

daboe01:
@dogild do you propose to revert to the traditional DOM only sizing (+1px) and moving the canvas sizing to the typesetter?
this would abrogate the enormous speed gain from the current canvas-based sizing method.

daboe01:
@Dogild _sizeWithFont is needed to prevent ceiling in CPTextView. Ceiling in CPTextView would e.g. break ligatures. Ceiling is needed, however, whenever we need to calculate DOM space.

daboe01:
+TextView

mrcarlberg:
@Dogild @cacaodev Have you got the answers you need for this or should @daboe01 provide more information?

cacaodev:
If the problem is in the width calculation, is it with the canvas method or with the "DOMMetricsElements" method ? And can we fix it there, meaning at a lower level than what this PR does ?

daboe01:
this is my understanding:
1. the width calculation from canvas is ok.
2. the problem arises when floats (aka sub-pixels) are assigned to the DOM css width propery.
3. the floats are effectively ceiled so that the text needs more space than advertised. You can see this here: http://jsfiddle.net/hzyw6b4f/1/

cacaodev:
Tx. To be honest I don't really get the relation between the "fractional width ceiled in elements" exposed in the jsfiddle and our problem but I had only a quick look. My question remains: can we fix that in CPPlatformString + (CGSize)sizeOfString:(CPString)aString withFont:(CPFont)aFont forWidth:(float)aWidth and only for browsers affected.

daboe01:
IMHO, we cannot fix this transparently in + (CGSize)sizeOfString:(CPString)aString withFont:(CPFont)aFont forWidth:(float)aWidth
This is because the higher accuracy from the canvas method is badly needed in CPTypesetter and at the same time breaks the DOM sizing prediction (as exposed in the fiddle).
This PR adds a private high-accuracy API for CPTypesetter and makes the public API behave as before the CPTextView merge.
I see identical results in chrome, safari and FF.

cacaodev:
I don't fully understand the problem here and the relation with ceiling exposed in http://jsfiddle.net/hzyw6b4f/1/ but I don't want to block this PR and the result clearly fixes the bug.
@daboe01  @mrcarlberg if you are ok with this, let's merge it.

daboe01:
+#ready-to-commit
-#needs-improvement

daboe01:
@mrcarlberg IMHO it is time to merge. this is an improvement over the current situation and even over the situation before the CPTextView merge (we did stupidly always add 1px back then)

mrcarlberg:
Merged! Thanks!

PR #2511, opened 2016-11-12
New: CPObject -instancesImplementSelector:(SEL)aSelector

cacaodev:
Tests if instances of given class implement a selector.
With test.

The naming is from cocoa where this method is semi-public, id est not documented but not underscored.

A possible usage is to compute in CPView the _viewFlags (custom layout & drawing) at the class level instead of the instance level and make them more accessible from outside.

### DIFF
--- a/Foundation/CPObject.j
+++ b/Foundation/CPObject.j
@@ -275,13 +275,13 @@ CPLog(@"Got some class: %@", inst);
 }
 
 /*!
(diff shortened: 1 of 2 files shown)


daboe01:
looks good to me.
+#ready-to-commit
+Foundation

daboe01:
-#new
Foundation
+#ready-to-commit

Dogild:
Thanks

PR #2514, opened 2016-11-13
Fixed: CPTypesetting used method call instead of function call in inner loop

daboe01:
caching the method-function and calling it directly should improve performance when typesetting large amounts of text.


AppKit/CPTextView/CPTypesetter.j
```diff
@@ -319,7 +320,7 @@ var CPSystemTypesetterFactory,
         measuringRange.length++;
 
         var currentCharCode = theString.charCodeAt(glyphIndex),  // use pure javascript methods for performance reasons
-            rangeWidth = [theString.substr(measuringRange.location, measuringRange.length) sizeWithFont:currentFont inWidth:NULL].width + currentAnchor;
+            rangeWidth = sizingFunction(theString.substr(measuringRange.location, measuringRange.length), nil, currentFont, NULL).width + currentAnchor;
```
mrcarlberg: For this function call to correct you need to pass the selector (in this case ```@selector(sizeWithFont:inWidth:)```) as the second argument instead of ```nil```. 

The reason for this is that a method implementation function looks like this. ```function(self, _cmd, arg1, arg2...)```. So if you send ```nil``` as the second argument ```_cmd``` will be ```nil```. ```_cmd``` is seldom used so it will work most of the time but will not be 100% correct.

mrcarlberg:
Have you made any performance measurements so you know this will make it faster? Sometimes the compiler is smart enough to optimise things so there is no real world performance gain. This can be different with different browsers as they have different compilers.

daboe01:
i just compared this to the latest master in chrome. no subjectively noticeable speed gain when pasting the full cptextview source code into the manual test.

daboe01:
+AppKit
+#needs-info

daboe01:
+AppKit
+#needs-info
-#new

daboe01:
+TextView

mrcarlberg:
I don't think it is a good idea to merge this "speed improvement" if we can't show that it is a real speed improvements. We don't want complex strange code that is not necessary.

PR #2518, opened 2016-12-02
New: Allow a text based control to stay on a single line when text wraps.

cacaodev:
A private flag have been added in _CPImageAndTextView. It must be set
to YES when it does not make sense for a control to hold multiline text.
This is now the case for all CPSegmentedControl.

Test: In CPSegmentedControlTest, hit « Add Segment with flexible
width » button. The new label should be vertically centered.


daboe01:
-#new
+AppKit

mrcarlberg:
Thanks! Merged!

PR #2519, opened 2016-12-02
Fixed:  TextView caret blinked during typing

daboe01:
previously, blinking was turned on too early. This is fixed by this PR.


daboe01:
+TextView
-#new
+#ready-to-commit

PR #2527, opened 2017-01-04
fixed: README.markdown was inconcise

daboe01:
moved XCode from requirements to the introduction. removed cygwin (too much detail for the frontpage IMHO)


mrcarlberg:
Everyone, what are your thoughts on this? Should we do this change in the readme file?

Dogild:
It should be something else IMO, like we have a XcodeCapp plugins which brings the integration with Xcode and Interface Builder

PR #2535, opened 2017-03-15
Adding support for the node engine

FrankReh:

PR #2538, opened 2017-05-04
Adding support for the node engine

FrankReh:
Replaces the first pull request: Adding support for the node engine #2535


mrcarlberg:
It looks like we don't need these changes anymore. I'll leave this open until I have verified this a little more.

daboe01:
-#new
+#needs-review
+Tools

mrcarlberg:
Yes, this can be closed now. It is not needed.

PR #2543, opened 2017-05-04
Remove objj_class.sub_classes property

FrankReh:
From the first day this class was added to the repo, the sub_classes
property, which is initialized with an empty array, has gone unused.

Removing it.  A little less code.  A little subsequent processing
and memory used.


mrcarlberg:
It looks like the property is not used as Frank has pointed out. Great catch!

mrcarlberg:
Merge! Thanks!

PR #2546, opened 2017-05-12
CPSplitView -trackDivider: performance

cacaodev:
Perf: Exit a loop when the current divider is found.

Fixed: Do not start tracking when the mouse is not over a divider (see mouseDown:)

Fixed: Properly catch single-click, double-click, dragging on a divider.
Previously, 
• a double-click was also catched as a single-click on mouse up,
• dragging was handled in mouse down before tracking actually begins.

Manual test: All split views can be collapsed when double-clicking the first divider.

### DIFF
--- a/AppKit/CPSplitView.j
+++ b/AppKit/CPSplitView.j
@@ -474,7 +474,6 @@ var ShouldSuppressResizeNotifications   = 1,
 
     return CPNotFound;
 }
-
 /*
     Tracks the divider.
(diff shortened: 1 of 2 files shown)


daboe01:
-#new
+AppKit

mrcarlberg:
@cacaodev Do you have anything more before I merge this?

cacaodev:
No, ready to go.

PR #2547, opened 2017-05-30
Fix exceptions in CPURLConnection and CPPredicateEditor

davidsiaw:
CPPredicateEditor would throw an exception if a CPCompoundPredicate was set as the object with no children. The code used to assume that if a CPPredicate had no children it was a CPComparisonPredicate. This is wrong. I had it do a type check instead.

CPURLConnection throws errors because credentials can only be set in the OPENED or SENT state.

Added some messages for the compiler to clear things up about images not being sprited.


mrcarlberg:
Thanks for submitting these improvements to the Cappuccino project.

To my understanding there are two fixes here. One for CPURLConnection and one for CPPredicateEditor. It would help us a lot if you create a pull request for each fix so we can discuss and accept them separately.

Second, it would also help a lot if your commit messages where following the standard. Please read this document: [CONTRIBUTING.md](https://github.com/cappuccino/cappuccino/blob/master/CONTRIBUTING.md)

cacaodev:
The bug seems to be in ```CPPredicateEditorRowTemplate - (CPArray)displayableSubpredicatesOfPredicate:(CPPredicate)predicate```
The doc says it should return the subpredicates of a compound, so it should even if the array is empty (currently in capp returns nil if subpredicates.count == 0).
Would that fix the problem for you ?

davidsiaw:
Understood. I'll separate the fixes into separate requests and try and follow the standard

PR #2557, opened 2017-08-23
fixed: cursor blinks for placeholder

daboe01:
previously, the cursor would blink when a binding-placeholder was shown. this cannot happen after this PR anymore.


daboe01:
previously, the cursor would blink when a binding-placeholder was shown. this cannot happen after this PR anymore.

daboe01:
+#ready-to-commit

daboe01:
-#new
+#ready-to-commit

PR #2558, opened 2017-08-29
Fixed: smart copy/paste did not work properly [+1]

daboe01:
previously, the smartness of copy/paste was derived from the last click-count.
this behavior is incorrect: the smartness should be based on the click-count of the original copy/cut event.
this error is fixed by this PR, that introduces a private pboard-type for this purpose (so does apple)


daboe01:
+#ready-to-commit

daboe01:
-#new

daboe01:
+1

mrcarlberg:
Merged! Thanks!

PR #2559, opened 2017-09-07
Fixed: CPTextView's onpaste handler used non-standard window.event

daboe01:
without this PR, FF breaks on paste from the native pasteboard
this is fixed by avoiding the use of (non-standard) window.event that seems to have been removed from FF recently


daboe01:
-#new
+#ready-to-commit

PR #2561, opened 2017-09-11
Full support for source map when debugging

mrcarlberg:
We can now debug and step inside the Objective-J source file thanks to new support for source maps.
The Cappuccino framework will now compile with source maps for the debug version. The Release version is not changed.

To turn this option on for your project add the ```-S``` flag in your Jakefile. Search for the ```setCompilerFlags``` function to find where the compiler options are set.

In the Browser you need to add ```SourceMap``` as compiler option to your index-debug.html file. The line should look something like this:
```OBJJ_COMPILER_FLAGS = [... , "SourceMap"];```

All browsers support source maps but at the moment Chrome as the best support.

Here is a short clip on what it looks like when stepping in the browser: http://mini.carlberg.org/dev/Objective-J-Debugging.mov

### DIFF
--- a/AppKit/Jakefile
+++ b/AppKit/Jakefile
@@ -37,7 +37,7 @@ appKitTask = framework ("AppKit", function(appKitTask)
         appKitTask.setCompilerFlags("-O2 " + INCLUDES);
     }
     else
-        appKitTask.setCompilerFlags("-DDEBUG -g -S " + INCLUDES);
+        appKitTask.setCompilerFlags("-DDEBUG -g -S --inline-msg-send " + INCLUDES);
 });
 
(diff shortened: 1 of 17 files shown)


daboe01:
-#new
+Tools

mrcarlberg:
I have found two bugs:

1. #if/#else/#endif can mess up line numbering so the lines afterward will be off. I'm working on a fix for this.
2. Sometimes Chrome will set a breakpoint on the wrong line. It works on other browsers so this is probably a thing in Chrome. I have filed an issue for this here: https://bugs.chromium.org/p/chromium/issues/detail?id=761263
Please go in an write a comment about it so they see that more people are affected by this.

mrcarlberg:
I have now committed changes that will fix bug number 1 above. Bug number 2 has to be fixed by Chrome.

Please test this so we can merge it into the master

mrcarlberg:
Merged!

PR #2563, opened 2017-09-14
Fixed: Editing text in tableviews throws DOM exceptions under certain conditions in Chrome

daboe01:
Previously, editing text in tableviews or combo boxes sometimes threw Failed to execute 'removeChild' on 'Node': The node to be removed is no longer a child of this node...

This only happens in chrome and seems to be due to an internal race condition. This annoying bug is worked around with this PR


daboe01:
-#new
+#needs-review
+Chrome

mrcarlberg:
This will remove the exception for sure. But why do we get this exception? Is it really a bug in Chrome that does this? Or is it something that we do?

If it really is a bug in Chrome we should file an issue for it. Also, what if we get an exception for some other reason, should we really silence that?

I think we should be very carefully with quick fixes like this.

daboe01:
the issue is discussed here:
https://stackoverflow.com/questions/21926083/failed-to-execute-removechild-on-node

daboe01:
IMHO the try/catch approach is cleaner and more maintainable than introducing yet another global flag to signal that we are in-blur

mrcarlberg:
I don't think a try catch is a clean way. What if we get an exception for some other reason. It will be silenced and it will be very hard to track down what is happening.

There must be some other way to do this. Maybe check if it is already removed or even a global flag.

I can't see that we can accept the proposed fix.

Any suggestions?

daboe01:
@mrcarlberg do you think we can fix it this way?

jasperdeb:
Will this PR be merged in the main source? 
Since my recent upgrade from 0.9.9 to 0.9.10 I have one particular CPOutlineView where I am experiencing this error. Quite annoying error...

daboe01:
@jasperdeb can you please confirm that this PR fixes the issue for you?

daboe01:
+#needs-improvement

daboe01:
@jasperdeb thank you! sorry to hear that. do you know a fix that works without try/catch?

jasperdeb:
If I comment out the following text, the error does not occur.

870 /*if (CPTextFieldInputIsActive)
871         element.blur();
872 
873     if (!CPTextFieldInputDidBlur)
874         CPTextFieldBlurHandler();
875 
876     CPTextFieldInputDidBlur = NO;
877     CPTextFieldInputResigning = NO;*/

As I am completely no Javascript developer (apart from Cappuccino) and am therefore not sure, but I assume that this code is not necessary as the removeChild already triggers the blur method?

jasperdeb:
523ebfe does not solve the problem.

Actually putten the variables 

CPTextFieldInputDitBlur = NO
CPTextFieldInputResigning = NO 

after 

if (element.parentNode == _DOMElement)
          element.parentNode.removeChild(element);

instead of before also solves the error.

(Is it possible for me to add a commit myself to this pull request?)

daboe01:
@jasperdeb do we have the solution with a5aa0c9 eventually?
@mrcarlberg can you accept this?

jasperdeb:
Indeed, this eliminates the error for me.

daboe01:
@jasperdeb thank you very much for finding the proper solution to this tricky problem.

mrcarlberg:
@daboe01 Can you please add some comments in this method why we need to set these variables after the removal of the child element. Please reference the stack overflow discussion also as you did before. The more info you provide the better.

daboe01:
@mrcarlberg thank you for this excellent suggestion. I just added some comments.

daboe01:
-#needs-review
+#ready-to-commit

mrcarlberg:
Ok, I have tested this and it looks like this will fix the problem.

But, we still have the same problem with CPTokenField. It inherit CPTextField but has its own implementation of this in resignFirstKeyResponder:

This must also be fixed.

daboe01:
excellent catch! i just addressed this with ae52809

AppKit/CPTokenField.j
```diff
@@ -528,12 +528,18 @@ CPTokenFieldDeleteButtonType     = 1;
     if (!CPTokenFieldInputDidBlur)
```
mrcarlberg: Before this line we will try to blur the input field. Should this not be under a ```if``` statement as in the CPTextField counter part? Otherwise we should be able to remove the variable ```CPTokenFieldInputIsActive``` as it is never read?

This is very complex part and I feel that I don't really understand the whole picture of it. I don't know if this is needed or not but maybe it should be better to have the same functionality as the CPTextField counterpart.

daboe01: @mrcarlberg i do not understand either.
as nobody complained about DOM exceptions from tokenfields,  my feeling is to be on the safe side and revert ae52809.
do you agree?

mrcarlberg: It is easy to get it to crash with a token field. That is how I found the problem.

mrcarlberg: I think we should have the same behavior in both the text field and token field. They should work in the same way as it will be easier for us to understand what is going on when we look at this in a couple of years.

daboe01: then i suggest to add the missing if statement.

mrcarlberg: This is complex and I really don't know why we do this if in the text field in the first place, but I think it will be safer and easier to understand to have the same functionality in both classes.

mrcarlberg: Yes, please add the ```if```

daboe01: we really need to rework all this on the basis of a CPTextView-based fieldeditor :-)

mrcarlberg:
Merged! Thanks!

PR #2572, opened 2018-01-03
Fixed: Nib2Cib will replace a CPTextFieldCell with a CPTextField

mrcarlberg:
As Cappuccino does not use cells a binding or connection made to a ```CPTextFieldCell``` will get lost when the Cib file is read as the cell is never stored in the Cib file.

This fix make sure that the binding / connection is made to the ```CPTextField``` instead of the cell by replacing the cell with the text field when reading the xib file. The same mechanism is used in ```NSButton``` and ```NSMatrix```.

This fix allows view based TableView bindings from a text field cell to the table cell view in the Interface Builder in Xcode.

This works in Cocoa.


mrcarlberg:
Merged! Thanks!

PR #2577, opened 2018-01-12
Fixed: Flatten now works with Node.js and JavaScriptCore

mrcarlberg:
This is excellent work done by @FrankReh with some small bug fixes by me.

The ToolsTest.j test case will not work now as it has to be run with Node.js

### DIFF
--- a/CommonJS/bin/flatten
+++ b/CommonJS/bin/flatten
@@ -1,10 +1,12 @@
-#!/usr/bin/env objj
+#!/usr/bin/env narwhal
 
-require("narwhal").ensureEngine("rhino");
+// Copy of the flatten objj script but this has been made a narwhal script.
+//
+// This works with any ES6 narwhal engine (aka jsc and node).  For now, the
+// only ES6 aspect of this script is the use of a single multiline template
+// string.
 
-@import <Foundation/Foundation.j>
-
-@import "../lib/cappuccino/objj-analysis-tools.j"
+var ObjectiveJRuntimeAnalyzer = require("../lib/cappuccino/objj-analysis-tools.js").ObjectiveJRuntimeAnalyzer;
 
 var FILE = require("file");
 var OS = require("os");
@@ -111,7 +113,8 @@ function main(args)
     flattener.load(mainPath);
 
     print("Loading default theme.");
-    flattener.require("objective-j").objj_eval("("+(function() {
+    // Transpile some objective-j code to js and then run it.
+    flattener.objj_eval(`(function() {
 
         var defaultThemeName = [CPApplication defaultThemeName],
             bundle = nil;
@@ -123,8 +126,7 @@ function main(args)
 
         var blend = [[CPThemeBlend alloc] initWithContentsOfURL:[bundle pathForResource:defaultThemeName + @".blend"]];
         [blend loadWithDelegate:nil];
-
-    })+")")();
+    })`)();
 
     var applicationJSs = flattener.buildApplicationJS();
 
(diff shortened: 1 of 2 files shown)


daboe01:
milestone=1.0
@mrcarlberg can you have a look why travis fails?
+#needs-improvement

mrcarlberg:
ToolsTest.j will fail as this new version of Flatten must use Node or JavaScriptCore to work. But if we use Node or JavaScriptCore the test case will also fail as Press can nor yet run on Node or JavaScriptCore.

So we are waiting for Press to be able to run on Node or JavaScriptCore. As of now I have not had the time to look at this. If someone want to take a look feel free to do so. The solution for Press is similar to this fix for Flatten.

daboe01:
milestone=1.1

daboe01:
-#new

daboe01:
@mrcarlberg can we proceed with this one, as we now have press for node in https://github.com/cappuccino/cappuccino/pull/3125?

daboe01:
why not write a Vite Plugin?

```javascript
import { defineConfig } from 'vite';
import { execSync } from 'child_process'; // Node.js module to run commands
import path from 'path';
import fs from 'fs';

/**
 * A custom Vite plugin to transpile Objective-J (.j) files to JavaScript.
 */
function objectiveJPlugin() {
  return {
    name: 'vite-plugin-objective-j',

    // The 'transform' hook is called for each file Vite processes.
    transform(code, id) {
      // We only care about files ending in .j
      if (!id.endsWith('.j')) {
        return null; // Let other plugins or Vite handle it
      }

      console.log(`Transforming Objective-J file: ${id}`);

      try {
        // Use the Objective-J compiler to transpile the file.
        // We pass the file path to the compiler. A temporary file could also be used.
        // The `-c` flag might mean "compile" and `-p` for "print to stdout".
        // YOU WILL NEED TO FIND THE CORRECT `objjc` COMMANDS.
        const jsCode = execSync(`objjc -p ${id}`, { encoding: 'utf-8' });

        // Return the transpiled JavaScript code to Vite.
        // Vite will now bundle this JS as if it were a normal .js file.
        return {
          code: jsCode,
          map: null, // Source maps could be generated if the compiler supports it
        };
      } catch (e) {
        // Throw an error if compilation fails
        this.error(`Failed to compile Objective-J file: ${id}\n${e.stderr}`);
      }
    }
  };
}

// vite.config.js
export default defineConfig({
  plugins: [
    // Add our custom plugin to the pipeline
    objectiveJPlugin(),
  ],
  resolve: {
    // Teach Vite how to resolve @import statements from Cappuccino
    alias: {
      // This is an example. You'd map the logical framework paths
      // to their actual locations on disk.
      'AppKit': path.resolve(__dirname, 'Frameworks/AppKit'),
      'Foundation': path.resolve(__dirname, 'Frameworks/Foundation'),
    }
  },
  build: {
    // Vite handles minification, chunking, etc., automatically here.
  }
});
```

todos: 
Re-create the logic of the `--framework` flag by setting up path aliases in `vite.config.js`. This will allow `@import <Foundation/CPObject.j>` to be resolved correctly.
Modify  `index.html` to load the main application file (e.g., `main.j`) as a module script. Vite and your plugin will handle the rest.
Run `npm run build`.

**Conclusion:**

While the old script is a clever, self-contained tool, it is a product of its time. **Vite can absolutely replace it**, and the benefits would be enormous:
*   **Speed:** Vite's development server is orders of magnitude faster.
*   **Ecosystem:** You gain access to the entire modern npm ecosystem of tools and libraries.
*   **Maintainability:** Your build logic becomes declarative configuration instead of a complex imperative script.
*   **Modern Features:** You get Hot Module Replacement (HMR), optimized production builds, and an easy path to using modern JS/CSS features.

The migration requires a one-time effort to create the transpilation plugin, but the long-term payoff in developer experience and performance is well worth it.

PR #2579, opened 2018-01-15
Fixed: The property list test cases can now run on the Node engine.

mrcarlberg:
The xml parser in the Node.js engine is more modern and can handle more cases. Our property list parser does not handle this. This fix makes the test cases for the property lists run on the Node.js engine.


daboe01:
-#new
+tools
+#ready-to-commit

mrcarlberg:
Merged! Thanks!

PR #2582, opened 2018-01-20
fixed: TextView bindings wrote data on blur even when not editable

### LINKED ISSUE #2574: CPTextView Edits Any Time Window Resigns Key/Main
I have a CPTextView with binding and even though it is non enterable and hasn't been touched in any way it acts as if it has been edited (i.e. updates bound value & send delegate notifications) any time the window containing it resigns key and/or main (haven't tested each condition, know that it has to be one or the other). Trivially reproducible.

daboe01:
previously, the objectValue was 'reversed' upon blur even if nothing changed.
This PR only reverses the objectValue when the current value in the view differs from the value in the model.

fixes #2574


cacaodev:
Is there a visible bug related to this except from a performance issue and is the current behavior different from CPTextField ?
If not, I prefer that we keep this bug for the moment and fix this later using the CPEditorRegistration protocol. I believe this is where it is handled in cocoa.

daboe01:
@cacaodev the issue is real (see #2574). i am seeing this myself when binding to read-only textviews. currently i am working around this with ugly patches in my app.

daboe01:
-#new
+TextView
+bindings

daboe01:
+#ready-to-commit

mrcarlberg:
I agree with @cacaodev that this needs to be addressed using the ```CPEditorRegistration``` protocol.

But I also agree that this should be fixed this way until someone implements it with the ```CPEditorRegistration``` protocol.

We can't just commit this without comments that this is a fix and should later be fixed using the ```CPEditorRegistration``` protocol.

So @cacaodev is it ok for you if this is a temporary fix until it is solved in a proper way?

@daboe01 please add some comments that this is a temporary fix etc. in the code.

cacaodev:
@mrcarlberg @daboe01 OK with that but we also need a confirmation that the same behavior is happening in other editable controls like CPTextView and is related to a bug in the bindings system.

daboe01:
@mrcarlberg i downgraded this PR to only prohibit reverse bindings in non-editable textviews. this is a trivial improvement without architectural strings attached.

mrcarlberg:
Merged! Thanks!

PR #2584, opened 2018-01-23
Fixed: CPComparisonPredicate does not handle nil inequality correctly

aksuska:
Fixed issue where CPComparisonPredicate always evaluated to NO anytime the inequality
operator was used with a nil operand. I.e. value != nil would evaluate to NO even if value is actually nil. Reduction for bug and comparison to Cocoa functionality available on request.

MAINTAINERS: as my first contribution, please let me know if there is anything I can do better with commits/messages or anything.


aksuska:
Sorry closing as I screwed it up

PR #2585, opened 2018-01-25
 Fixed: CPComparisonPredicate does not handle nil inequality correctly.

aksuska:
Fix and test for a bug where CPComparisonPredicate incorrectly evaluates inequality comparisons with a nil operand (e.g. value != nil) to false (NO) when they should be true (YES).


cacaodev:
@aksuska Sorry, I missed this one. Do we still need it ?

aksuska:
Yes, it is. I kept getting combined pull requests so I cleaned everything out and started from scratch, although I have not re-submitted this pull request as I was waiting to hear what process is needed to submit multiple pull requests.

PR #2587, opened 2018-02-24
FIXED: _CPRTFParser  called non existing factory method of CPNumber

daboe01:
see title.


daboe01:
+#ready-to-commit

mrcarlberg:
This is an ok fix.

Will add that it can be arguable that this code is ok and the method is missing for the ```CPNumber``` class. Cocoa do has this method and so should Cappuccino.

So if someone wants to add this method and some other missing ones for ```CPNumber``` please feel free.

mrcarlberg:
Merged! Thanks!

PR #2588, opened 2018-02-24
Fixed: CPTextContainer called an nonexistant method

daboe01:
this PR removes that particular call, that is superfluous anyway.


daboe01:
+#ready-to-commit

mrcarlberg:
Merged! Thanks!

PR #2596, opened 2018-03-07
Fixed: Crash when triple click an empty CPTextField that is not editable but selectable

mrcarlberg:
This call will crash as ```_DOMTextElement``` is ```nil```.

This check will make sure it is not called when no element is present. This is the same way as it is done in other parts of this class.


mrcarlberg:
Merged! Thanks!

PR #2598, opened 2018-03-10
 FIXED: return key was not swallowed by textview but triggered the default button.

### LINKED ISSUE #2580: CPTextView Does Not "Swallow" Return Key
When editing in a CPTextView, if you hit the enter/return key but you have a push button with the same key equivalent, the text view allows the keystroke to bubble up and triggers the push button. It should, as Cocoa does, "swallow" the keystroke so no other responder can act on it.

daboe01:
previously, the return key was unconditionally triggered in each runloop after the event was processed.
this PR introduces a flag that enables the responder chain to temporarily suppress this.

fixes #2580


daboe01:
-#new
+AppKit
+#needs-review

daboe01:
milestone=1.0

daboe01:
-#new

mrcarlberg:
Merged! Thanks!

PR #2600, opened 2018-03-14
fixed: key equivalent takes precedence over event tracking which is not the case in cocoa

### LINKED ISSUE #1953: CPPopUpButton doesn't disable button key equivalents while menu visible
Working with a CPPopUpButton, while the menu is up via the mouse, I can use ESC to invoke the action of a CPButton in the same view.

daboe01:
previously, key shortcuts 'stole' any event before modal loops could even look into it. this PR defers key shortcuts handling after processing the eventListeners as is the case in cocoa.
fixes #1953


daboe01:
-#new
+#needs-review
+AppKit
milestone=1.0

daboe01:
-#new
+#needs-review
+AppKit
milestone=1.0

mrcarlberg:
Merged! Thanks!

Issue #2545, opened 2017-05-06
Key Window Won't Order Front on Click


aksuska:
If a window is key but another window is ordered front, clicking on the key window will *not* bring it to front, but it should. Steps to reproduce: in a project wth two windows (not a platform window), open the first window using -makeKeyAndOrderFront: and then call orderFront: on the second. The first window title bar will still show as "active" and clicking on the window will not bring it to front. If, however, you navigate away form the browser and return, it will work.

cappbot:
**Milestone:** Someday.  **Label:** #new.  **What's next?** A reviewer should examine this issue.

daboe01:
i saw this when the control did not return YES for needsPanelToBecomeKey 
do you have a reduction?

daboe01:
-#new
+#needs-info
+AppKit

cappbot:
**Milestone:** Someday.  **Labels:** #needs-info, AppKit.  **What's next?** Additional information should be added as a comment to this isuse.

aksuska:
I have attached a reduction as a jaked project but without frameworks. When loaded, Window 1 will be made key and front. Click the button in the window to order Window 2 to front. It won't be key or main, of course, but if you click on Window 1 to bring to front you will see that it won't. You have to click on Window 2 to make it key/main or switch to a different application and back to the browser (both windows will be not key or main) and then it will allow you to bring Window 1 to front.
[Archive.zip](https://github.com/cappuccino/cappuccino/files/1632759/Archive.zip)

Here is the source:
[Source.zip](https://github.com/cappuccino/cappuccino/files/1632768/Source.zip)

daboe01:
thank you.
i am seeing this also.
interestingly, the issue goes away as soon as the platform re-becomes key window.
-#needs-info
+#acknowledged
+#needs-patch

cappbot:
**Milestone:** Someday.  **Labels:** #acknowledged, #needs-patch, AppKit.  **What's next?** This issue needs a volunteer to write and submit code to address it.

daboe01:
@mrcarlberg can you close this please?

mrcarlberg:
This is fixed by #2602

PR #2602, opened 2018-03-18
FIXED: click to key window in the background did not bring it forward

daboe01:
previously, any click to a key window was simply forwarded to the clicked view without bringing the window to front. is is not the case in cocoa. here, the key window is always ordered front when clicked.

this PR introduces a check whether the keyWindow is the frontmost window and issues a makeKeyAndOrderFront unless the window is already in front.

fixes issue #2545


daboe01:
-#new
+#needs-review
+AppKit
milestone=1.0

cacaodev:
It's okay, I also missed it. Just don't be surprised if I ask more tests before merging any PRs from anyone in the future.

Issue #2601, opened 2018-03-17
NSColor Method variable redeclared in case statement.


bedalton:
I was trying to resolve some variables in a plugin I'm working on, and noticed that `result` variable which is declared at the start of the method [NSColor NS_initWithCoder:] is redeclared on line 96 within the case block:
```
case 6: // named color
            var catalogName = [aCoder decodeObjectForKey:@"NSCatalogName"],
                colorName   = [aCoder decodeObjectForKey:@"NSColorName"],
                color       = [aCoder decodeObjectForKey:@"NSColor"],
                result      = nil;
```
`result` is returned only at the end of the method, outside the `case 6: //named color` scope.

I'm not sure if this causes any problems, and I'm not too sure how I would test it, but I thought I would point it out.

cacaodev:
Hi @bedalton,
Thanks for pointing this out. You are right, this redeclaration seems unwanted.
The easiest way to get this fixed quickly is to open a pull request.

bedalton:
The plugin I made now analyses redeclared variables, and I've found quite a few in the framework, should I make a pull request where I change all of them? Or just one at a time, or just bring each one to attention first?

bedalton:
Are redeclared variables ever desirable?

daboe01:
hi @bedalton, please do it one at a time.

bedalton:
Oh okay, I didn't know that. Should I still go through the other similar cases, or not?

mrcarlberg:
I think it is good to change this if you want to spend the time on it. It is confusing when variables are redeclared. So, yes, please do it.

daboe01:
+#wont-fix

cappbot:
**Milestone:** Someday.  **Label:** #wont-fix.  **What's next?** A reviewer or core team member has decided against acting upon this issue.

cappbot:
**Milestone:** Someday.  **Label:** #wont-fix.  **What's next?** A reviewer or core team member has decided against acting upon this issue.

PR #2604, opened 2018-03-20
Fixed: Variable in if statement overshadows variable in enclosing block

bedalton:
If statement variable `instantiatedObject` on line 233 of
_CPCibObjectData.j overshadows a variable with the same name declared
on line 229 in the enclosing block. This variable is later added to array outside
the enclosing if statement, resulting in a potential loss of information.

Refs: #2601


mrcarlberg:
This fix will not change the behavior of the code. But, redeclarations of a variable can be confusing so this is a good change.

daboe01:
+#ready-to-commit

cacaodev:
Merged thanks.

Issue #2471, opened 2016-08-09
CPPopUpButton setTitle:@"" behaviour does not match NSPopUpButton setTitle:@"" behaviour


enquora:
Unless I'm completely misinterpreting the behaviour of NSPopUpButton setTitle:@"" (test case attached [Popup Test.zip](https://github.com/cappuccino/cappuccino/files/408164/Popup.Test.zip)), the analogous CPPopUpButton is incorrect.

The NS behaviour _does not_ add an additional menu item when the title is set to @"", it simply sets the display title to @"", wheres CP does.

```
    if ([self pullsDown])
    {
        var items = [[self menu] itemArray];

        if ([items count] <= 0)
            [self addItemWithTitle:aTitle];

        else
        {
            [items[0] setTitle:aTitle];
            [self synchronizeTitleAndSelectedItem];
        }
    }
    else
    {
        var index = [self indexOfItemWithTitle:aTitle];
        if (index < 0)
        {
            [self addItemWithTitle:aTitle];
            index = [self numberOfItems] - 1;
        }
        [self selectItemAtIndex:index];
    }
```

This appears incorrect, and certainly results in different (and problematic) visual results when one is using the popup as a momentary select button with a label elsewhere.

cappbot:
**Milestone:** Someday.  **Label:** #new.  **What's next?** A reviewer should examine this issue.

daboe01:
@enquora 

from https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSPopUpButton_Class/#//apple_ref/occ/instm/NSPopUpButton/setTitle:

If the receiver displays a pop-up menu, this method changes the current item to be the item with the specified title, ***adding a new item by that name if one does not already exist***. If the receiver displays a pull-down list, this method sets its title to the specified string.

can you please provide a minimal mac application that demonstrated the (undocumented) extra-path for @""?

daboe01:
-#new
+#needs-info

cappbot:
**Milestone:** Someday.  **Label:** #needs-info.  **What's next?** Additional information should be added as a comment to this isuse.

daboe01:
i just tried your example and indeed, the capp implementation seems incorrect.
-#needs-info
+#accepted
+AppKit

cappbot:
**Milestone:** Someday.  **Labels:** #accepted, AppKit.  **What's next?** A reviewer should examine this issue.

[...]

PR #2607, opened 2018-03-24
 fixed: deviation from cocoa when calling setTitle:''

daboe01:
previously, CPPopUpButton added the empty string to the item array. Cocoa clears the title and sets objectValue to -1.
this PR makes CPPopUpButton behave as in cocoa.
fixes issue #2471


cacaodev:
Looks good to me.
Do you know if in cocoa this is also the behavior when using bindings. If you bind the content binding to an array containing @"" and select it, does it automatically set the selected index to -1 ?

daboe01:
cocoa inserts empty strings into the item lists.
bildschirmfoto 2018-03-31 um 17 32 46

```ObjC
    NSMutableDictionary *value = [[NSMutableDictionary alloc] init];
    [value setObject:[NSString stringWithFormat:@""] forKey:@"name"];
    [self.myAC addObject:value];
    [self.myAC rearrangeObjects];
```

AppKit/CPPopUpButton.j
```diff
@@ -491,9 +491,17 @@ CPPopUpButtonStatePullsDown = CPThemeState("pulls-down");
 
         if (index < 0)
         {
-            [self addItemWithTitle:aTitle];
 
-            index = [self numberOfItems] - 1;
+            // this ist to match cocoa where setting an empty string does not add it but simply clears the title
+            // and sets objectValue to -1
+            if (aTitle === '')
+                [self selectItemAtIndex:-1];
```
cacaodev: What about `index = -1` ?

daboe01: i am not sure i understand you question 100%.

index = -1 means "this popupbutton has currently no selected item".

daboe01: i have to check tonight the value in cocoa with bindings.

daboe01:
-#new
+AppKit
milestone=1.0

AppKit/CPPopUpButton.j
daboe01: @cacaodev the object value seems to equal the selected index in cocoa when the popupbutton is filled via bindings, regardless whether the title is the empty string.

Issue #583, opened 2010-04-02
Bug: CPBorderlessBridgeWindowMask can be resized smaller than window min size


klaaspieter:
Steps:
1. Create Cappuccino running in the browser using a full bridged window
2. Set a min size on the full-bridged window
3. Resize the browser to less than the minimum window size

Result:
The window resizes and truncates the content.

Expected:
One of two things:
- The browser (or Cappuccino) to show horizontal scrollbars
- The browser window size is constrained to the minimal window size

Additional:
Of the two expected options I think the first is the best because it conforms to how a browser normally behaves.

PR #2609, opened 2018-03-26
fixed: setMinSize: had no effect in platform windows

### LINKED ISSUE #101: CPWindow setMinSize does not set overflow properly [+1]
CPWindow includes utilities to setMinSize and setMaxSize. While these seem to work correctly, when the user resizes the window to less than it&rsquo;s minimum size, the overflow of the window/contentView should be changes to add browser scroll bars into the window.

[original LH ticket](http://cappuccino.lighthouseapp.com/projects/16499/tickets/76)

 This ticket has 0 attachment(s).

daboe01:
previously, platform windows were clipped, even when the were resized below the minSize.
this PR makes platform windows show native scroll-bars as suggested by #583.

fixes #101


daboe01:
-#new
+#needs-review
+AppKit
milestone=1.0

daboe01:
i just realized that this PR has some issues.
mouse clicks are off, when the body is scrolled natively.
+#needs-improvement

daboe01:
now it works in my testing.
-#needs-improvement

mrcarlberg:
Merged! Thanks!

daboe01:
under super rare circumstances, theWindow._minSize.height seems to become off-by-one due to rounding issues. this causes a native scroll bar to appear only to scroll for a single pixel. not a big issue for me, but worth mentioning here.

PR #2612, opened 2018-03-30
fixed: touch-scrolling did not work

### LINKED ISSUE #1548: CPTableview / CPScrollview do not scroll on iOS
Attempting to scroll on safari / opera mini on iOS 5 using 2 fingers does not work. 

On an iPhone scrolling can be achieved only by dragging the row selection above/below the safari webview, however on an iPad the webview meets the bottom of the page so scrolling down (and therefore up) is impossible.

daboe01:
previously, two-finger events were silently ignored.
this PR converts these into cappuccino scrollEvents.

fixes #1548


daboe01:
works on an ipad pro

jasperdeb:
Also tested this on an iPad 
- tested on Chrome (65.0.3325.152) and Safari (iOS 11.0)
- works as well in a CPOutlineView as in a CPTableView.

Only issue I am now experiencing is a quite "fat" vertical scroller. But given the code that has been changed, I do not see this as an issue related to this PR?

screenshot

Good job @daboe01 (considering this to be an issue that is dragging on so long).  Important point solved as I am no longer to tell a tablet is not supported by a web app developed in 2017-18 ...

daboe01:
@jasperdeb thank you for testing. tell @didierkorthoudt about your thoughts on scroller in aristo3

PR #2613, opened 2018-03-31
NEW: Add CSS theming support

didierkorthoudt:
This first PR adds support for CSS theming. Aristo3 will be build upon this system.

The CSS theming proposed system is designed to be 100% compatible with current theming system. It is then possible to switch from a classic theme (like Aristo2) to a CSS theme (like Aristo3).

The main idea is to extend CPColor and CPImage with CSS components, and then add support for these components in low level UI components : CPImageView, CPView, _CPImageAndTextView.

CPTheme, CPThemeBlend and _CPObject+Theme are also adapted in order to provide supporting functionalities.

To create a CPColor that uses CSS, simply : 
```
buttonCssColor = [CPColor colorWithCSSDictionary:@{
                                                       @"background-color": A3ColorBackgroundWhite,
                                                       @"border-color": A3ColorActiveBorder,
                                                       @"border-style": @"solid",
                                                       @"border-width": @"1px",
                                                       @"border-radius": @"3px",
                                                       @"box-sizing": @"border-box"
                                                       }
                                    beforeDictionary:@{
                                                       @"background-color": @"rgb(225,225,225)",
                                                       @"bottom": @"3px",
                                                       @"content": @"''",
                                                       @"position": @"absolute",
                                                       @"right": @"21px",
                                                       @"top": @"3px",
                                                       @"width": @"1px"
                                                       }
                                     afterDictionary:@{
                                                       @"content": @"''",
                                                       @"bottom": @"3px",
                                                       @"right": @"6px",
                                                       @"top": @"1px",
                                                       @"position": @"absolute",
                                                       @"height": @"14px",
                                                       @"width": @"9px",
                                                       @"margin": @"2px 0px 2px 0px",
                                                       @"background-image": @"url(%%packed.png)",
                                                       @"background-position": @"0px -64px",
                                                       @"background-repeat": @"no-repeat",
                                                       @"background-size": @"100px 400px"
                                                       }];
```
As you can see, you can also deal with the after and before pseudo-elements. If you don't need them, just use ``buttonCssColor = [CPColor colorWithCSSDictionary:@{ xxx }];``.

The exact same principle is used for CPImage but, of course, you have to specify a size : 
```
    imageSearch = [CPImage imageWithCSSDictionary:@{
                                                    @"background-image": @"url(%%packed.png)",
                                                    @"background-position": @"-16px -32px",
                                                    @"background-repeat": @"no-repeat",
                                                    @"background-size": @"100px 400px"
                                                    }
                                             size:CGSizeMake(16,16)];
```

In this example, you see a typical usage of a CSS based CPImage : using an image contained in a single packed image (thus reducing image loading overhead). Please note the URL of the single image : ``url(%%packed.png)``. The ``%%`` will be replaced at theme loading by the path to the theme blend resources folder.

Of course, a CSS based CPImage can be anything else, like : 
```
    regularImageNormal = [CPImage imageWithCSSDictionary:@{
                                                           @"border-color": A3ColorActiveBorder,
                                                           @"border-style": @"solid",
                                                           @"border-width": @"1px",
                                                           @"border-radius": @"50%",
                                                           @"box-sizing": @"border-box",
                                                           @"background-color": A3ColorBackgroundWhite,
                                                           @"transition-duration": @"0.35s",
                                                           @"transition-property": @"all",
                                                           @"transition-timing-function": @"ease"
                                                           }
                                                    size:CGSizeMake(16,16)];
```
(one of the objectives of Aristo3 is to use as few images as possible)

To be considered and treated as CSS based, a theme must set to YES the attribute "css-based" for CPView in the themeDescriptor (default value is NO) : 
```
+ (CPView)themedView
{
    var view = [[CPView alloc] initWithFrame:CGRectMake(0, 0, 100, 100)];

    [self registerThemeValues:[[@"css-based", YES]] forView:view];

    return view;
}
```

You can use the method ``(BOOL)isCSSBased`` on a CPTheme, CPColor or CPImage object to determine how to cope with it in your own code.

Please note that other UI widgets must be adapted in order to optimize CSS based theme usage (for example, to reduce the number of used DIV's). Those modifications will be part of subsequent PR's.


daboe01:
-#new
+#needs-review
+Theme
milestone=1.0

didierkorthoudt:
Well, this is THE trick of themes : background colors are images doing shadows, borders, backgrounds and so on. CSS background just does the same...  ;-)

theredhead:
but a background _has_ a color, a color _does not have_ a background. Red _is_ a color but red has no margin or padding etc.

Does that make sense?

didierkorthoudt:
Have a look at Aristo2. You’ll see that, for example, window background also draws window border...  ;-)

theredhead:
well I would find that weird too. 

Note that trying to stat an argument, I'm not even trying saying anything remotely negative about your work here, I actually really like what i've seen so far and can't wait to play with it! I'm a big fan.

It's just that creepy feeling I get when I read some code that uses names that describe a well known thing but do, or are responsible for, something not in that well known definition.

 I'm the guy that has major headaches over the use of the word "right" where I believe it should say "privilege" in our companies codebase. Nobody else understands the issue and I can't do anything about it, but it is definitely wrong and it irks me every single time I see it. (one person cannot give or take away anybodies' rights, but one person can extend or revoke privileges, they are not synonyms).
I can live with it, but it requires a bit more brainpower.

mrcarlberg:
Yeah, ```CPColor``` has a weird concept. Original it can be a regular color but also an image. This means that you can paint anything with a ```CPColor```. I guess to mimic this with CSS we have to support all these things like margins, paddings and so on.

The alternative is to rewrite a large part of how colors are being used in Cappuccino. This would be a really big task. I think this is the best way to move forward.

mrcarlberg:
@didierkorthoudt I can't find that this pull request does anything 😉

mrcarlberg:
And that is a good thing! 😄 
I have tested it on our project with our own Theme.

theredhead:
@mrcarlberg that confirms my suspicion and means I'm not crazy, or at least not entirely 😀. Hope we can have Aristo 3 soon!

didierkorthoudt:
@mrcarlberg OK, I’ll do that ASAP. I agree we should redo all this from scratch in order to clean and optimize but this will be a disruptive move. And before, lots of things remain to do...  ;-)

didierkorthoudt:
@mrcarlberg OK, I’ll do that ASAP. I agree we should redo all this from scratch in order to clean and optimize but this will be a disruptive move. And before, lots of things remain to do...  ;-)

mrcarlberg:
Looks great! Just one keyboard slip to correct 😉

AppKit/CPTheme.j
```diff
@@ -319,6 +319,27 @@ var CPThemeNameKey          = @"CPThemeNameKey",
 #pragma mark -
 #pragma mark CSS Theming
 
+// The code below adds support for CSS theming with 100% compatibility with current theming system.
+// The idea is to extend CPColor and CPImage with CSS components and adapt low level UI components to
+// support this new kind of CPColor/CPImage. See CPImageView, CPView and _CPImageAndTextView.
+//
+// To be considered and treated as CSS based, a theme must set to YES the attrivute "css-based" for CPView
```
mrcarlberg: attrivute looks misspelled 😃

mrcarlberg:
Merged! Thanks!

PR #2619, opened 2018-04-03
Formatting: remove redeclared variables in CGPath.j

bedalton:


daboe01:
good catch!
+#ready-to-commit

PR #2626, opened 2018-04-05
fixed: backspace did not trigger scrolling properly 927d6d2

daboe01:
previously, deleting text did not keep the selection visible under some circumstances.
this PR fixes that with a call to scrollRangeToVisible:


daboe01:
+TextView
+#ready-to-commit

PR #2677, opened 2018-04-10
Animating frame origin was not working on recent FireFox versions

cacaodev:
Previously, the format for the matrix value of the transform property was matrix(a,b,c,d,tx,ty) where tx and ty had the "px" suffix. [According to Mozilla doc](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix) , tx and ty were accepted (in my understanding not required) with this suffix before FireFox 16 and required after.

All parameters are now numbers without suffix and this code is expected to work
with all FireFox versions supporting css animations.


daboe01:
looks good to me
-#new
+AppKit
+#ready-to-commit

mrcarlberg:
Merged! Thanks!

PR #2678, opened 2018-04-12
Fix: CPTabview placeholderView inconsistent variable name case

### LINKED ISSUE #2674: CPTabView missing setter: _setPlaceholderView
On line 706, a method is called on an variable `_source` with the selector `_setPlaceholderView`. 
https://github.com/cappuccino/cappuccino/blob/77eb6b358f38f24b60bce77bdffdc492aa412ffb/AppKit/CPTabView.j#L706

As far as I can tell, this method does not exist. The closest I could find was there is a property on CPTabView (the files namesake) with the name `_placeHolderView`
https://github.com/cappuccino/cappuccino/blob/77eb6b358f38f24b60bce77bdffdc492aa412ffb/AppKit/CPTabView.j#L70
Adding `@accessors` did not help as there is a case difference between `_setPlaceHolderView` from accessor and `_setPlaceholderView` which is the method called.
The variable types are all listed with `id` for the most part, so I cannot figure out what the actual variable types are.

bedalton:
In CPTabView.j there was a mismatch on the variable name -- `_placeHolderView` vs `_placeholderView`. This pull request fixes this issue by using the variable name `_placeholderView` in all instances.

Fixes #2674

### DIFF
--- a/AppKit/CPTabView.j
+++ b/AppKit/CPTabView.j
@@ -67,7 +67,7 @@ var CPTabViewDidSelectTabViewItemSelector           = 1 << 1,
 
     CPSegmentedControl      _tabs;
     _CPTabViewBox           _box;
-    CPView                  _placeHolderView;
+    CPView                  _placeholderView @accessors;
 
     CPTabViewItem           _selectedTabViewItem;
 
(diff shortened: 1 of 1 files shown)


cacaodev:
Merged, thanks a lot @bedalton !

PR #2681, opened 2018-04-28
FIXED: When attaching an externally owned tracking area to a view, view tracking areas were not correctly handled

didierkorthoudt:
When an externally owned tracking area is attached to a view, two problems were corrupting the view tracking areas handling : 
- CPControl and CPTextField were using -(void)removeAllTrackingAreas for their own needs but this was also removing the external tracking area, without any chance to reinstall it. This is fixed by avoiding the use of -(void)removeAllTrackingAreas and do a finer tracking area management ;
- If the externally owned tracking area was attached BEFORE putting the view in the view hierarchy, then the view had at least a tracking area (the externally owned) and had then no chance to install its own tracking areas as, in this case, -(void)updateTrackingAreas was not called when the view is put in the view hierarchy. This is fixed by directly calling [self updateTrackingAreas] at view initialization.

This PR also removes the CPView -removeAllTrackingAreas utility method as it is « dangerous » to use without very good knowledge of underlying view hierarchy tracking area management. This method :
- removes all tracking areas whatever the owner is ;
- doesn’t exist in Cocoa.


cacaodev:
Can you check if updateTrackingArea is called when a view is initialized from a cib. Tx.

didierkorthoudt:
@cacaodev Well, I don't know how I've missed this but 3 classes were always using removeAllTrackingAreas. The last commit updates method -updateTrackingAreas.

May I suggest to close PR #2707 and to remove -removeAllTrackingAreas method here so everything will be coherent for integration tests ?

cacaodev:
OK. Just to clarify: it is ok to have a PR failing tests in Travis because of a dependency on another PR. My concern with #2707 is that there are direct calls to -removeAllTrackingAreas in CPViewTest.

didierkorthoudt:
Ok then. I've closed #2707 and pushed the removal of -removeAllTrackingAreas here.

didierkorthoudt:
I've just ```jake test``` and found a unit test using removeAllTrackingAreas. Fixing it right now.

didierkorthoudt:
Removed the use of -removeAllTrackingAreas from CPViewTest.j unit test.

PR #2684, opened 2018-04-28
FIXED: Before this PR, CPSearchField was handling cursor updates by direct DOM manipulation. This PR uses tracking areas.

didierkorthoudt:
This PR needs PR #2681


AppKit/CPSearchField.j
```diff
@@ -773,6 +769,63 @@ var CPAutosavedRecentsChangedNotification = @"CPAutosavedRecentsChangedNotificat
 
 @end
 
+#pragma mark -
+
+@implementation CPSearchField (CPTrackingArea)
+{
+    CPTrackingArea      _searchButtonTrackingArea;
```
cacaodev: We do not declare instance variable in categories, neither does objective-c. I heard from @mrcarlberg that the compiler tolerates it but for consistency these should go in the  main class implementation.

didierkorthoudt: Well, in fact, I was considering this more like a class extension... (thus permitting declaration of properties and instance variables). This way, the code is really isolated and, but that's my own feeling, is easier to read / understand. Also, when adding new things (like use of tracking areas or CSS theming), the original code remains untouched.
But, OK, if it was a true class extension, I wouldn't have put "(CPTrackingArea)" (I've tried, of course found that we do not support class extension, and also considered that it was more clear to have a kind of "subject", like for a category).

Are you sure you prefer not to use this ?

didierkorthoudt: (I love to isolate / structure / ... the code as it is easier to understand / maintain ;-) )

cacaodev: I kind of agree with you, it would be a cool feature to have on objective-j. But we can't experiment it just in one place. It has to be tested, validated and documented. If you want, you could open a new issue to discuss that.

mrcarlberg: I think it is ok to use this feature. In Objective-C you can declare a property in categories. Objective-J does not support property declarations anywhere. We do support instance variable in a category and it is not supported in Objective-C as @cacaodev said above.

You could say that if you want a property in a category in Objective-J you have to declare it as an instance variable.

cacaodev:
All right. @didierkorthoudt can you please initialize these instance variable with their default value and then I think it will be ready to merge.

cacaodev:
Yes.

didierkorthoudt:
Done.

But are you sure ? The guidelines say ```In Objective-J, instance variables are initialized to nil automatically. Don't add explicit initializations to nil or NO in an init method.```

didierkorthoudt:
No problem, of course. I also had to change my habits when I've started some months ago after reading the guidelines (to be precise, after that Aparajita and Dogild told me to do so, as I was using my old way of writing code).

Should I revert the last commit ?

cacaodev:
As you wish.

didierkorthoudt:
Well, I've reverted so we don't introduce intentional guidelines bypass.

Thank you for your reviewing.

mrcarlberg:
Just for your information. It is true that all instance variables are initialized to ```nil```, but not ```NO``` if it is a boolean. This is because we inherit from Javascript and ```nil```and ```NO``` is not the same thing in this language.
In Objective-C ```nil``` and ```NO``` is the same thing.

didierkorthoudt:
Well, maybe should we adapt the guidelines to only talk about ```nil``` in order to avoid problems ?

PR #2685, opened 2018-04-28
FIXED: Before this PR, CPButton setImage was also changing the alternate image

didierkorthoudt:
inState:CPThemeStateNormal was missing so changing the image was done also for alternateImage.


cacaodev:
Merged, thanks !

PR #2686, opened 2018-04-28
FIXED: Before this PR, loading of a CPTabView from a CIB file could crash the app

didierkorthoudt:
When loading a CPTabView from a CIB file, when items views are decoded and inserted as tab view item, they are not completely initialized. The problem here is that they have no _scaleSize. When inserting them, if no tab is already selected, the first one is then selected, and a setFrameSize is called that crashes because of the lack of _scaleSize.

Moreover, this case seems to have been taken in account (see awakeFromCib) but not fully.

The fix consist on adding a flag to _insertTabViewItems method indicating if this selection can be done (which is YES all the time but not during initWithCoder).


Dogild:
Thanks for the fix. Im wondering, if half of the works is already done in awakeFromCib, what is the reason to not fix the issue there ?

mrcarlberg:
It is not good if a view is doing the last part of the initialization in ```awakeFromCib:```. The reason for this is that all views are expecting that all other views has done the initialization when ```awakeFromCib:``` is called.

It is much better to do the last part of the initialization in the method ```_cibInstantiate```. It is called after all the init methods are done for each view but before any of the ```awakeFromCib:```. The method ```_cibInstantiate``` can return a new view that will be replaced or has to return ```self```.

This is something that we need to fix in many different views classes as sometimes crashes will occur when loading a complex but good cib file.

With that said, I think this fix is good enough for now as we need to go back to the ```CPTabView```class and move the last initialization part from ```awakeFromCib:``` to ```_cibInstantiate```.

mrcarlberg:
It is not good if a view is doing the last part of the initialization in ```awakeFromCib:```. The reason for this is that all views are expecting that all other views has done the initialization when ```awakeFromCib:``` is called.

It is much better to do the last part of the initialization in the method ```_cibInstantiate```. It is called after all the init methods are done for each view but before any of the ```awakeFromCib:```. The method ```_cibInstantiate``` can return a new view that will be replaced or has to return ```self```.

This is something that we need to fix in many different views classes as sometimes crashes will occur when loading a complex but good cib file.

With that said, I think this fix is good enough as we need to go back to the

mrcarlberg:
Merged! Thanks!

PR #2688, opened 2018-04-28
FIXED: CPTextField forced height to 29 instead of using theme value min-size

didierkorthoudt:
In + (CPTextField)textFieldWithStringValue and + (CPTextField)roundedTextFieldWithStringValue methods, the height of the created textfield was forced to 29 instead of using theme specified value (« min-size »)


AppKit/CPTextField.j
```diff
@@ -169,7 +169,8 @@ CPTextFieldStatePlaceholder = CPThemeState("placeholder");
 
 + (CPTextField)textFieldWithStringValue:(CPString)aStringValue placeholder:(CPString)aPlaceholder width:(float)aWidth theme:(CPTheme)aTheme
 {
-    var textField = [[self alloc] initWithFrame:CGRectMake(0.0, 0.0, aWidth, 29.0)];
+    var minSize = [aTheme valueForAttributeWithName:@"min-size" forClass:CPTextField],
+        textField = [[self alloc] initWithFrame:CGRectMake(0.0, 0.0, aWidth, minSize)];
```
cacaodev: Should be minSize.height

AppKit/CPTextField.j
```diff
@@ -190,7 +191,8 @@ CPTextFieldStatePlaceholder = CPThemeState("placeholder");
 
 + (CPTextField)roundedTextFieldWithStringValue:(CPString)aStringValue placeholder:(CPString)aPlaceholder width:(float)aWidth theme:(CPTheme)aTheme
 {
-    var textField = [[CPTextField alloc] initWithFrame:CGRectMake(0.0, 0.0, aWidth, 29.0)];
+    var minSize = [aTheme valueForAttributeWithName:@"min-size" forClass:CPTextField],
+        textField = [[CPTextField alloc] initWithFrame:CGRectMake(0.0, 0.0, aWidth, minSize)];
```
cacaodev: idem

didierkorthoudt: Idem  ;-)

didierkorthoudt:
Strange : everything OK on my local clean Jake install but Travis reported 5 errors on CPScrollViewTest, CPKeyValueBindingSimpleBindingsTest, CPWindowTest, CPViewControllerTest. I don't understand what could be related to my code... Any idea ?

didierkorthoudt:
Sorry, spoke too quickly. Jake test fails also on my local install. Checking why...

didierkorthoudt:
Well, if I protect the use of minSize, that is : ```minSize ? minSize.height : 0```, everything is OK. But minSize should always exist as there's a default value in CPControl (CGSizeMakeZero()). What I suspect is that tests are done without theme support.

Do I commit this "protected" version ?

didierkorthoudt:
Well, I have committed the protected version, forgot to mention.

mrcarlberg:
@cacaodev @didierkorthoudt How can we move this PR forward?

didierkorthoudt:
milestone=1.1

mrcarlberg:
@didierkorthoudt I believe the problem here is that `aTheme` is `nil` when running the tests as no themes are loaded. This will give you `nil` as a value for `min-size`.

To get the default value for the theme value you have to ask an instance of `CPTextField`, not the theme.

I guess that best solution should be to test if `aTheme` is nil and if so create a zero size. It will make it very clear when a zero size is created and why. I think the current guard is to generic.

Maybe some thing like this: `var minSize = aTheme ? [aTheme valueForAttributeWithName:@"min-size" forClass:CPTextField] : CGSizeMake(0,0)`

didierkorthoudt:
I've pushed the better guardian suggested by @mrcarlberg and added a default value corresponding to the previous behavior (using 29.0 as height).

mrcarlberg:
@didierkorthoudt I think your merge with the master did duplicate the `NSUserNotificationAlertStyle`key in Tools/XcodeCapp/XcodeCapp/Info.plist

didierkorthoudt:
@mrcarlberg Indeed… Strange as I've resolved the conflict shown by GitHub and I think it was suggested like that… (or I've done it too quickly…). I'll revert my commit and check. Sorry for this…

mrcarlberg:
@didierkorthoudt Your check is ok just remove the two rows from the Info.plist in XcodeCapp

mrcarlberg:
Merged! Thanks!

PR #2689, opened 2018-05-01
FIXED: _CPMenuItemStandardView didn't take in account the left-margin theme attribute

didierkorthoudt:
_CPMenuItemStandardView didn't take in account the left-margin theme attribute to set position of the status mark.


cacaodev:
Can you add a test (possibly in an existing one) showing the effect of the left-margin attr on menu items with and without a state marker.Tx.

didierkorthoudt:
I'm getting mad with this test... For what I can see, actual layout is done only when the user opens the menu, not when creating the menu. So I can observe that something changes when I change the value of left-margin but positions are obviously wrong. When I check them after opening the menu, OK everything seems correct.

So, I'm searching how to force the menu to compute its actual layout but without success until now... (without, of course, using the mouse to open the menu, as we are on unit tests...)

Any idea ?

mrcarlberg:
@cacaodev @didierkorthoudt How can we move this PR forward?

didierkorthoudt:
milestone=1.1

Dogild:
Need to sync with master

mrcarlberg:
@didierkorthoudt I think your merge with the master did duplicate the NSUserNotificationAlertStylekey in Tools/XcodeCapp/XcodeCapp/Info.plist

mrcarlberg:
Merged! Thanks!

PR #2691, opened 2018-05-01
FIXED: CPSearchField was positionning its menu with hard coded offset

didierkorthoudt:
This PR introduces a new theme attribute (search-menu-offset) with hard coded values as default value.


cacaodev:
The question here is: when a new theme attribute is created, does it have to be included in theme descriptors ?
One argument in favor is that we do not have documentation for theme attributes. The only place where you can find them all if you want to change a default or create your own theme, is the  theme descriptor file.
@Dogild @mrcarlberg what do you think ?

didierkorthoudt:
What we can do to facilitate theme creation is to document the Aristo3 themeDescriptor.j and include, for each UI widget, the complete list of attributes, their use and their default value, even if not all are used for this particular theme. I think we should also specify if one attribute is inherited from superclass. What do you think ? It would be a long work but it could help.

AppKit/CPSearchField.j
```diff
@@ -73,7 +73,8 @@ var CPAutosavedRecentsChangedNotification = @"CPAutosavedRecentsChangedNotificat
             @"image-cancel": [CPNull null],
             @"image-cancel-pressed": [CPNull null],
             @"image-search-inset" : CGInsetMake(0, 0, 0, 5),
-            @"image-cancel-inset" : CGInsetMake(0, 5, 0, 0)
+            @"image-cancel-inset" : CGInsetMake(0, 5, 0, 0),
+            @"search-menu-offset": CGSizeMake(10, -4)
```
cacaodev: Can you check if the concept of "size offset" already exist in cappuccino or cocoa. If not, we should get the same result with an inset.

didierkorthoudt: Well, this may be my fault. I think this could be CGPointMake instead of CGSizeMake. I used CGSizeMake as I think of offset width (for horizontal offset) and offset height (for vertical offset) but this may be only in my head... What do you think ?

cacaodev: if it represents a distance between a location and some origin, a CGPoint seems more appropriate.

didierkorthoudt:
@cacaodev As the discussion is only related to attribute documentation, is it possible to merge this PR ? Thx

Dogild:
@didierkorthoudt can you rebase with master please. This is ready to be merged

Dogild:
@cacaodev I guess the theme descriptor is for a showcase and not for documentation purpose. Ideally,  the showcase could have widgets to tune other widgets in live. But it is far from that :)

mrcarlberg:
@didierkorthoudt Can you resolve the conflicts before I merge this one

didierkorthoudt:
@mrcarlberg Done. I hope I did it correctly (it's the first time I do this operation).

mrcarlberg:
Merged! Thanks!

PR #2694, opened 2018-05-03
FIXED: CPTabView tabs were not reacting as in modern Cocoa

didierkorthoudt:
in modern Cocoa, when you select a tab (mouse down, holding) and drag upon another tab, the second one reacts and can then be selected by releasing the mouse over it. 

This PR uses a subclass of CPSegmentedControl in order to adapt the behavior.

This PR needs PR #2687


cacaodev:
We should either:
1/ Use -trackSegment: subclassing but make the code more elegant and compact by calling super when possible. And include #2687 in the mouseUp event.
Or 2/ Put this PR under a private flag directly in CPSegmentedControl and use a private delegate also in CPSegmentedControl for -isSegmentEnabled: that would return -_sendDelegateShouldSelectTabViewItem: in CPTabView.

didierkorthoudt:
@cacaodev I think that the subclassing is the most structural solution as CPSegmentedControl is not designed to have the needed comportment and we have decided to use it for something else (tabs management). So it's coherent to subclass it.

Regarding calling super, I don't see any place in the only one method where to do such a thing. The behavior is so different that it wouldn't be elegant to try to mimic CPSegmentedControl original method just to call super somewhere. The suggested code is also more structured than the original one by the use of a switch statement instead of if / if / else if (where I would have used if / else if / else if). Everything else calls original class methods.

OK, I can include #2687 in this one (you asked me to separate the two issues) and close it.

Thank you for your review !..  

:-)

I need all those PR's to be merged to be able to continue towards Aristo3...

cacaodev:
OK let's do that. If you prefer the switch statement over if/else, then can you please change it also in CPSegmentedControl or if not, leave if/else in the subclassing (I have no preference). It makes it easier to compare the implementations and maintain on the long term. Then I will have some minor code style requests.

didierkorthoudt:
Closed PR #2687.
Refactored code of -trackSegment in CPSegmentedControl to match subclass here (PR #2704).
Ready.

AppKit/CPTabView.j
```diff
@@ -887,3 +893,78 @@ var CPTabViewItemsKey               = "CPTabViewItemsKey",
...
+
+            if (_trackingSegment === CPNotFound)
+                return;
+
+            if ((_trackingSegment !== _selectedSegment) &&
+                [_tabView _sendDelegateShouldSelectTabViewItem:[_tabView tabViewItemAtIndex:_trackingSegment]])
+            {
+                [self setSelected:YES forSegment:_trackingSegment];
+                _selectedSegment = _trackingSegment;
+                [self sendAction:[self action] to:[self target]];
```
cacaodev: You can select the tab directly here, no need to send the action, and of course remove the action setting in init.

AppKit/CPTabView.j
```diff
@@ -887,3 +893,78 @@ var CPTabViewItemsKey               = "CPTabViewItemsKey",
...
+            }
+
+            [self drawSegmentBezel:_trackingSegment highlight:NO];
+
+            _trackingSegment = CPNotFound;
+
+            return;
+            break;
```
cacaodev: No need for break if you return.

AppKit/CPTabView.j
```diff
@@ -887,3 +893,78 @@ var CPTabViewItemsKey               = "CPTabViewItemsKey",
...
+
+        case CPLeftMouseDown:
+
+            if (currentSegment > CPNotFound)
+            {
+                _trackingSegment = currentSegment;
+                [self drawSegmentBezel:_trackingSegment highlight:YES];
+            }
+
```
cacaodev: No space here as per our [guidelines](https://github.com/cappuccino/cappuccino/blob/master/CONTRIBUTING.md#cappuccino-coding-style-guidelines)

cacaodev: The previous comments also apply to #2704

AppKit/CPTabView.j
```diff
@@ -527,10 +531,7 @@ var CPTabViewDidSelectTabViewItemSelector           = 1 << 1,
 
 - (void)mouseDown:(CPEvent)anEvent
 {
-    var segmentIndex = [_tabs testSegment:[_tabs convertPoint:[anEvent locationInWindow] fromView:nil]];
-
-    if (segmentIndex != CPNotFound && [self selectTabViewItemAtIndex:segmentIndex])
-        [_tabs trackSegment:anEvent];
+    [_tabs trackSegment:anEvent];
```
cacaodev: Unless there is a reason I don't see, it seems right to leave this events filtering.

AppKit/CPTabView.j
```diff
@@ -887,3 +893,78 @@ var CPTabViewItemsKey               = "CPTabViewItemsKey",
...
+}
+
+- (void)trackSegment:(CPEvent)anEvent
+{
+    var type           = [anEvent type],
+        location       = [self convertPoint:[anEvent locationInWindow] fromView:nil],
+        currentSegment = [self testSegment:location];
+
+    switch (type) {
+
```
cacaodev: Open bracket on a new line, thanks.

AppKit/CPTabView.j
didierkorthoudt: Don't think I'm trying to argue on everything you say (that's really not the case) but guidelines say : ```There should be blank lines around bracketed code blocks and control structures```

;-)

AppKit/CPTabView.j
didierkorthoudt: OK.

AppKit/CPTabView.j
didierkorthoudt: Sorry for this. I always open it on a new line but Xcode putted it like that and I've certainly forgotten to change it.

AppKit/CPTabView.j
didierkorthoudt: OK.

AppKit/CPTabView.j
didierkorthoudt: Well, if we leave the -selectTabViewItemAtIndex call, it will indeed select the tab view at mouseDown whereas one goal of the PR is to select it at mouseUp.

cacaodev: No selection but you still need to filter out events that are not over a segment. Currently all mouse clicks on the content view are forwarded to the segmented control.

didierkorthoudt: Yes, indeed. I've considered that :

as the segmented control is called to test on which segment (or none) the click is done ;
as, if the click is done on a segment, the control segment will redo the same check on the trackSegment method ;
then it was OK to call trackSegment directly.
But you're right, this can be filtered. Moreover I would say this must be filtered because it introduces a disruption with Cocoa.

Thank you.

AppKit/CPTabView.j
```diff
@@ -527,9 +529,7 @@ var CPTabViewDidSelectTabViewItemSelector           = 1 << 1,
 
 - (void)mouseDown:(CPEvent)anEvent
 {
-    var segmentIndex = [_tabs testSegment:[_tabs convertPoint:[anEvent locationInWindow] fromView:nil]];
-
-    if (segmentIndex != CPNotFound && [self selectTabViewItemAtIndex:segmentIndex])
+    if ([_tabs testSegment:[_tabs convertPoint:[anEvent locationInWindow] fromView:nil]] !== CPNotFound)
         [_tabs trackSegment:anEvent];
 }
```
cacaodev: Just one last thing: I think the "right" way to intercept events is to use -hitTest:
You override this method and return the segmented control (_tabs) if the mouse is over a segment and call super otherwise.

didierkorthoudt: I see in the tabView init that we set hitTests to NO for the segmented control.
What do you think about removing this and letting the segmented control doing its stuff ?
I think this was done to filter on selectTabViewItemAtIndex but as we removed this...

didierkorthoudt: Just tested and it works...

PR #2698, opened 2018-05-04
FIXED: CPTableView was using hard coded values instead of theme attributes

didierkorthoudt:
This PR :

- adds a new theme attribute « header-view-height » instead of forcing the header view height to the row height ;
- uses the defined theme attributes « alternating-row-colors » and « grid-color » instead of hard coded values.


cacaodev:
Can you have a look at NSTableHeaderView.j. It seems it's not picking the new attribute for the header height.

didierkorthoudt:
You're right (one again...  ;-)

I didn't have a look in there. 

Done now. I'll open a new PR fixing this.

Thank you for watching !..

PR #2699, opened 2018-05-05
Fixed: CPSearchPanel did not support PredicateBindings

### LINKED ISSUE #2560: CPSearchField does not support the predicate binding
Tried to bind a CPSearchField to an ArrayController predicate, and got the following error:

> <CPSearchField 0x0130b6> is not key value coding-compliant for the key predicate

Same binding in native cocoa succeeds.

As a workaround, I'm delegating the CPSearchField and catching the "controlTextDidChange" message, where I can successfully tell the ArrayController to set the filter predicate, to the same value I was trying on IB.

daboe01:
previously, predicate bindings to CPSearchField threw an exeption
this PR introduces a specific binder class to support predicate bindings to CPSearchField.
fixes #2560


mrcarlberg:
Thanks! Merged!

cacaodev:
What is the problem with the manual test ? The text in the rows seems misplaced and I can't type in the search field after a row have been added.

cacaodev:
@mrcarlberg @daboe01 can you have a look at this ? tx.

daboe01:
@cacaodev @mrcarlberg  the text is also misplaced in Manual/TableTest/TableBindings. It looks like there is an issue in the current master that was introduced by a recent commit outside from this PR.

daboe01:
i just verified that 45d951071f80da082eb310d9576953a07ac53890 is good. the offending stuff was commited later.

daboe01:
the bad commit is 5bfafaf241a4789d4fe143efbbc657faddb0c0f8

PR #2701, opened 2018-05-06
NEW: Added dynamic attributes set for CSS based themes

didierkorthoudt:
If the default theme is CSS based and exposes some dynamic attributes (like Aristo3 does for color schema), you can switch "live" to another set of values by calling 
```[CPTheme updateDefaultThemeDynamicAttributesSetWithSet:aDynamicSet]``` 
where aDynamicSet is a dictionary with keys being the names of exposed dynamic attributes (see the themeDescriptor.j of the CSS based theme you use) and values, accordingly to the type of the corresponding default values provided by the theme (CPColor, a CSS color string, ...).

You can provide key-value pairs for a subset of the exposed dynamic attributes. If no key-value pair is specified for a dynamic attribute, the theme default value will be used.

To restore default theme values, simply pass nil as dynamic set : 
```[CPTheme updateDefaultThemeDynamicAttributesSetWithSet:nil]```

During theme loading, if your application responds to -(CPDictionary)themeDynamicAttributesSet, the returned values set is directly applied to the default set. This avoids displaying the theme with its default values and then redisplaying it with your values.

The -(CPDictionary)themeDynamicAttributesSet must be declared in a CPApplication category (and not in your application delegate) as it will be called during application initialisation, before the delegate has been defined.

 To use default theme values, simply return nil.

 TYPICAL USAGE:

 One can imagine that a user can choose between some predefined color schemas. You can store this user preference in standard user defaults:

      [[CPUserDefaults standardUserDefaults] setObject:ConstantRepresentingAColorSchema forKey:@"userTheme"];
      [[CPUserDefaults standardUserDefaults] synchronize];

 As you want to use the choosen color schema at run time, you've implemented -(CPDictionary)themeDynamicAttributesSet (see below), so updating
 the UI as soon as the user has made his choice is easy :

      [CPTheme updateDefaultThemeDynamicAttributesSetWithSet:[[CPApplication sharedApplication] themeDynamicAttributesSet]];

 If the user chooses the theme default color schema, you can then simply remove the user default:

      [[CPUserDefaults standardUserDefaults] removeObjectForKey:@"userTheme"];
      [[CPUserDefaults standardUserDefaults] synchronize];

 Your CPApplication category could then be something like:

      @implementation CPApplication (themeColorSchema)

      - (CPDictionary)themeDynamicAttributesSet
      {
          var userTheme = [[CPUserDefaults standardUserDefaults] objectForKey:@"userTheme"];

          if (!userTheme)

              // Return nil to have the theme default color schema
              return nil;

          switch (userTheme)
          {
              case ConstantRepresentingAColorSchema:
                  return @{
                           @"ExposedDynamicAttribute1":    [CPColor colorWithRed:245.0/255.0 green:164.0/255.0 blue:9.0/255.0 alpha:0.85],
                           @"ExposedDynamicAttribute2":    [CPColor colorWithRed:245.0/255.0 green:164.0/255.0 blue:9.0/255.0 alpha:0.85],
                           @"ExposedDynamicAttribute3":    @"rgb(245,164,9)",
                           @"ExposedDynamicAttribute4":    @"url(./Resources/mypacked.png)"
                          };
                  break;

              case AnotherConstantRepresentingAColorSchema:
                  return @{
                           @"ExposedDynamicAttribute1":    [CPColor colorWithRed:39.0/255.0 green:216.0/255.0 blue:93.0/255.0 alpha:0.85],
                           @"ExposedDynamicAttribute3":    @"rgb(39,216,93)"
                          };
                  break;

              ...
          }
      }

      @end

### CODE REVIEW

**Comment by didierkorthoudt** [comment]
milestone=1.1

**Comment by daboe01** [comment]
-#new
+Theme
+#ready-to-commit

**Comment by mrcarlberg** [comment]
@didierkorthoudt Needs to resolve the merge conflict on this one

File: Tools/XcodeCapp/XcodeCapp/Info.plist [hunk 1]
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>NSUserNotificationAlertStyle</key>
+	<string>alert</string>
```
mrcarlberg: @didierkorthoudt You two lines have sneaked in here too. 😄

didierkorthoudt: Strange… Should be in my own branch… I think I'll have to rebase it !.. OK, I adapt.

mrcarlberg:
Merged! Thanks!

PR #2705, opened 2018-05-08
FIXED: click to key window in the background did not bring it forward (take 2)

daboe01:
this is an improved version of the reverted #2602 that takes into account a somehow unexpected ordering of [CPApp orderedWindows]

previously, any click to a key window was simply forwarded to the clicked view without bringing the window to front. is is not the case in cocoa. here, the key window is always ordered front when clicked.

this PR introduces a check whether the keyWindow is the frontmost window and issues a makeKeyAndOrderFront unless the window is already in front.

fixes issue #2545


mrcarlberg:
Merged! Thanks!

PR #2706, opened 2018-05-11
FIXED: nib2cib was using hard coded frame adjustment for CPBox

didierkorthoudt:
This PR introduces, for views, the same nib2cib-adjustment-frame theme attribute that exists for controls, and uses this technic for CPBox.


didierkorthoudt:
@cacaodev Do you have something specifically in mind ?

cacaodev:
Why would this adjustment be theme specific ?
The final frame for the enclosing  view can be the same for all theme and then we can have insets, margins in themes.

didierkorthoudt:
Before this proposal, there was a hard coded frame adjustment for CPBox :  
```
        // small hack to position the box pixel perfect
        var frame = [self frame];

        // The primary and secondary boxes have a well-like look both in Cocoa and Cappuccino, but there
        // are some sizing differences. Regular custom boxes have no size differences.
        if (_boxType !== CPBoxSeparator && (_boxType === CPBoxPrimary || _boxType === CPBoxSecondary))
        {
            frame.origin.y += 4;
            frame.origin.x += 4;
            frame.size.width -= 8;
            frame.size.height -= 6;
        }
```

This was conflicting with trying to be pixel-accurate in Aristo3. So I decided to make this frame adjustment dynamic. Watching what is done for controls (that is theme specific adjustment), I thought that the best solution what to do the same for views. So this PR.

You can have theme specific adjustments. For example, Aristo2 is not pixel-accurate (generally in heights) with Cocoa, Aristo3 should be...

You've asked me for unit test. Do you think of something in particular ? I don't see any nib2cib unit test...

mrcarlberg:
@cacaodev How can we move this PR forward?

mrcarlberg:
@didierkorthoudt Can you please resolve the conflicts with the master branch

didierkorthoudt:
@mrcarlberg @cacaodev As discussed on Gitter, adding nib2cib adjustment on CPView (what is done via this PR) will permit to deal with precise positioning. I almost forgot that I've already proposed this here...  ;-)

didierkorthoudt:
@mrcarlberg The conflict is cleared. Can we go with this one ? To answer @cacaodev on why adjustment is theme specific, visual layout positioning depends on actual look of a widget and this actual look is specified by the theme. Also, it is already theme specific for controls. This PR is just an extension of this. I think that I forgot to simplify the code as a control is also a view. So we might remove adjustment code from control and use the one from view. Please wait while I check (and adapt if necessary).

didierkorthoudt:
Also found a small bug. Fixing.

didierkorthoudt:
@mrcarlberg Well, I've got a problem with this PR (and branch). I'll prepare a new one, then I'll close this one. Sorry.

didierkorthoudt:
OK, I close this one which is replaced by #2760 and #2761

PR #2712, opened 2018-05-19
FIXED: Before this PR, nib2cib wasn't using new theme attribute header-view-height

didierkorthoudt:


Tools/nib2cib/NSTableHeaderView.j
```diff
@@ -38,7 +38,7 @@
         if (_bounds.size.height === 17)
         {
             var theme = [Nib2Cib defaultTheme],
-                height = [theme valueForAttributeWithName:@"default-row-height" forClass:CPTableView];
+                height = [theme valueForAttributeWithName:@"header-view-height" forClass:CPTableView] || [theme valueForAttributeWithName:@"default-row-height" forClass:CPTableView];
```
cacaodev: The default value is [CPNull null]. Inclusive OR works with booleans or nil.

cacaodev: Can you check against CPNull instead of nil. Another solution would be to have a default float value of 25 like row-height.

cacaodev:
It seems there is also a wrong nil check in #2698 that have been merged 
involuntarily. Can you fix this in this pr. Tx.

didierkorthoudt:
Well, watching for this in #2698, I've found other problems with default theme values.
I'll push in a few minutes a specific PR.
Sorry for this, I think I've just duplicated another fix but it was not adapted here. Surely some too late in the night work.

didierkorthoudt:
Done. See #2732.

cacaodev:
This can be closed in favor of #2732, right ?

didierkorthoudt:
Well, I'm wondering if we don't have to completely remove this nib2cib treatment ?.. You're right, this seems to be redondant with #2732.

PR #2718, opened 2018-05-21
FIXED: Before this PR, legacy scroller won't ever loose its selected theme state

didierkorthoudt:
As no animation is required (and to avoid messing things), CPThemeStateSelected is directly unset.


cacaodev:
This early return now seems unnecessary:
https://github.com/didierkorthoudt/cappuccino/blob/93c6f0a98c8df3e5d24858fcbb81d85bbd4c02b1/AppKit/CPScroller.j#L463

didierkorthoudt:
Indeed. Removed.

PR #2719, opened 2018-05-23
FIXED: Before this PR, CPScrollView resets its scroller style to global scroller style

didierkorthoudt:
This is a disruption from Cocoa. When scroller style (overlay or legacy) is set, in Cocoa, CPScrollView keeps it, whatever happens, until another scroller style is specified.


didierkorthoudt:
Unit test CPScrollViewTest fails. I'll check this ASAP.

didierkorthoudt:
OK, the unit test is fixed.

The problem was that the ```[CPScrollView setGlobalScrollerStyle:CPScrollerStyleOverlay]``` was called after the creation of the scroll view and this PR indeed has an impact on this as now, when a scroll view has a specific scroller style, it keeps it unless setScrollerStyle is called.

I think this behavior is now coherent with Cocoa (even if setGlobalScrollerStyle doesn't exists in Cocoa).

Is this OK for you @cacaodev  ?

cacaodev:
This introduces a regression because -setGlobalScrollerStyle: cannot be used dynamically anymore.
In Cocoa, this method is private but exists (you can set globally the scroller style in the preferences panel) and I am pretty sure it overrides setScrollerStyle:. I didn't check but It seems natural to me that "global" overrides the instance method.
Is this related to Aristo3 ?

didierkorthoudt:
Well, just checked on Cocoa and, indeed, changing scroller style in System Preferences *after* launching the app and setting the scroller styles, changes them.

I'll propose then another fix that will keep this behavior.

didierkorthoudt:
After digging into why the unit test wasn't OK, I found that we have 2 distinct deviations from Cocoa, all related to how scroll views are notified of a global scroller style modification : when such a change is done (via [CPScrollView setGlobalScrollerStyle:XXX]), a notification is posted so that every existing scrollView can adapt the scroller style.

Those 2 deviations have been tested on a Cocoa application, playing with System Preferences (which is the Cocoa equivalent of [CPScrollView setGlobalScrollerStyle:XXX]).


First problem (the one I've fixed by the previous commit) :

0. Have a global scroller style, say legacy
1. Create a new scrollView (it then gets the global scroller style, which is normal)
2. Use -setScrollerStyle on this scrollView to change to, say, overlay
3. Insert the scrollView in the view hierarchy

Result : the scrollView gets back to the global scroller style, say, legacy.

This is a deviation from Cocoa.

Explanation of the origin of the problem :

When we insert a scrollView in the view hierarchy, CPView calls (in -viewWillMoveToSuperview) _addObservers. CPScrollView overrides this method to add notifications listening (which is OK). But it also calls _didReceiveDefaultStyleChange:nil to be sure to get the last global scroller style. This is some kind of fix for the second problem below.

But this introduces a side effect : it override the -setScrollStyle of step 2.

Fix : to fix this, we have to remove the call to _didReceiveDefaultStyleChange.

Side effect : this reintroduces problem two.



Second problem (reintroduced by fix of first problem) :

0. Have a global scroller style, say legacy
1. Create a new scrollView (it then gets the global scroller style, which is normal)
2. Use [CPScrollView setGlobalScrollerStyle:CPScrollerStyleOverlay]
3. Insert the scrollView in the view hierarchy

Result : the scrollView missed the global change of scroller style and keeps then, say, legacy

This is a deviation from Cocoa.

Explanation of the origin of the problem :

As mentioned in the first problem, notifications listening is added during execution of _addObservers, which is called in -viewWillMoveToSuperview, thus when adding the scrollView in the view hierarchy.

Thus, when a setGlobalScrollerStyle is called BEFORE the scrollView is in the view hierarchy, the scrollView doesn't get it (as it is not yet listening to notifications).

The fix that was (badly) done is to force the scrollView to get the global scroller style in _addObservers, so when it gets inserted in the view hierarchy. But as this has side effect (see first problem), we have to remove this fix.

As solution could be to add notifications listening at scrollView initialization. But as we don't implement dealloc mechanism, we would have no chance to remove this listening, and this is bad.

Fix : to fix this, we have remove the use of notifications on setGlobalScrollerStyle and implement another way of change all scrollViews scroller type when setGlobalScrollerStyle is called.

What I suggest (and tested) is that CPScrollView class maintains an array of all created scrollViews and in setGlobalScrollerStyle, to call -setScrollerStyle on each of them.

Even if a scrollView is release, this would be OK as we can call -setScrollerStyle on a nil object.

BTW, we can't remove the scrollView from the array when it is removed from the view hierarchy as it could be further reinserted.

Note : implementing dealloc would be a good solution but as we are on JavaScript, we can't.



What is you opinion on this ?

The 2 fixes are ready to commit and I've added unit tests to verify the good behavior.

cacaodev:
You can just init the _scrollerType instance variable with a default value meaning "no style have been set explicitly". Then the computed property -scrollerType would return the global style only if this  instance variable is the default (works before notifs are registered). Of course all direct ivar access need to be replaced with the previous computed property.
Something like this : 3673b06

didierkorthoudt:
Well, I've investigated this kind of approach before the proposed one and I think I've found a running condition that ended my research in this way. It was something like : 

0. Say global is overlay
1. Create the scrollview. _scrollType is then -1 
2. Set the type to overlay (silly but why not). _scrollType is no more -1
3. Change global to legacy
4. Insert in the view hierarchy.

As _scrollType is not -1, it will not get the global legacy.

Also : 

0. Global legacy
1. Create (_scrollType -1)
2. Put in view hierarchy (then, listening)
3. Global change to overlay (_scrollType is then overlay as listening is active)
4. Remove from view hierarchy (then, stop listening)
5. Global change to legacy (missed as not listening anymore)
6. Put back in view hierarchy 

As _scrollType is no more -1 (but overlay), it will not be adapted to last global change (legacy).

The problem is really how listening is activated and removed.

That's why I've searched something else...

I agree this is not the ideal way of doing but, as said before, we don't have any chance to use end of life of an object...

didierkorthoudt:
@cacaodev What do you think ?

cacaodev:
I think it is hard to tell without knowing the magnitude of this bug. Is it just a user issue where someone creates a scroll view and decides suddenly to change the global scroll style before inserting into the hierarchy as you describe in your previous comment ? In this case, I wouldn't worry too much, a warning saying "don't do that" should be enough.
If this is a problem that arise in the framework when developping, then I will need more context. If this is confirmed, one option is the one you propose, the other one is just to start observing at init time. In the latter case, the risk is that a scroll view is created and never added to the window (and never removed, observers stay active). We should evaluate this risk but it seems quite low.

didierkorthoudt:
Well, those deviations from Cocoa were mainly discovered when trying to add unit tests.

It's hard to evaluate the risk... One can imagine that you create a scrollview that will be attached to a tab view. When the corresponding tab is selected, your scrollview is inserted in the view hierarchy. When another tab is selected, your scrollview is removed from the view hierarchy. Then the user could select a new global scroller style and when he'll select again the scrollview tab, the global style won't be applied (as the scrollview is no more listening because out of the view hierarchy, while still living).

The main point is not that this is a crucial risk, but rather that we've found a deviation and have a solution to fix it.

(and leaving pending listeners seems bad to me  ;-)

didierkorthoudt:
@cacaodev @mrcarlberg I've resolved conflict in the unit test. Can we go with this one ?

mrcarlberg:
I don't think this is a big issue as I guess one will seldom change scroller style in real projects. I might be wrong on this.

But I think we should leave this out of version 1.0.

We will take a look at this after the release of version 1.0

didierkorthoudt:
@mrcarlberg OK for me.

didierkorthoudt:
milestone=1.1

daboe01:
-#new
+#ready-to-commit
+Theme

mrcarlberg:
@cacaodev What do you think? How should we proceed with this one?

cacaodev:
yes

mrcarlberg:
Do these change create a memory leak when an application is removing and adding new scroll views during its lifetime?
This is not a good solution as old scroll views needs to be removed from the global list when removed. Can it be done?

daboe01:
moved to #3038
+#duplicate

PR #2723, opened 2018-05-27
fixed: wording in index.html

daboe01:
trivial rewording in index.html


mrcarlberg:
I think this is a good change. It is more of an Application compared to Site

PR #2724, opened 2018-05-30
fixed: CPDateFormatter was agnostic for the German language

daboe01:
previously, relative German terms were missing and the separator was not the point as is the case in Germany. this PR fixes all this when the CPLocaleLanguageCode is "de"


daboe01:
+Foundation
+i18n
+#needs-review

mrcarlberg:
Merged! Thanks!

PR #2727, opened 2018-06-02
Refactor: Clean up build script for capp_lint

rramsden:
This shell script looked out of place to me, we already have convenience functions defined in `common.jake` (copying man pages, etc.) so we don't need to shell out to a bash script I think.

I'm also wondering if `/usr/local/narwhal/bin` needs to be hardcoded as well. In another tool `SYSTEM.prefix` is used to copy an executable:
https://github.com/cappuccino/cappuccino/blob/master/Tools/capp_env/Jakefile#L8


mrcarlberg:
Thanks for this clean up.

mrcarlberg:
This will install the tool when building. We don't want to do that. Installation should be done when installing not building.

mrcarlberg:
And yes, we do this in other places when building but we don't want to add more of this

rramsden:
@mrcarlberg Thanks for the feedback, mind offering some guidance on this one?

Would it be okay to add an "install" subtask to the Tool Jakefile here:
https://github.com/cappuccino/cappuccino/blob/master/Tools/Jakefile#L5

Then rename "build" to "install" here since there is no build step required for `capp_lint`. Afterwards, I can open other PR's to clean up and separate build/install steps from other tools.

mrcarlberg:
Sorry for not getting back to you on this.

Yes, I think this is the correct way of doing it.

mrcarlberg:
@rramsden How can we proceed with this pull request ?

rramsden:
@mrcarlberg Sorry about the delay. Will close for now until I have the time to submit a proper pull request. I did try my proposed solution ー renaming "build" to "install" but this conflicted with other Jake build tasks in common.jake. I'll try a second attempt when I have the time 🙇

PR #2733, opened 2018-06-22
Fixed:  native touch scroll event was not cancelled when handling touchscrolling

daboe01:
previously, the browser tried to scroll the whole page under some circumstances when touch scrolling in cappuccino
this is fixed by cancelling the native event before emitting the synthetic cappuccino scroll event.


daboe01:
i am closing this as it turns out, this PR is not fixing the issue that scroll events on scrollviews are additionally scrolling the body.
the best solution to the issue is to add: style="position: fixed" to the body definition in your index.html

PR #2740, opened 2018-07-09
FIXED: CPScroller superfluous setFrame: calls

cacaodev:
Before this commit, CPScrollerView was not correctly differentiating the _scroller-width_ attribute value between the vertical scroller and the horizontal scroller.
The scroller width was picked from a common CPScroller instance cached in CPScroller and was incorrect for one of the two directions.

The bug was not visible because the method `-_adjustScrollerSize` is correcting the size at each  layout pass. However, this was causing unnecessary calls to `setFrame:` in a performance sensitive method triggered by scroll events.

The ScrollView is now picking the right width from the theme and the `setFrame:` call is skipped.

TESTS: Manual/CPScrollView2/
TODO : The `-_adjustScrollerSize` method should be enabled with a flag only when a width change is explicitly requested.

### DIFF
--- a/AppKit/CPScrollView.j
+++ b/AppKit/CPScrollView.j
@@ -168,14 +168,13 @@ var CPScrollerStyleGlobal                       = CPScrollerStyleOverlay,
 
 + (CGSize)contentSizeForFrameSize:(CGSize)frameSize hasHorizontalScroller:(BOOL)hFlag hasVerticalScroller:(BOOL)vFlag borderType:(CPBorderType)borderType
 {
-    var bounds = [self _insetBounds:CGRectMake(0.0, 0.0, frameSize.width, frameSize.height) borderType:borderType],
-        scrollerWidth = [CPScroller scrollerWidth];
+    var bounds = [self _insetBounds:CGRectMake(0.0, 0.0, frameSize.width, frameSize.height) borderType:borderType];
 
     if (hFlag)
-        bounds.size.height -= scrollerWidth;
+        bounds.size.height -= [_horizontalScroller scrollerWidth];
 
     if (vFlag)
-        bounds.size.width -= scrollerWidth;
+        bounds.size.width -= [_verticalScroller scrollerWidth];
 
     return bounds.size;
 }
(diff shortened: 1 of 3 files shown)


mrcarlberg:
Merged! Thanks!

AppKit/CPScrollView.j
```diff
@@ -168,14 +168,13 @@ var CPScrollerStyleGlobal                       = CPScrollerStyleOverlay,
 
 + (CGSize)contentSizeForFrameSize:(CGSize)frameSize hasHorizontalScroller:(BOOL)hFlag hasVerticalScroller:(BOOL)vFlag borderType:(CPBorderType)borderType
 {
-    var bounds = [self _insetBounds:CGRectMake(0.0, 0.0, frameSize.width, frameSize.height) borderType:borderType],
-        scrollerWidth = [CPScroller scrollerWidth];
+    var bounds = [self _insetBounds:CGRectMake(0.0, 0.0, frameSize.width, frameSize.height) borderType:borderType];
 
     if (hFlag)
-        bounds.size.height -= scrollerWidth;
+        bounds.size.height -= [_horizontalScroller scrollerWidth];
```
mrcarlberg: @cacaodev I think I did a misstake when reviewing this as I never saw that `_horizontalScroller` is a instance variable that is not available in this class method

AppKit/CPScrollView.j
```diff
@@ -168,14 +168,13 @@ var CPScrollerStyleGlobal                       = CPScrollerStyleOverlay,
...
 
     if (vFlag)
-        bounds.size.width -= scrollerWidth;
+        bounds.size.width -= [_verticalScroller scrollerWidth];
```
mrcarlberg: Same thing here on `_verticalScroller`

AppKit/CPScrollView.j
```diff
@@ -185,14 +184,13 @@ var CPScrollerStyleGlobal                       = CPScrollerStyleOverlay,
     var bounds = [self _insetBounds:CGRectMake(0.0, 0.0, contentSize.width, contentSize.height) borderType:borderType],
         widthInset = contentSize.width - bounds.size.width,
         heightInset = contentSize.height - bounds.size.height,
-        frameSize = CGSizeMake(contentSize.width + widthInset, contentSize.height + heightInset),
-        scrollerWidth = [CPScroller scrollerWidth];
+        frameSize = CGSizeMake(contentSize.width + widthInset, contentSize.height + heightInset);
 
     if (hFlag)
-        frameSize.height += scrollerWidth;
+        frameSize.height += [_horizontalScroller scrollerWidth];
```
mrcarlberg: And here

AppKit/CPScrollView.j
```diff
@@ -185,14 +184,13 @@ var CPScrollerStyleGlobal                       = CPScrollerStyleOverlay,
...
 
     if (vFlag)
-        frameSize.width += scrollerWidth;
+        frameSize.width += [_verticalScroller scrollerWidth];
```
mrcarlberg: and here

PR #2743, opened 2018-07-29
Fixed: _CPObjectAnimator initialize method did not match superclass return

bedalton:
_CPObject animator labeled return type as BOOL, not only is this value not returned, but the super class CPProxy's initialize method returns void. Changed return type to void in _CPObject animator, to match superclass


cacaodev:
merged, thanks !

PR #2744, opened 2018-08-08
FIXED: _CPThemeAttribute valueForState was, under some circumstances, ignoring theme settings

didierkorthoudt:
**Problem:**
Using InterfaceBuilder, put a cell based table view. The resulting CPTextField that will be used to populate the table, will have a text-color attribute set for _tableDataView_ theme state, say to black (this example is simplified for clarity). If your theme specifies that text-color should be red if the cell is selected (that is for theme state _tableDataView+selectedTableDataView_), this setting will be ignored.

**Explanation:**
The _CPThemeAttribute valueForState method works as follow : 
1. Search in the cached values for a direct state match. If found, return the associated value.
2. Search in the values for a direct state match. If found, return the associated value.
3. Search in the values for a partial state match (the longest one). If found, return the associated value.
4. Try to use the "normal" state value. If found, return it.
5. Ask the _themeDefautAttribute. If value returned, return it.
6. Return the _defaultValue.

The problem here is that, for state _tableDataView+selectedTableDataView_, we have this : 
1. Nothing found (if first call).
2. Nothing found (as the text-color attribute is set for theme state _tableDataView_).
3. A partial match is found (_tableDataView_ is a subset of _tableDataView+selectedTableDataView_) so this value (black) is returned, instead of using the _tableDataView+selectedTableDataView_ value specified in the theme (red).

**Solution:**
In step 3 : 
- search for a partial state match in attribute values and keep match length
- ask _themeDefautAttribute for the same thing (value and length)

If the length of the match found by the _themeDefaultAttribute is longer than the match found by the attribute itself, use the theme match.

A small other problem is also fixed by this PR : in previous implementation, the step 4 (use of the "normal" state value) was only done if the state is composite. This fix uses step 4 whatever the state length.


Dogild:
Looks good to me. Good job @didierkorthoudt

mrcarlberg:
Merged! Thanks!

PR #2747, opened 2018-08-11
Fixed: added the method _isCharacterASeparator. This was added for re…

Dogild:
…factoring purpose


cacaodev:
Thanks @Dogild for this but what do you think of my approach: create a static character set containing separators and then test for inclusion against it ?

Dogild:
You are right!

cacaodev:
Closing in favor of #2762

PR #2750, opened 2018-08-16
fixed: touch scrolling events to scrollviews also scrolled the document body

daboe01:
modern mobile browsers forbid to cancel touch events via preventDefault() by default.
this caused touch scrolling events to scrollviews to also scroll the document body.
this is fixed by this PR by means of providing the passive: false option.
i tested this on an ipad pro with both mobile safari and mobile firefox.
i also tested this on macbook pro chrome / firefox -> no issues


daboe01:
i tested with the TableBindings manual test.

mrcarlberg:
Merged! Thanks!

PR #2753, opened 2018-08-17
FIXED: Before this PR, starting a column drag within a scrolled down table would scroll the table content to the bottom

didierkorthoudt:
Say you have a table with 50 rows. Scroll your table until you see row 30 on top of the visible rect like this :

bug1

Then start dragging a column left or right.

Your table will automatically scroll down until reaching the last row like this : 

bug2

The problem is in the way used to constrain the drag of the column to avoid autoscroll vertically.
The bug here is in the method `[view convertPoint:aPoint toView:aView]` : `aPoint` should be in the coordinate system of view and it is not.

This PR fixes this by always setting the vertical coordinate to the height of the header. This way, mouse is considered at the very top of the visible rect and starts then no autoscroll.

### DIFF
--- a/AppKit/CPTableHeaderView.j
+++ b/AppKit/CPTableHeaderView.j
@@ -550,7 +550,7 @@ var CPTableHeaderViewResizeZone = 3.0,
 - (void)_autoscroll:(CPEvent)theEvent localLocation:(CGPoint)theLocation
 {
     // Constrain the y coordinate so we don't autoscroll vertically
-    var constrainedLocation = CGPointMake(theLocation.x, CGRectGetMinY([_tableView visibleRect])),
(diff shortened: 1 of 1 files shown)


cacaodev:
The bug here is in the method [view convertPoint:aPoint toView:aView];
aPoint should be in the coordinate system of view and it is not.
Your fix achieve the same result. You can you just use the CGRectGetMaxY function for clarity and update the commit message about the origin of the problem.

didierkorthoudt:
@mrcarlberg Well, I've asked myself the same question after @cacaodev suggestion and checked if it works with simply MaxY and yes. There's a difference of 1 pixel and this is too small to initiate a vertical autoscroll. So I think we can use it this way. BTW, while doing this check, I've found another old bug of CPTableView that I'll fix in a new PR quickly (possibly today).

PR #2754, opened 2018-08-17
FIXED: Before this PR, when dragging a partially masked column, the dragged view was truncated

didierkorthoudt:
Say you have a CPTableView with 3 columns, the rightmost one being partially masked like this : 

bug1

When you drag this column, the dragged view (or ghost view) is clipped : 

bug2

This PR fixes this by using columnRect width instead of columnVisRect width.


mrcarlberg:
Merged! Thanks!

PR #2755, opened 2018-08-17
FIXED: Before this PR, when dragging a selected column, selection color remained in place (visual artefact)

didierkorthoudt:
Say you have a table with a selected column, like : 

bug1

If you start dragging this column, the selection color remains on the freed space : 

bug2

This is because, when creating the dragged view (or ghost view), the "will be dragged" column was indeed unselected but without any refresh of the display.

This PR fixes this by simply adding a setNeedsDisplay:YES.

### DIFF
--- a/AppKit/CPTableView.j
+++ b/AppKit/CPTableView.j
@@ -3189,6 +3189,7 @@ Your delegate can implement this method to avoid subclassing the tableview to ad
 
     // While dragging, the column is deselected in the table view
     [_selectedColumnIndexes removeIndex:columnIndex];
+    [self setNeedsDisplay:YES];
 
     return dragView;
 }


cacaodev:
Since this method is supposed to just construct the drag view, it would be better to have the display call in the parent method CPTableHeaderView- (void)_startDraggingTableColumn:(CPInteger)aColumnIndex at:(CGPoint)aPoint.

didierkorthoudt:
Well, strictly considering OOP, CPTableHeaderView asks CPTableView to construct the drag view, so it's CPTableView responsibility to take all necessary actions to provide the drag view. Also, CPTableView knows what to do and if / when it has to be refreshed. The drag view could be constructed in another completely different way not needing the CPTableView to be refreshed. So code isolation / abstraction also conducts to leave the setNeedsDisplay in the construction method.

Do you agree with that @cacaodev  ?

cacaodev:
I totally agree.

didierkorthoudt:
OK, so I leave like this ?

mrcarlberg:
Merged! Thanks!

PR #2756, opened 2018-08-17
FIXED: Before this PR, when an unfocused selected column was dragged, focused selection color was used

didierkorthoudt:
Unfocused selected column should draw in unfocused selection color (gray). Before this, when dragging the column, it was using the focused selection color (blue), then returned to gray when drag ended.

This was due to an unconditional use of selectionHighlightColor.

This PR fixes this by conditionally using selectionHightlightColor or unfocusedSelectionHighlightColor.


mrcarlberg:
Merged! Thanks!

PR #2757, opened 2018-08-17
NEW: Add first responder management based on content view to CPScrollView

didierkorthoudt:
This will (un)set CPThemeStateFirstResponder on CPScrollView based on the content view.

For now, only CPTableView's can initiate first responder state on a CPScrollView (and only if there's no in-cell editing).

This should be adapted when global focus ring management will be added to Cappuccino.

With this PR, it's easy to add a focus ring around a CPScrollView containing a CPTableView which is the first responder of the window (this will be used by Aristo3).

### DIFF
--- a/AppKit/CPScrollView.j
+++ b/AppKit/CPScrollView.j
@@ -1510,6 +1510,51 @@ Notifies the delegate when the scroll view has finished scrolling.
 
 @end
 
+#pragma mark -
+
+@implementation CPScrollView (FirstResponder)
+
+// Those 4 next methods are needed to (un)set CPThemeStateFirstResponder based on content view
+
+- (void)viewWillMoveToWindow:(CPWindow)aWindow
+{
+    [super viewWillMoveToWindow:aWindow];
+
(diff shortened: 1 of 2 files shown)


mrcarlberg:
@didierkorthoudt Can you please look into why the test case will fail.

CPScrollViewTest.......
addFailure test=[CPScrollViewTest testNotificationsRegistered]: Notications registered for the scrollView in the notification center are wrong expected:<@[
    @"CPScrollerStyleGlobalChangeNotification",
    @"_CPWindowDidChangeFirstResponderNotification"
]> but was:<@[
    @"CPScrollerStyleGlobalChangeNotification"
]>

didierkorthoudt:
@mrcarlberg Well, I've added a new notification listening in order to implement this new functionality. So I guess that the unit test is not adapted. I check immediately. Sorry for this.

didierkorthoudt:
@mrcarlberg Done.

mrcarlberg:
Merged! Thanks!

PR #2758, opened 2018-08-17
FIXED: Before this PR, Nib2Cib wasn't using theme attribute "header-view-height"

didierkorthoudt:
Nib2Cib was forcing the header view height to the default-row-height attribute.

With this PR, if a header-view-height attribute value exists, Nib2Cib uses it. If not, Nib2Cib uses default-row-height attribute value, as before.


cacaodev:
The `header-view-height` is now defined in CPTableView, meaning that it will always exist. it should be ok to just replace `default-row-height` with `header-view-height`

didierkorthoudt:
@cacaodev You're right. Done.

mrcarlberg:
Merged! Thanks!

PR #2759, opened 2018-08-18
FIXED: CPTrackingArea manual test was using removed method -removeAllTrackingAreas

didierkorthoudt:
There's no more -removeAllTrackingAreas.


mrcarlberg:
Merged! Thanks!

PR #2761, opened 2018-08-21
FIXED: Before this PR, nib2cib was using hard coded frame adjustment for CPBox

didierkorthoudt:
This PR uses #2831 to use a theme specific adjustment frame attribute instead of hard coded value. Default value corresponds to old hard coded value.


didierkorthoudt:
#milestone=1.1

didierkorthoudt:
#milestone=1.1

daboe01:
-#new
+Theme
+#ready-to-commit

AppKit/CPBox.j
```diff
@@ -90,7 +90,8 @@ CPBelowBottom = 6;
             @"inner-shadow-size": 6.0,
             @"inner-shadow-color": [CPNull null],
             @"content-margin": CGSizeMakeZero(),
-        };
+            @"nib2cib-adjustment-primary-frame": CGRectMake(4, -4, -8, -6)
```
Dogild: Why this is not called `nib2cib-adjustment-frame` as it is for other controls ?

mrcarlberg:
I have two issues on this:
1. An answer on Dogild's question above
2. Are the default values correct? Looks like `y` is -4 and the removed code adds 4?

AppKit/CPBox.j
didierkorthoudt: Well, in fact I wanted to show that this adjustment will only be applied on primary (and secondary) boxes (and not on all types of boxes)... This way, other types may use nib2cib-adjustment-frame the usual way…

didierkorthoudt:
Regarding the +-4 question, I have to check but I think this is because when using nib2cib-adjustment-frame, vertical coordinates are not yet reversed.

Tools/nib2cib/NSBox.j
```diff
@@ -80,6 +66,33 @@
     return self;
 }
 
+// Override
+
+- (CGRect)_nib2CibAdjustment
+{
+    if ((_boxType !== CPBoxSeparator) && ((_boxType === CPBoxPrimary) || (_boxType === CPBoxSecondary)))
```
mrcarlberg: @didierkorthoudt One last thing..... I hope 😄.... Why does this `if` first test `_boxType !== CPBoxSeparator` and after that `(_boxType === CPBoxPrimary) || (_boxType === CPBoxSecondary)`. Is the first test really needed or am I missing something here?

didierkorthoudt: You're absolutely right, I've just copied the old test and should have seen this… I'll adapt, thank you !

didierkorthoudt:
I've simplified the test. Are we OK on this one ?

mrcarlberg:
Merged! Thanks!

PR #2763, opened 2018-08-21
FIXED: Problem with nib2cib introduced by #2758

didierkorthoudt:
header-view-height theme attribute is defined at the table view level. When trying to use this value in NSTableHeaderView, _tableView is not completely initialized. This caused the problem with #2758.

The solution provided in this PR is to let the table view set the header height, when initialized.


didierkorthoudt:
Done but online as I'm not in front of my dev Mac. But I think that the risk of a side effect is very low...  ;-)

mrcarlberg:
Merged! Thanks!

PR #2764, opened 2018-08-22
Define header-view-height attribute in ThemeDescriptors

cacaodev:


mrcarlberg:
@didierkorthoudt  This looks like a much better solution?

didierkorthoudt:
@mrcarlberg Well, indeed it's simpler to force the presence of the header-view-height in Aristo2 theme. I thought about this solution after discovering the bug. But, and that's why I've changed my mind, what if someone has a custom theme ? He'll have to adapt it. What I wanted is a transparent fix. Never mind the presence or not of the attribute in the theme, it will work...

mrcarlberg:
@cacaodev I believe your fix for this will break any custom theme? Your solution is cleaner and maybe should be the final version.

What do you think of leaving this for after version 1.0 so custom themes will not break?

cacaodev:
If retro-compatibilty with previous custom themes is an issue, then replacing 
`[theme valueForAttributeWithName:@"header-view-height" forClass:CPTableView];` with 
`[_tableView  currentValueForThemeAttribute:@"header-view-height"];` seems enough unless there is something I am missing. 
After jake install and nib2cib, I did not see any issue with that. (tested also without the attribute in themeDescriptors).

mrcarlberg:
Merged! Thanks!

didierkorthoudt:
I've installed latest master and tried to use nib2cib. I've then got an error message saying that CPTableView doesn't have the attribute "header-view-height".
I've check the source file and found that it does indeed have one.

I remember that, during preparing my own fix, I've found that in NSTableHeaderView, _tableView is not completely initialized (I think that theme attributes are not loaded yet). That's why I've decided to report the fix in CPTableView, as proposed.

But this fix does try to use table view attributes.

cacaodev:
I reverted this because it is causing an issue in some xib files when nib2cibed.
We will have to find out why some attributes are available and not others in nib2cib because this may be a serious problem.

Issue #2708, opened 2018-05-14
Multiline text in CPTableView


jasperdeb:
I have a CPTableView, where I have a column defined as:

tableColumn = [[CPTableColumn alloc] initWithIdentifier:@"text"];
[[tableColumn headerView] setStringValue:@"Clausules in contract paragraaf:"];
[tableColumn setWidth:1300];
[tableColumn setEditable:YES];
[tableViewContractContent addTableColumn:tableColumn];

When I try to load the following multiline text

![image](https://user-images.githubusercontent.com/20356278/40004961-31e594e4-5797-11e8-8567-bb789b2412f2.png)

into the column, I have not the result I was hoping for with the code base 0.9.10 and master. Nevertheless, version 0.9.9 does give me what I expected.

![image](https://user-images.githubusercontent.com/20356278/40004791-bce9ecc6-5796-11e8-969c-e0d4be779134.png)

Am I doing something wrong and have API's changed, or is this a bug?

daboe01:
@jasperdeb try ```ObjJ
[[myTable dataView] setLineBreakMode:CPLineBreakByWordWrapping];
```

**Comment by jasperdeb** [comment]
@daboe01 Thank you! In combination with the vertical alignment set to CPTopVerticalTextAlignment, this works great if the cell is not edited. 

When the cell has the focus during the edit, the word wrapping does not occur.

![image](https://user-images.githubusercontent.com/20356278/40042737-a1ef81ec-5822-11e8-9e48-79288d46df4e.png)
![image](https://user-images.githubusercontent.com/20356278/40042780-c90244a4-5822-11e8-9d79-a39125d6f3c9.png)

How can I make sure the wrapping also occurs during edit?

**Comment by cacaodev** [comment]
Can you provide more info about the context ? 
Is it a cell-based or a view-based table view, is the table view also created in code, and if cell-based what is happening in the edit delegate. A full example would be very useful.

**Comment by jasperdeb** [comment]
Actually the code is quite simple.
The CPTableView is manually created (therefore not in a cib-file).
Definition of the table column in the code is as follows:

```
tableView = [[CPTableView alloc] init];
...
tableColumn = [[CPTableColumn alloc] initWithIdentifier:@"identifierText"];
[tableColumn setEditable:YES];
[tableView addTableColumn:tableColumn];

textField = [[CPTextField alloc] init];
[textField setVerticalAlignment:CPTopVerticalTextAlignment];
[...]

PR #2765, opened 2018-08-26
FIXED: Support for CPLineBreakMode in editing mode

jasperdeb:
Pull request following on issue #2708.

Changing the values for "whiteSpace" from pre to pre-wrap seems to fix the problem for me. Don't really understand why the lines in the CPTableView where wrapping correctly when displayed but not edited.

- pre | Whitespace is preserved by the browser. Text will only wrap on line breaks. Acts like the pre tag in HTML
- pre-wrap | Whitespace is preserved by the browser. Text will wrap when necessary, and on line breaks

The fix appears to only work when using the method [textField _setWraps:YES];

### DIFF
--- a/Foundation/CPAttributedString.j
+++ b/Foundation/CPAttributedString.j
@@ -27,6 +27,8 @@
 @import "CPRange.j"
 @import "CPString.j"
 
+CPAttachmentCharacter = "\ufffc";
+
 /*!
     @class CPAttributedString
     @ingroup foundation
@@ -767,6 +769,11 @@
     {
         return copyRangeEntry(entry);
     }];
+}
+
++ (CPAttributedString)attributedStringWithAttachment//(CPTextAttachment)attachment
+{
+    return [self initWithString:CPAttachmentCharacter];
 }
 
 //Private methods
--- a/Foundation/CPAttributedString.j
+++ b/Foundation/CPAttributedString.j
@@ -27,8 +27,6 @@
 @import "CPRange.j"
 @import "CPString.j"
 
-CPAttachmentCharacter = "\ufffc";
-
 /*!
     @class CPAttributedString
     @ingroup foundation
@@ -769,11 +767,6 @@ CPAttachmentCharacter = "\ufffc";
     {
         return copyRangeEntry(entry);
     }];
-}
-
-+ (CPAttributedString)attributedStringWithAttachment//(CPTextAttachment)attachment
-{
-    return [self initWithString:CPAttachmentCharacter];
 }
 
 //Private methods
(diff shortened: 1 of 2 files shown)


mrcarlberg:
@cacaodev @daboe01 Can you two come with input on this fix. IS this a good solution?

Also, is this something that should be included in version 1.0?

jasperdeb:
I agree that this is a little more complicated if all line break modes should be supported, and this only helps my specific case. Also changing the defaults should not be necessary.

A problem seems to be that not all of them have a direct equivalent in the whitespace property (as for example CPLineBreakByCharWrapping).

For reference:
[https://www.w3schools.com/jsref/prop_style_whitespace.asp](url)

mrcarlberg:
It is ok to just solve your specific case but we have to make sure we don't break anything. It is of course better to solve all cases if possible.

jasperdeb:
I have copied the code from _CPImageAndTextView.j which sets the DOM-properties for the lineBreakMode and restored the default settings as they are in the current master.

What do you think of this fix?

jasperdeb:
@cacaodev Do you agree with this fix? Or do I need to improve further?

cacaodev:
@jasperdeb Thanks for this !
It would be great if you could create a quick manual test for this. Something that shows how CPTextField behaves with these line breaks in editing and non editing modes.

jasperdeb:
@cacaodev A manual example is made on [this](https://bouwcloud.be/Reduction2/) link. This shows the current bad behaviour in the current master. The code is available in [this](https://bouwcloud.be/Reduction2/AppController.j) link.

jasperdeb:
Now with creating this text, I see a different behaviour in the browsers. In Safari there was nothing to fix. In Chrome and Firefox my PR is needed.

Foundation/CPAttributedString.j
```diff
@@ -769,6 +771,11 @@
     }];
 }
 
++ (CPAttributedString)attributedStringWithAttachment//(CPTextAttachment)attachment
```
daboe01: this is not useful as is. we the attachment in order to generate an attributedStringWithAttachment

daboe01:
@jasperdeb please do not mix topics in one PR.
please undo the unrelated commits and create a separate PR for attributedStringWithAttachment

Foundation/CPAttributedString.j
jasperdeb: I saw this morning that I accididentally mixed up and I already undo the commit in this branch.

mrcarlberg:
@jasperdeb @cacaodev @daboe01 What is the status of this pull request? I want to move forward with this.

jasperdeb:
@mrcarlberg I accidentally deleted the branch with the fix as I accidentally mixed commits for something else into this. Should I close and reopen? Or can I attach a new branch to this PR?

mrcarlberg:
Github has some kind of copy of your branch but I'm not exactly sure how to get hold of it. Try to google for an answer. Should be possible to retrieve it.

jasperdeb:
I made a new PR with the code change. Therefore I close this one...

PR #2767, opened 2018-08-28
FIXED: This PR fixes a side effect of #2744 on table data views

### LINKED ISSUE #2766: Setting the color of text in a CPTableview row
Until now I have used a version of the master (downloaded and compiled 12th of March), now today I have compiled the latest version.

Before, the way I changed the text color in a cell-based CPTableView row in the following delegate

- (void)tableView:(CPTableView)aTableView willDisplayView:(id)aView forTableColumn:(CPTableColumn)aTableColumn row:(CPInteger)rowIndex
{ 
   [aView setTextColor: [CPColor redColor]];
} 

This used to work, but no longer in the master I have compiled today.

didierkorthoudt:
**Problem**
Before this PR, using, for example, `setTextColor` on a CPTextField used in a tableView would have no effect.

**Explanation**
PR #2744 corrected the way a theme attribute returns its value for a theme state. This enables a more precise theme state specified in the theme to be taken in account even if a more wider theme state is specified in the view.

But this has impact on textFields used in a tableView because CPControl `setTextColor` (for example) uses `[self setValue:aColor forThemeAttribute:@"text-color"]` which will result in setting this value for the `CPThemeStateNormal`. But, when in a tableView, the "normal" state is not `CPThemeStateNormal` but `CPThemeStateTableDataView`.

**Solution**
When setting / getting a value without specifying a theme state, if the view has themeState `CPThemeStateTableDataView`, then the value is set/get for themeState `CPThemeStateTableDataView` ; if not, `CPThemeStateNormal` is used as before.

Fixes #2766 and #2708


didierkorthoudt:
@cacaodev @mrcarlberg Please wait. I've a better / cleaner fix for this. Performing tests as spotted by @jasperdeb.

jasperdeb:
Just did a quick test with the tableviews in my app where I have seen the issue, and it seems to fix the issue with the setTextColor AND the top vertical alignment.
@didierkorthoudt Great job!

didierkorthoudt:
@jasperdeb Thank you. The problem when you fix an old old bad behavior, is that code has been written using this behavior. So fixing introduces side effects on this "not directly related" code. Sorry for this.

didierkorthoudt:
@mrcarlberg I think we're ready now for this one.

didierkorthoudt:
Just to be complete : CPTextField and CPCheckBox override setValue:forThemeAttribute to store the value for 2 theme states : CPThemeStateNormal (the default behavior) and CPThemeStateTableDataView.

Why storing in CPThemeStateTableDataView ? Because when the control is in a table view, the "lowest" theme state is CPThemeStateTableDataView, not CPThemeStateNormal. This explains issues #2766 and #2708.

Why storing in the 2 theme states instead of storing in CPThemeStateNormal when the control is outside a table view and in CPThemeStateTableDataView when it is inside ? Because you can create the control, set some values and only then put it in a table view. So values would be stores in CPThemeStateNormal (as control is not yet in a table view) but after being inserted in the table, values will be read from CPThemeStateTableDataView.

mrcarlberg:
Merged! Thanks!

PR #2768, opened 2018-08-28
FIXED: Before this PR, CPScrollView suffered from a side effect of #2718

didierkorthoudt:
#2718 introduced a side effect on CPScrollView scrollers. Removing a guardian condition on `-(void)fadeOut` revealed an incorrect management of `allowFadingOut` property.


mrcarlberg:
Merged! Thanks!

PR #2769, opened 2018-08-28
Add --rsync-base option to bootstrap.sh

enquora:
Existing --copy-local option to use local cappuccino-base has following limitations which are cumbersome when updating bootstrap process
* local source must be a zip archive - which requires an extra step after making changes to source
* local zip archive is erased after install - which makes repetitive testing more difficult

This commit intended to allow testing of updates to cappuccino-base narwhal-jsc.
With macOS 10.14 and later, narwhal-jsc must use modern c++ standard library.
libedit must do the same, so the dependency reference to it is updated to version which is current at this time.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -185,6 +185,7 @@ while [ $# -gt 0 ]; do
         --copy-local)       local_distrib="$2"; shift;;
         --github-user)      github_user="$2"; shift;;
         --github-ref)       github_ref="$2"; shift;;
+        --rsync-base)       local_base="$2"; shift;;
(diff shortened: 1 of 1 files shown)


mrcarlberg:
To test this pull request you have to checkout this branch and bootstrap with the following command:
```./bootstrap.sh --clone --github-user enquora --github-ref feature/64bit-jsc```

mrcarlberg:
Merged! Thanks!

aljungberg:
Fixed by 9ddbdfb80a08274c5ecae76be8606a877189eeba.

#fixed

PR #2773, opened 2018-09-01
Update XcodeCapp code and Jakefile to install properly and launch wit…

enquora:
…hout crashing when built with Xcode 10/ macOSSDK10.14

Disable YRKSpinningProgressIndicator threaded animation (causing immediate crash on launch since Xcode 10 beta 6).
Simplify Jakefile build process to 'xcodebuild install' to avoid permissions problems when building (still requires sudo when using Cappuccino defaults).
Increment version number to 4.0.1 to reflect minor nature of change.
Update About window version number field to fully display and properly align longer version numbers.

### DIFF
--- a/Tools/XcodeCapp/Jakefile
+++ b/Tools/XcodeCapp/Jakefile
@@ -44,14 +44,14 @@ task ("build", function()
 
     if (executableExists("xcodebuild"))
     {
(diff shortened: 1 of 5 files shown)


mrcarlberg:
Merged! Thanks!

PR #2777, opened 2018-09-30
FIXED: Before this PR, when a CPSearchField predicate was bound specifying a predicate format using more than one time the value of the search field ($value), only the first occurence was correctly set, others were set to null

didierkorthoudt:
This was due to how the predicate format was used : all ```$value``` were replaced by ```%@```, then used in ```[CPPredicate predicateWithFormat:formatString, aValue]``` (so specifying only once the value).

The fix simply replaces all ```$value``` directly by the value (surrounded by quotes).


didierkorthoudt:
milestone=1.1

cacaodev:
Is there a manual test where we can see this ? If not, can you add it to an existing one a provide a new one. tx.

mrcarlberg:
What if the "value" is a double quote (")? I guess that work now but will not after this PR?

didierkorthoudt:
Better solution pushed. This one avoids the problem of double quotes.
One can use manual test `CPSearchFieldPredicateBinding` and test with following predicate format : `name beginswith $value OR name endswith $value`

AppKit/CPSearchField.j
```diff
@@ -915,16 +915,25 @@ var CPRecentsAutosaveNameKey            = @"CPRecentsAutosaveNameKey",
...
 
     if (aValue)
-        [_controller setFilterPredicate:[CPPredicate predicateWithFormat:formatString, aValue]];
+    {
+        var valueCount   = [[_predicateFormat componentsSeparatedByString:@"$value"] count] - 1,
+            formatString = _predicateFormat.replace(/\$value/g, "%@"),
+            values       = @[];
+
+        for (var i = 0; i < valueCount; i++)
+            [values addObject:aValue];
```
mrcarlberg, line 927: Nice creative solution here 😃 

What about doing it like this?
```
var values = @[],
    formatString = _predicateFormat.replace(/\$value/g, function(x) {[values addObject:aValue]; return "%@";});             
```

didierkorthoudt: Even more creative !.. 😄
I love this one !.. (I should have found this by myself 😉)
OK, I'll adapt ASAP.

mrcarlberg:
Thanks! Merged!

PR #2779, opened 2018-10-06
New: support contextmenus for CPTextView

daboe01:


AppKit/CPTextView/CPTextView.j
```diff
@@ -1134,6 +1154,21 @@ Sets the selection to a range of characters in response to user action.
...
 }
 
+- (CPMenu)menuForEvent:(CPEvent)anEvent
+ {
+     var myMenu = [super menuForEvent:anEvent];
+     if (_selectionRange.length === 0)
+     {
+         [[myMenu itemAtIndex:0] setEnabled:NO];
+         [[myMenu itemAtIndex:1] setEnabled:NO];
+     }
```
Dogild: Why do you need to do this ?

daboe01: this is to mimic cocoa behavior in case of an empty selection: cut and paste in the context menu need to be disabled.

PR #2786, opened 2018-10-27
FIXED: CPStepper was not receiving the enabled status from IB

didierkorthoudt:
Adapted nib2cib to get the enabled status from cell.

Adapted CPStepper to force specific enabling code to be executed.


daboe01:
+AppKit
-#new

PR #2794, opened 2019-02-20
Fixed: the class method +initialize was always sent even if the root class did not implement it.

mrcarlberg:
Forwarding and Invocation worked also on the +initialize method.

To mimic Objective-C correctly the +initialize method should not be sent on a class hierarchy if it does not implement it.
Also the +initialize method should not allowed to be forwarded to another target with Invocation or any other forwarding mechanism.

Added test cases for this and cleaned up some test cases that succeeded for the wrong reason.

This change will not affect many projects as all classes inherit from CPObject.


daboe01:
-#new
+Foundation
+#ready-to-commit

PR #2803, opened 2019-03-28
Fixed: Return type CPPlatformPasteboard.j should return BOOL

bedalton:
Method _mayRequireDOMPasteboardElementHack:flags had return type of void, despite returning a BOOL value, and also being annotated with a doc comment stating that it should return a BOOL value

Some methods had return type "boolean", but should possibly have been BOOL


Dogild:
Thanks for all of these contributions!

PR #2819, opened 2019-03-31
fixed: removed outdated duplicate method definitions in CPFontManager.j

### LINKED ISSUE #2816: CPFontManager has duplicate methods
In CPFontManager class, there are two sets of duplicate methods. The first is 
https://github.com/cappuccino/cappuccino/blob/0be913ab18c9ef74289f87b7f9842b9429705f9e/AppKit/CPFontManager.j#L193-L204
and 
https://github.com/cappuccino/cappuccino/blob/0be913ab18c9ef74289f87b7f9842b9429705f9e/AppKit/CPFontManager.j#L386-L427

The second set is:
https://github.com/cappuccino/cappuccino/blob/0be913ab18c9ef74289f87b7f9842b9429705f9e/AppKit/CPFontManager.j#L206-L220
and 
https://github.com/cappuccino/cappuccino/blob/0be913ab18c9ef74289f87b7f9842b9429705f9e/AppKit/CPFontManager.j#L273-L305

I do not know if the framework handles them differently, as I do not know truly how Objective-J works. They just seem like they would clash.

daboe01:
previously we had duplicate method definitions in CPFontManager for convertFont:toHaveTrait: and convertFont:
these are removed by this PR
fixes #2816

### DIFF
--- a/AppKit/CPFontManager.j
+++ b/AppKit/CPFontManager.j
@@ -190,35 +190,6 @@ CPRemoveTraitFontAction = 7;
     return ([aFont isBold] ? CPBoldFontMask : 0) | ([aFont isItalic] ? CPItalicFontMask : 0);
 }
 
-- (CPFont)convertFont:(CPFont)aFont
-{
-    if (!_activeChange)
-        return aFont;
-
-    var addTraits = [_activeChange valueForKey:@"addTraits"];
-
-    if (addTraits)
-        aFont = [self convertFont:aFont toHaveTrait:addTraits];
-
-    return aFont;
-}
-
-- (CPFont)convertFont:(CPFont)aFont toHaveTrait:(CPFontTraitMask)addTraits
-{
-    if (!aFont)
-        return nil;
-
-    var shouldBeBold = ([aFont isBold] || (addTraits & CPBoldFontMask)) && !(addTraits & CPUnboldFontMask),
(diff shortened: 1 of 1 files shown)


daboe01:
-#new
+#ready-to-commit

AppKit/CPFontManager.j
```diff
@@ -190,35 +190,6 @@ CPRemoveTraitFontAction = 7;
...
-    var shouldBeBold = ([aFont isBold] || (addTraits & CPBoldFontMask)) && !(addTraits & CPUnboldFontMask),
-        shouldBeItalic = ([aFont isItalic] || (addTraits & CPItalicFontMask)) && !(addTraits & CPUnitalicFontMask),
-        shouldBeSize = [aFont size];
-
-    // XXX On the current platform there will always be a bold/italic version of each font, but still leave
-    // || aFont in here for future platforms.
-    aFont = [CPFont _fontWithName:[aFont familyName] size:shouldBeSize bold:shouldBeBold italic:shouldBeItalic] || aFont;
-
-    return aFont;
-}
```
Dogild: It looks like the duplicate method does not have the same body. How do you know which one was the one to keep ?

daboe01: our compiler picks the second definition in case of duplicates.
that aside, i know, because i am the author.

PR #2820, opened 2019-03-31
fixed: typo in function call to _CPDecimalSetZero

### LINKED ISSUE #2818: CPDecimal.j possibly bad function call
In CPDecimal.j  a function CPDecimalSetZero is called. 
 https://github.com/cappuccino/cappuccino/blob/0be913ab18c9ef74289f87b7f9842b9429705f9e/Foundation/CPDecimal.j#L944

A function named CPDecimalSetZero does not appear to exist, but one prefixed with an underscore _CPDecimalSetZero does:
https://github.com/cappuccino/cappuccino/blob/0be913ab18c9ef74289f87b7f9842b9429705f9e/Foundation/CPDecimal.j#L329-L336

Should this function call be changed?

daboe01:
previously, the code in CPDecimalDivide called a nonexistant function under certain conditions.
this is fixed by this PR
fixes #2818

### DIFF
--- a/Foundation/CPDecimal.j
+++ b/Foundation/CPDecimal.j
@@ -941,7 +941,7 @@ function CPDecimalDivide(result, leftOperand, rightOperand, roundingMode)
 
         if (result._exponent + exp < CPDecimalMinExponent)
         {
-            CPDecimalSetZero(result);
+            _CPDecimalSetZero(result);
             return error;
         }
     }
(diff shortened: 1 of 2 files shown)


daboe01:
-#new
+Tests
+#needs-review

daboe01:
this is what happened: the test checked the loss of accuracy exception was fired caused by a division operation. actually, the call to a missing function (namely CPDecimalSetZero) caused a different kind of exception that made the test pass erroneously. this made the test fail after fixing the missing underscore, because the division did not ask for raising the exception that was tested (this feature is not on by defauilt). this PR fixes the test in that the division is done with the loss of accuracy exception enabled.

mrcarlberg:
Merged! Thanks"

PR #2821, opened 2019-03-31
fixed: typo in initWithGraphicsPort: declaration

### LINKED ISSUE #2817: CPGraphicsContext question
In the method declaration:
https://github.com/cappuccino/cappuccino/blob/0be913ab18c9ef74289f87b7f9842b9429705f9e/AppKit/CPGraphicsContext.j#L95
Should the parameter CPContext be CGContext, as CPContext does not seem to exist, and the only use of this method passes in a CGContext object?

daboe01:
title says it all.
fixes #2817


daboe01:
-#new
+#ready-to-commit

PR #2823, opened 2019-03-31
fixed: CPBrowserDelegate_browser_imageValueForItem_ was defined twice

### LINKED ISSUE #2625: CPBrowser.j redeclares variable CPBrowserDelegate_browser_imageValueForItem_
In AppKit/CPBrowser.j the file scoped variable `CPBrowserDelegate_browser_imageValueForItem_`  is declared twice in the same block. First at line 65 and then line 67 with the values `1 << 7` and `1 << 9` respectively. Later on in the code when this variable is referenced, it shows up in two if statements one starting on line 214:
https://github.com/cappuccino/cappuccino/blob/f6d892023e6042682ce7b6e174fcc3386561c176/AppKit/CPBrowser.j#L214-L215
and another on line 220
https://github.com/cappuccino/cappuccino/blob/f6d892023e6042682ce7b6e174fcc3386561c176/AppKit/CPBrowser.j#L220-L224

the if blocks have the same condition statement, and the bodies are almost identical, but the second listing adds `_delegateSupportsImages = YES;`

I don't believe this causes any problems, but it's something that came up when reviewing the code with my plugin.

daboe01:
subject says it all.
fixes #2625


daboe01:
-#new
+#ready-to-commit

PR #2760, opened 2018-08-21
FIXED: Before this PR, nib2cib-adjustment-frame was only usable on CPControl and subclasses

didierkorthoudt:
With this PR, nib2cib-adjustment-frame is now declared and managed at CPView level, so it’s usable for widgets other than controls, like, for example, CPTabView.

PR #2831, opened 2019-04-14
FIXED: Before this PR, only CPControl could use nib2cib-adjustment-frame.

didierkorthoudt:
With this PR, now all CPView can.


didierkorthoudt:
milestone=1.1

daboe01:
-#new
+Theme

Tools/nib2cib/NSView.j
```diff
@@ -25,6 +25,7 @@
 
 @import <AppKit/CPView.j>
 
+@class Nib2Cib
```
mrcarlberg: Should not the corresponding line in NSControl.j be removed?

didierkorthoudt: You're right. Done in new commit.

didierkorthoudt:
I've just noticed that PR #2760 deals with the same matter.
I'll merge in this PR what was complementary done in the older one.
So, please wait until a new commit.

mrcarlberg:
Thanks! Merged!

PR #2844, opened 2019-04-21
FIXED: Before this PR, buttons were sometimes missing their alternate image

didierkorthoudt:
PR #2721 fixed a « feature » of Xcode where IB sometimes stores NSFont values in the alternate image property.

But this fix was rejecting anything beside NSButtonImageSource.

But sometimes, alternate images are stored as NSCustomResource.

This fix rejects the value only if it is an NSFont object.


mrcarlberg:
Merged! Thanks!

PR #2845, opened 2019-04-22
FIXED: Before this PR, cell based table views that use specific font, reverted to standard font when selected/editing

didierkorthoudt:
This was (yet another) side effect of theming system fix.

This PR simply forces the dataView created by nib2cib to set the specified font also for CPThemeStateSelectedDataView and CPThemeStateEditing states.


cacaodev:
Can you nib2cib one of the table manual test where this bug is visible and include it in this PR. tx.

mrcarlberg:
Merged! Thanks!

PR #2846, opened 2019-04-29
FIXED: CPSearchField menu was broken

### LINKED ISSUE #2789: CPSearchField Menu Broken
The CPSearchField menu will not open until a search is performed. This bug shows in the manual test. Also in the manual test, but not shown in my own testing, is that the menu will stay open and become unresponsive if the field is cleared, and possibly also when a menu item is chosen.

didierkorthoudt:
When no recent searches were done, the menu wasn’t showing, even if search categories were specified.

This PR fixes #2789


mrcarlberg:
Merged! Thanks!

PR #2847, opened 2019-05-23
fixed: FF66 does not emit CPEvents anymore for backspace, delete and the arrow keys

daboe01:
before this PR, backspace and the arrow keys did not emit CPEvents anymore
this PR readjusts the cappuccino compatibility layer to the recent changes in FF.


mrcarlberg:
Merged! Thanks!

PR #2849, opened 2019-05-29
Fixed: backspace triggered page back in FireFox

daboe01:
before this patch, the backspace key caused the page back action in firefox.
this is almost always unwanted by the user and not the case in chrome/safari.
this patches cancels the backspace event if no editing takes place.


mrcarlberg:
Merged! Thanks!

PR #2850, opened 2019-06-05
Fixed: If sourcemap is available include it when building with a nondebug version of the tools.

mrcarlberg:
The sourcemap was not included when building from command line with the release (the default) version of the tools. This results in compiled code without the source map included for debug


mrcarlberg:
Merged! Thanks!

PR #2852, opened 2019-06-24
New: msgSend will return undefined instead of nil when the receiver is undefined [+1]

mrcarlberg:
The ```msgSend``` function has always returned ```nil``` if the receiver has been ```nil``` or ```undefined```.
Example:
```
var a = undefined;
var b = [a someSelector]; // b = nil
```

The variable ```b``` is now ```nil```.

This pull request adds the ability for the ```msgSend``` function to return ```undefined``` if the receiver is ```undefined```. If the receiver is ```nil``` it still return ```nil```.
The above example again:
```
var a = undefined;
var b = [a someSelector]; // b = undefined
```

The variable ```b``` is now ```undefined```.

### Background.
The use of ```nil``` in Objective-C corresponds to the use of ```null``` in C and C++ and stands for "no value". A fundamental function in Objective-C is when you send a message to a receiver that is ```nil``` the result is again a ```nil``` value. This can be very convenient as you don't need the check if a receiver is nil before sending a message to it.

Objective-J adds the ability the use class structures and message send functionality for Javascript in the same way as Objective-C adds this to C and C++. Javascript also has the value ```null``` and is handled in the same way for Objective-J. Javascript also has the value ```undefined``` that stands for "not yet defined". This can sometimes be confusing and many try to handle ```null``` and ```undefined``` as the same thing. This is almost what Objective-J and the Cappuccino frameworks always has been doing. They try to always check for both ```null``` and ```undefined``` and always return ```nil``` but never ```undefined```. We can say that Objective-J understands ```undefined``` but will try to translate it to ```nil```.

### Why do we need this change.
Javascript has both ```null``` and ```undefined``` and Objective-J is a superset of Javascript. It is now very hard to use ```undefined``` in Objective-J code as it will always try to translate it to ```nil```. There are some Javascript libraries that use both ```null``` and ```undefined``` values and they are very hard to use from Objective-J programs today.  Same if you decide to use ```nil``` and ```undefined``` as a value in your own Objective-J code. This small change will make ```undefined``` a full member of the Objective-J language. It should be as it is a superset of Javascript. The Cappuccino frameworks will still handle ```undefined``` but only return ```nil```.

### How will it impact.
Today most methods will never return ```undefined``` so this is a very limited problem. If any does this the main concern is if some code will only check for ```nil``` and not ```undefined```.
Example:
```
var a = [someObject someSelector];
If (a === nil) a = 42;
...
```
If the variable ```someObject``` contains ```undefined``` the variable ```a``` will now be set to ```undefined``` instead of ```nil```.  But the variable ```a``` will then never be set to 42.
The correct way is to check for both ```nil``` and ```undefined``` like this:
```
var a = [someObject someSelector];
If (a == nil) a = 42;
...
```

I have tried this pull request on our project that is fairly large and I don't find any problems.

### DIFF
--- a/AppKit/CPBrowser.j
+++ b/AppKit/CPBrowser.j
@@ -463,7 +463,7 @@ var CPBrowserDelegate_browser_acceptDrop_atRow_column_dropOperation_            
 
 - (id)itemAtRow:(CPInteger)row inColumn:(CPInteger)column
 {
-    return [_tableDelegates[column] childAtIndex:row];
+    return [_tableDelegates[column] childAtIndex:row] || nil;
 }
 
 - (BOOL)isLeafItem:(id)item
@@ -473,7 +473,7 @@ var CPBrowserDelegate_browser_acceptDrop_atRow_column_dropOperation_            
 
 - (id)parentForItemsInColumn:(CPInteger)column
 {
-    return [_tableDelegates[column] _item];
+    return [_tableDelegates[column] _item] || nil;
 }
 
 - (CPSet)selectedItems
--- a/Objective-J/ObjJAcornCompiler.js
+++ b/Objective-J/ObjJAcornCompiler.js
@@ -2447,7 +2447,9 @@ ArrayLiteral: function(node, st, c) {
             buffer.concat(++st.receiverLevel + "");
             buffer.concat(" = (CPArray.isa.method_msgSend[\"alloc\"] || _objj_forward)(CPArray, \"alloc\"), ___r");
             buffer.concat(st.receiverLevel + "");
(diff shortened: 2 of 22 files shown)


daboe01:
+1

daboe01:
-#new
+Objective-J

cacaodev:
@mrcarlberg do you have anything to add ?

mrcarlberg:
@cacaodev No, this pull request is ready to commit I guess. We have used it in our project for a couple of months now.

cacaodev:
Merged. Thanks for the detailed commit message.

PR #2853, opened 2019-06-26
Fixed: Optimization for bindings with longer key paths in table views.

mrcarlberg:
This prevents a buildup of large arrays with a lot of values and then choose one row and throw away the rest. This is great for speed and a must for lazy loading.

Previous there was a optimization for this but it was only effective for short combined key path like ```a.b```. This pull request makes it effective for any kind of length of a combined key path like ```a.b.c.d...```.

This is done by returning the object at a row in the first found CPArray in the key path
that is divided in a first and second part.
The first part is never a combined key path. The second part can be a combined key path.
If this optimization is not done we will create an array with the valueForKeyPath value on each row and then pick the wanted value for the row and throw away all the other rows. It is much more effective to first pick the row and then do the valueForKeyPath on the rest of the key path.

### DIFF
--- a/AppKit/CPTableColumn.j
+++ b/AppKit/CPTableColumn.j
@@ -639,6 +639,53 @@ CPTableColumnUserResizingMask   = 1 << 1;
     }
 }
 
+
+/*!
+    @ignore
+
+    This method will return the object at a row in the first found CPArray in the key path
+    that is divided in a first and second part.
+    The first part is never a combined key path. The second part can be a combined key path.
+    If this optimization is not done we will create an array with the valueForKeyPath value on each row and then pick
+    the wanted value for the row and throw away all the other rows. It is much more effective to first
+    pick the row and then do the valueForKeyPath on the rest of the key path.
+    When the second part is depleated it will stop the search for a CPArray and return the
+    current object for the first part. The second part will then be nil.
+    The secondPartRef will always be updated with the rest of the key path that can be applied
+    to the returned object.
+    It will stop the search if the object is nil
+*/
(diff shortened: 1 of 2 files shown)


mrcarlberg:
@cacaodev You have a lot of understanding of how the table view is working. Can you take a look at this pull request?

cacaodev:
Hi @mrcarlberg,
I think it is a very good idea. If I understand well, previously asking for a specific row of "array.value" was optimized but not "controller.array.value", right ?
To go further, we could also support fetching multiple rows (CPIndexSet) instead of one row. Does it make sense for the table column binding ?

mrcarlberg:
@cacaodev
Before it was optimized for ```controller.array.value``` too as it divided the path into two parts, ```controller.array``` and ```value```. It then took out the row from ```controller.array``` and just the value for ```value``` on that row.
It was not optimized for ```controller.array.a.value``` as it was devided the path into ```controller.array.a``` and ```value```. When it takes out the row from ```controller.array.a``` it first has to get the ```.a``` value for all rows. Pick the row and throw away all the other values. Then continue with the ```value``` value.

With this optimization it will take each subpart of the key path and look for an array. It will start with ```controller``` and as it will not be an array it will continue with ```controller.array```. As it is an array it will stop and pick the row. After this it will take the value for the rest of the key path ```a.value``` on just that row.

This is much more effective on complex key paths. Specially if a lazy loading framework like LightObject is used.

To support fetching multiple rows might be good but I think that is something for the future.

mrcarlberg:
@cacaodev Are you ok with this?

cacaodev:
Merged, thanks !

PR #2854, opened 2019-06-26
New: Allow lazy loading control for combined observer key paths

mrcarlberg:
This pull request adds a ```kvoValueForKey:``` method used by ```_CPKVOForwardingObserver``` for getting values.
The default action is to just return the ```valueForKey:``` value.
Override this method if you want any other behavior.
It can be used to just return ```nil``` if you are implementing a lazy load behavior on a class.
It can allow a combined binding to observe this object without trigger any lazy loading.

### DIFF
--- a/Foundation/CPKeyValueObserving.j
+++ b/Foundation/CPKeyValueObserving.j
@@ -224,6 +224,18 @@
     }
 }
 
+/*!
+    This valueForKey method is used by _CPKVOForwardingObserver when getting values.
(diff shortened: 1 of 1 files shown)


daboe01:
-#new
+Bindings
+#needs-review

mrcarlberg:
This has been working in our projects for the last 6 months

mrcarlberg:
Merged! Thanks!

PR #2857, opened 2019-08-18
fixed: pressing capslock inserted whitespaces under certain conditions

daboe01:
the original implementation of CPPlatformWindow+DOM.j was faulty in two ways: firstly, _capsLockActive was accidentally a global variable, secondly, pressing capslock accidentally emitted keyDown:(CPEvent)events

both issues are fixed in this PR

### DIFF
--- a/AppKit/Platform/DOM/CPPlatformWindow+DOM.j
+++ b/AppKit/Platform/DOM/CPPlatformWindow+DOM.j
@@ -747,6 +747,8 @@ var _capsLockActive;
             if (_keyCode === CPKeyCodes.CAPS_LOCK)
             {
                 _capsLockActive = YES;
+
+                // we need to break out in order to prevent a keyDown: event from pressing the CAPS_LOCK key
                 break;
             }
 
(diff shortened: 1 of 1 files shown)


daboe01:
-#new
+AppKit
+#needs-review
+TextView

cacaodev:
Can you remove the leading underscore like the other class variables.

mrcarlberg:
@daboe01 ```_ capsLockActive``` is an instance variable and not a global variable. Do we need to change that?

daboe01:
@mrcarlberg _capsLockActive used to be a global variable. my PR turns this into a class variable of name capsLockActive (not an instance variable). this makes sense to me, because we can only have one caps-lock state at a time.

mrcarlberg:
@daboe01 I think ```_capsLockActive``` is declared as an instance variable on line 72 in the file```Platform/CPPlatform.j```. ```Platform/DOM/CPPlatformWindow+DOM.j``` is just a category for the class.

daboe01:
@mrcarlberg thank you for spotting this. i adjusted the fix to leave the ivar alone and just fix the actual issue.

mrcarlberg:
@daboe01 I looked into this again and what do you think of solving this in another way?
A press on the caps lock key should result in a flagsChanged event.
I have also included the caps lock state in the modifierFlags. I believe this is how Cocoa works.

mrcarlberg:
Please check my commit above

daboe01:
@mrcarlberg looks like your approach also fixes the issue. good job!

daboe01:
-#needs-improvement
-#needs-review
+#ready-to-commit

mrcarlberg:
Merged! Thanks!

PR #2860, opened 2019-09-19
clearsFilterPredicateOnInsertion option is for insertion only

### LINKED ISSUE #2791: FilterPredicate Binding of CPArrayController Not Being Applied
When set in xib using IB, checking the filterPredicate during runtime returns nil. Latest master, Xcode 10.1

cacaodev:
Previously, the filter predicate was removed on setContent: or on content binding. When this option is on, the FP should be removed only on insertion, as the name of the option suggests it.

With Tests: Foundation unit test.
fixes #2791

### DIFF
--- a/AppKit/CPArrayController.j
+++ b/AppKit/CPArrayController.j
@@ -297,27 +297,18 @@
         class.
     */
 
-    if (_clearsFilterPredicateOnInsertion)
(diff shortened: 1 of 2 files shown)


mrcarlberg:
Merged! Thanks!

PR #2862, opened 2019-09-27
Fixed: Make sure we handle undefined when testing for nil values [+1]

mrcarlberg:
The Objective-J language handles both nil/null and undefined after this pull request #2852. The Cappuccino frameworks does not always handle both nil/null and undefined. This commit will make sure most compares always checks for undefined too.


daboe01:
+1

cacaodev:
Conflicts need to be resolved ...

mrcarlberg:
Merged! Thanks!

PR #2863, opened 2019-09-30
Fix strange lines that appear in Google Chrome

davidsiaw:
## Context 

This PR attempts to fix this bug for window shadows by preventing the side shadows from wrapping around creating random thin lines at particular offsets.

On google chrome very strange lines have been appearing due to divs not being exactly pixel perfect in size on chrome:

image

By making the CSS repeat only along the axes that they need to, we eliminate the wrap-around repeats that cause the dark lines to appear.

## Addendum

This could be a bug with Chrome but I struggle to find any bug report that describes this. I can't be the only one to see this though.

This does not happen on Firefox as far as I can see.

However, the bug continues to manifest itself with the dark edges of the HUD window. Presumably because usually the fill covers the entire area.

image

I could not find an issue on this on the issue tracker so I just decided I would fix it. Please leave notes if you have any better ideas on what exactly is going on. These visual quirks are extremely distracting.


daboe01:
-#new
+Chrome
+Theme

daboe01:
+1

davidsiaw:
Hi, would be nice if a reviewer could examine this issue

daboe01:
+#ready-to-commit

PR #2864, opened 2019-09-30
Misc Fixes

davidsiaw:
Hello Cappuccino maintainers

I enjoy using cappuccino and over the time I have been using it I discovered bugs. These fixes have been piling up and I think it is time I made them available to the main repo so people can enjoy the fixes.

Here is a summary of fixes in this PR:
- Break infinite recursion under chrome when removing tokens in a token field
- Fix stack overflow happening when CPOutline view receives a drop
- Fix some errors during compilation, and make a warning about 32KB image data-urls.

### DIFF
--- a/AppKit/CPOutlineView.j
+++ b/AppKit/CPOutlineView.j
@@ -757,7 +757,7 @@ var CPOutlineViewCoalesceSelectionNotificationStateOff  = 0,
 
     var children = itemInfo.children;
 
-    for (var i = [children count]; i >= 0; i--)
+    for (var i = [children count] - 1; i >= 0; i--)
     {
         var child = children[i];
         [self _addPendingItem:child];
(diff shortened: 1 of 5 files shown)


mrcarlberg:
Thanks David for these fixes.

It is simpler for us if you make separate pull requests for each fix. Sometimes we have issues with something and it will delay the other parts of the commit. It is also easier to discuss each fix in a separate thread.

Also, we want to have have a special style and format on commit messages. Please read more on this [here](https://github.com/cappuccino/cappuccino/blob/master/CONTRIBUTING.md)

Finally I have a question about when you are testing if a value is nil? Why do you use `!aValue`? It is not the same as `aValue == nil`.

davidsiaw:
Well that is embarrassing. Thanks for your comment. I will cut this up into smaller PRs when I can.

PR #2866, opened 2019-10-18
Fixed: CPSet enumerateObjectsUsingBlock: will honor the stop variable

mrcarlberg:
Test case also added.

Deprecated the very old behaviour that the block could return YES to abort the enumeration.


daboe01:
-#new
+Foundation

PR #2867, opened 2019-10-18
Fixed: In modal mode a focus event will make the modal window the key window

mrcarlberg:
The default was to make the first window in the most background layer the key window if nothing else was used.
This does not work well if there is a modal window that should have all the attention.

This solution will first choose the modal window if in modal mode. If not the old variant is used.


daboe01:
-#new
+AppKit

daboe01:
looks good to me

daboe01:
+#ready-to-commit

mrcarlberg:
Merged! Thanks!

PR #2874, opened 2019-11-28
Fixed: CPTextView keyboard issue in recent Firefox versions

### LINKED ISSUE #2873: entering hashtag key adds a pound-key in FF70
'#' gives you £#

daboe01:
previously, wrong characters were emitted from keyDown: events in more recent Firefox versions. The reason is that keyCodes do not correspond to characters but rather refer to keys on the keyboard.

The bug is exposed in CPTextView that relies on keyDown: events. CPTextFields are not affected because they are based native input elements.

This patch uses the key property that corresponds to characters and not keys.
Works for me on FireFox, Safari and Chrome.

fixes #2873

### DIFF
--- a/AppKit/Platform/DOM/CPPlatformWindow+DOM.j
+++ b/AppKit/Platform/DOM/CPPlatformWindow+DOM.j
@@ -739,8 +739,8 @@ _CPPlatformWindowWillCloseNotification = @"_CPPlatformWindowWillCloseNotificatio
                 characters = KeyCodesToUnicodeMap[_keyCode];
 
             // The problem with keyCode is that this property refers to keys on the keyboard and not to characters
-            // This is why String.fromCharCode does not work in more recent versions of Firefox
-            // E.g. pressing a '#' gives you a charCode of 163, which refers to '£' and not '#'
+            // This is why String.fromCharCode does not always work in more recent versions of Firefox
+            // E.g. pressing a '#' on a German keyboard gives you a charCode of 163, which refers to '£' and not '#'
             // The property key works fine, though. From there we can get the actual character more robustly.
             // Therefore we prefer key over keyCode whenever possible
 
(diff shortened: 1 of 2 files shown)


daboe01:
-#new
+FireFox
+AppKit

daboe01:
+#needs-review

mrcarlberg:
Can you please add some comments in the code that describe this behavior. Also why this works better. Some examples in the comments would be fantastic too.

daboe01:
@mrcarlberg done so. is this sufficient?

daboe01:
-#needs-review

daboe01:
@mrcarlberg my userbase of ~100 per day has been testing this for 2 weeks (we are unfortunately a FF shop). no issues have been reported so far.

mrcarlberg:
After looking into this a bit I think this might fix the problem in Firefox but the base problem is that `event.charCode` and `event.keyCode` are deprecated. The fix make sure we use the `key` property when the length is one but most of the logic still use the old properties.

Do you know if `charCode` corresponds the the `code` property all the time?

daboe01:
@mrcarlberg the issue with firefox is that String.fromCharCode(e.charCode) does not always match the key that was pressed, at least not so on a german keyboard.

KeyboardEvent.key is the correct alternative to get the key that was pressed according to https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key 

we already have several other workarounds for weird charCode behaviors scattered around 
 CPPlatformWindow+DOM.j

daboe01:
@mrcarlberg this issue is a real show stopper in Germany. Here, FF-usage is still very prevalent. I therefore suggest to merge this fix sooner than later. IMHO a full rework of CPPlatformWindow+DOM.j to use key instead of charCode should be pursued in a second step.

mrcarlberg:
Yes, a rework of this has to be done in the future. I will merge this.

mrcarlberg:
Merged! Thanks!

PR #2876, opened 2019-12-09
NEW: CPSplitView modern Cocoa behavior implementation

didierkorthoudt:
With the PR, split views implement modern Cocoa behavior as of OSX 10.13 (High Sierra).

Please refer to the Cocoa documentation at : https://developer.apple.com/documentation/appkit/nssplitview?language=objc

Please note that we don't (yet) support NSSplitViewController, so neither NSSplitViewItem. We also don't support holding priority.

In this new implementation, we also get rid of direct DOM elements manipulations to render dividers.

Compatibility with Aristo2 should be assumed by this PR.


cacaodev:
This looks very cool but also very hard to review.
Ideally could you split the PR into commits on a per-feature basis. One for arrangedSubviews support, one related to layout, one for the Aristo 3 support, one for the tracking areas modifications. It would really ease te process. tx.

didierkorthoudt:
@cacaodev Well, after reviewing the code, it seems difficult… New layout management and arrangedSubviews introduction are really tied… What could maybe done is to split modern behavior & Aristo3 support. But, for the latest, comments in the code are directly self explaining (I think).

AppKit/CPSplitView.j
```diff
@@ -1453,45 +1787,63 @@ var CPSplitViewDelegateKey          = "CPSplitViewDelegateKey",
...
+
+        _isCSSBased = [[self theme] isCSSBased];
+
+        // Final operations will be performed in _finalizeInitWithCoder
     }
 
     return self;
 }
 
+- (void)_finalizeInitWithCoder
```
mrcarlberg: Can't we just rename this method to `- (id)_cibInstantiate` and add a `return self;` at the end of it? With that we could remove the `performSelector:` on the runloop above? It will be called upon after all the views and objects in the cib are initialized.

didierkorthoudt: New commit implementing your suggestion. Works in my testing.

mrcarlberg:
Merged! Thanks!

PR #2872, opened 2019-11-21
fixed: attachment support was missing in CPTextView

daboe01:
Previously, CPTextView was only able to handle text.
This PR introduces support for embedding CPViews into the text.
You can use this via the CPTextStorage method ```-attributedStringWithAttachment:(CPView)someView ```
These are treated as individual characters with their respective rectangular sizes.

PR #2877, opened 2019-12-10
fixed: replace hard coded background color with themed color

daboe01:
previously, CPTextView used a card coded white background.
This PR makes the background color themeable

### DIFF
--- a/AppKit/CPTextView/CPTextView.j
+++ b/AppKit/CPTextView/CPTextView.j
@@ -190,6 +190,11 @@ var kDelegateRespondsTo_textShouldBeginEditing                                  
 #pragma mark -
 #pragma mark Class methods
 
++ (CPString)defaultThemeClass
+{
+    return "textview";
+}
+
 + (CPDictionary)themeAttributes
 {
     return @{
--- a/AppKit/CPTextView/CPTextView.j
+++ b/AppKit/CPTextView/CPTextView.j
@@ -198,7 +198,8 @@ var kDelegateRespondsTo_textShouldBeginEditing                                  
 + (CPDictionary)themeAttributes
 {
     return @{
-            @"bezel-color": [CPNull null]
+            @"background-color": [CPNull null],
+            @"content-inset": [CPNull null]
         };
 }
 
--- a/AppKit/CPTextView/CPTextView.j
+++ b/AppKit/CPTextView/CPTextView.j
@@ -250,8 +250,10 @@ var kDelegateRespondsTo_textShouldBeginEditing                                  
 #endif
 
     _selectionRange = CPMakeRange(0, 0);
-    _textContainerInset = [self currentValueForThemeAttribute:@"content-inset"];
-    _textContainerOrigin = CGPointMake(_bounds.origin.x, _bounds.origin.y);
+
+    // We have to || with CGSizeMakeZero() for theme compilation
+    _textContainerInset = [self currentValueForThemeAttribute:@"content-inset"] || CGSizeMakeZero();
+    _textContainerOrigin = CGPointMake(_bounds.origin.x + _textContainerInset.width, _bounds.origin.y + _textContainerInset.height);
 
     _selectionGranularity = CPSelectByCharacter;
 
--- a/AppKit/CPTextView/CPTextView.j
+++ b/AppKit/CPTextView/CPTextView.j
@@ -198,7 +198,7 @@ var kDelegateRespondsTo_textShouldBeginEditing                                  
 + (CPDictionary)themeAttributes
 {
     return @{
-            @"background-color": [CPColor whiteColor],
+            @"background-color": [CPColor textBackgroundColor],
             @"content-inset": CGSizeMake(2, 0)
         };
 }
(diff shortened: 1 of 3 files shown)


kerusan:
The setting of the background color need to be set whenever there is a state change. I think that maybe the draRect: method is where it can be. Now it is set only in initWithFrame: and during  drag operation.

kerusan:
Also there has to be a method called +(CPString)defaultThemeClass

kerusan:
As we in our project use a pattern color instead of color values, there has to be a style setting that sets transparency in the pattern, like

#if PLATFORM(DOM)
        _DOMElement.style.cursor = "text";
        _DOMElement.style.background = "transparent";

#endif

Then you can use both variants.

kerusan:
The background color can be be set as a background pattern and then the content inset has to be altered. So we must have that "content-inset" value too.

didierkorthoudt:
Found the caret position problem origin. Testing a fix.

didierkorthoudt:
Sadly not enough…

daboe01:
@didierkorthoudt i reworked cared positioning in #2872. maybe @mrcarlberg can merge #2872 in?

didierkorthoudt:
@daboe01 Did you also adapt the initial caret position, like in my last commit ?

daboe01:
@didierkorthoudt no, i missed that. no merge conflict ahead. thank you!

mrcarlberg:
Great work on this!

As this is a text view we also need to set a couple of theme states depending on what state the view is in. There is a need for this so the background can be different depending on the state. Examples of this could be `CPThemeStateEditable`, `CPThemeStateHovered` and `CPThemeStateDisabled`.

mrcarlberg:
Also, what about the suggestion from @kerusan to make the background style `transparent`?

didierkorthoudt:
@mrcarlberg I could add the various theme states, just be aware that this is a disruption from Cocoa (which is not a problem). Regarding the transparency of the background, we should then add a property ?

kerusan:
@didierkorthoudt  Since theming is a "disruption" from Cocoa this could be done anyway.

mrcarlberg:
@didierkorthoudt Do you mean that a text view in Cocoa does have the same look when it is being edit or being disabled?

kerusan:
@didierkorthoudt If you compare with the CPTextField it has no property for transparence. I think the main reason for seeing the dom property "background" to "transparent" was to allow setting of color patterns. There it makes sense.

didierkorthoudt:
@mrcarlberg For what I can see on a test app, yes, there's no visual clue that a text field is editable or not (there's no disabled state !)... Sometimes Cocoa is not very consistent. There's even no focus ring !..

didierkorthoudt:
@kerusan OK, I'll add this  ;-)

didierkorthoudt:
@kerusan Well, maybe that is could be done with a CSS Color (the one created for Aristo3 and normally already present in 1.0) ? What do you think ?

mrcarlberg:
@didierkorthoudt Ok, but we have these states on CPTextField. I think we should add those on CPTextView too. So we are compliant with our self 😄

didierkorthoudt:
@mrcarlberg OK, that's a perfect criteria !..  😄  I'll work on this.

kerusan:
@didierkorthoudt I'm not familiar with the CSS solution I just tried to copy the CPTextField solution and it works ;-)

daboe01:
-#new
+TextView
+Theme

daboe01:
@mrcarlberg can you please elaborate what needs to be done to get this ready to commit?

daboe01:
@mrcarlberg looks good to me. what needs to be improved?

AppKit/CPTextView/CPTextView.j
```diff
@@ -209,7 +216,7 @@ var kDelegateRespondsTo_textShouldBeginEditing
         [self setEditable:YES];
         [self setSelectable:YES];
         [self setRichText:NO];
-        [self setBackgroundColor:[CPColor whiteColor]];
+        [self setBackgroundColor:[self currentValueForThemeAttribute:@"background-color"]];
```
daboe01, line 219: shouldn't this be?
```objj
[self setBackgroundColor:[CPColor textBackgroundColor]];
```

AppKit/CPTextView/CPTextView.j
```diff
@@ -1129,7 +1138,7 @@ Sets the selection to a range of characters in response to user action.
         dragPlaceholder = [[CPTextView alloc] initWithFrame:placeholderFrame];
         [dragPlaceholder._textStorage replaceCharactersInRange:CPMakeRange(0, 0) withAttributedString:placeholderString];
 
-        [dragPlaceholder setBackgroundColor:[CPColor colorWithRed:1 green:1 blue:1 alpha:0]];
+        [dragPlaceholder setBackgroundColor:[self currentValueForThemeAttribute:@"background-color"]];
```
daboe01, line 1141: shouldn't this be?
```objj
[self setBackgroundColor:[CPColor textBackgroundColor]];
```

PR #2879, opened 2019-12-16
Fixed: Update Tracking areas only once when doing setFrame: on a view

mrcarlberg:
This speed improvement is very notable when resizing a browser window of altering the size of
split views.
The fix uses a variable on the window to control when to update the tracking areas.
If the view is not connected to a window the tracking areas will not be updated as they will
be when added to a window.

### DIFF
--- a/AppKit/CPView.j
+++ b/AppKit/CPView.j
@@ -253,7 +253,6 @@ var CPViewHighDPIDrawingEnabled = YES;
     CPAppearance        _effectiveAppearance;
 
     CPMutableArray      _trackingAreas          @accessors(getter=trackingAreas, copy);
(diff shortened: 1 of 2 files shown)


cacaodev:
There is another possible optimization when we need to traverse the view hierarchy up->down and run something on some views only: you set a flag recursively from bottom to top on the dirty view (here when the tracking area is installed). That way you paint a path that will be used exclusively when updating from top to bottom. This would need a separate PR of course.

mrcarlberg:
@cacaodev True, that is also a good way to speed up things. But that is outside of the scope for the pull request. If needed we can add this at some point in the future.

mrcarlberg:
Merged! Thanks!

PR #2880, opened 2019-12-17
Fixed: CPAppearance has better performance now when views needs layout

mrcarlberg:
This makes updating views much faster when they need layout. This will make resizing browser
window and changing split views sizes much smoother.

The effectiveAppearance is inherited from the superview. The superviews appearance is passed
along as it is recomputed down the view hierarchy.
Also the current appearance is saved on each view to make it possible to only change the theme
states when the appearance is changed.


mrcarlberg:
Merged! Thanks!

PR #2886, opened 2020-02-05
fixed: _glyphsFrames and _glyphsOffsets could be used uninitialized

daboe01:
previously, CPTextView could crash under certain esoteric conditions from accessing properties of undefined.
this is fixed by this PR


mrcarlberg:
Merged! Thanks!

PR #2887, opened 2020-02-09
FIXED: Before this PR, a click in a not first key responder CPTextField would misplace the caret if text alignment was not CPLeftTextAlignment

didierkorthoudt:
This PR adapts the calculation of where in the textfield text the mouse is clicked to take in account the text alignment.

Before the PR:
CPTextFieldClickBefore

After the PR:
CPTextFieldClickAfter


mrcarlberg:
Merged! Thanks!

PR #2888, opened 2020-02-25
FIX: Before this PR, the method attributeBySettingValue ignored existing values

didierkorthoudt:
This PR simply uses `- (_CPThemeAttribute)attributeBySettingValue:(id)aValue forState:(ThemeState)aState` (correctly handling existing values) with CPThemeStateNormal.


mrcarlberg:
Merged! Thanks!

PR #2889, opened 2020-03-01
FIXED: Before this PR, CPButton never declares that its label should be single line

didierkorthoudt:
This PR simply uses the _usesSingleLineMode property of _CPImageAndTextView.


mrcarlberg:
Merged! Thanks!

PR #2890, opened 2020-03-03
Fixed: Removed use of global variables, undeclared types

mrcarlberg:
and other things that can trigger compiler warnings.

Some unused code is also removed.


mrcarlberg:
This pull request is broken. Will create a new one!

PR #2891, opened 2020-03-08
Fixed: JavaScriptCore (WebKit) has a limit at 65536 on number of arguments in a function.

mrcarlberg:
The compiler will crash for really large Objective-J files as the array with code segments will hit this limit.
The solution is to append the array item by item.

I do not find any change on performance with this new implementation.

### DIFF
--- a/Objective-J/ObjJAcornCompiler.js
+++ b/Objective-J/ObjJAcornCompiler.js
@@ -472,7 +472,7 @@ StringBuffer.prototype.isEmptySourceNode = function()
 StringBuffer.prototype.appendStringBufferString = function(stringBuffer)
 {
     // We can't do 'this.atoms.push.apply(this.atoms, stringBuffer.atoms);' as JavaScriptCore (WebKit) has a limit on number of arguments at 65536.
-    // Other browsers also has simular limits.
+    // Other browsers also have simular limits.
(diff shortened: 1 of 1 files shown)


Objective-J/ObjJAcornCompiler.js
```diff
@@ -471,7 +471,18 @@ StringBuffer.prototype.isEmptySourceNode = function()
 
 StringBuffer.prototype.appendStringBufferString = function(stringBuffer)
 {
-    this.atoms.push.apply(this.atoms, stringBuffer.atoms);
+    // We can't do 'this.atoms.push.apply(this.atoms, stringBuffer.atoms);' as JavaScriptCore (WebKit) has a limit on number of arguments at 65536.
+    // Other browsers also has simular limits.
```
daboe01: has->have

daboe01:
-#new
+Tools

Objective-J/ObjJAcornCompiler.js
mrcarlberg: Thanks for finding this!

timhowell01:
This fixes the problem with compiling large classes for me.

mrcarlberg:
Merged! Thanks!

PR #2892, opened 2020-03-10
Fixed: Loading cached theme attributes will miss previous cached value.

mrcarlberg:
Loading cached theme attributes will miss previous cached value if superclass is done before subclass

### DIFF
--- a/AppKit/_CPObject+Theme.j
+++ b/AppKit/_CPObject+Theme.j
@@ -205,7 +205,7 @@ var NULL_THEME = {};
 
         if (cachedAttributes)
         {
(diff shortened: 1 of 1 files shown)


mrcarlberg:
Can someone look at this one as it feels weird that this has gone unnoticed for so long. Maybe I'm missing something?

didierkorthoudt:
I can have a look (and I wonder if this could explain some weird things I've seen…).

daboe01:
-#new
+AppKit

didierkorthoudt:
Seems OK for me…

mrcarlberg:
Merged! Thanks!

PR #2894, opened 2020-03-22
fixed: selection and backspace are recently broken in CPTokenField

### LINKED ISSUE #2795: Tokenfield selection and editing is broken
you can see this in the manual test. deleting highlighted tokens via backspace leaves the page in firefox, fails to delete token in chrome. you cannot place the caret between tokens via mouse (firefox+chrome)  + arrow keys do not work in firefox.

daboe01:
this bug was introduced by commit 26aab29
 26aab29 was meant to improve iOS experience but makes the tokenfield loose FR status prematurely.
this PR helps CPTokenField to regain FR status after bluring the inputfield by clicking on a token.
it secondly compensates for the backspace event not being propagated anymore in all circumstances
fixes #2795

### DIFF
--- a/AppKit/CPTokenField.j
+++ b/AppKit/CPTokenField.j
@@ -584,6 +584,12 @@ CPTokenFieldDeleteButtonType     = 1;
         // Snap to the token if it's only half visible due to mouse wheel scrolling.
         _shouldScrollTo = aToken;
     }
+
+    // this is a hack to compensate for a recent adoption in FR management as introduced by commit #26aab29
+    // this PR makes the tokenfield loose FR status prematurely, so we have to regain here in order to make deleteForward: and friends work
+    setTimeout(function(){
+        [[self window] makeFirstResponder:self];
+    }, 50);
 }
 
 // ===========
@@ -964,8 +970,8 @@ CPTokenFieldDeleteButtonType     = 1;
                 [self _selectToken:tokenView byExtendingSelection:NO];
             }
         }
-        else
-            [self _removeSelectedTokens:nil];
+        // we have to remove unconditionally because the backspace is not propagated anymore starting from commit #26aab29
+        [self _removeSelectedTokens:nil];
     }
     else
     {


daboe01:
-#new
+AppKit
+#needs-review
+regression

daboe01:
works with FF/chrome/safari, all tested on mac

daboe01:
found something: clicking between two tokens does not give us a cursor as it should.

daboe01:
+#needs-improvement

daboe01:
i just tested an old version of cappuccino and it turns out that clicking between two tokens never worked.

-#needs-improvement

daboe01:
@mrcarlberg would you prefer a runloop approach (not tested though)?

```ObjJ
[[CPRunLoop currentRunLoop] performSelector:@selector(makeFirstResponder:) target:[self window] argument: nil order:0 modes:[CPDefaultRunLoopMode]]
```

daboe01:
> The fix referenced above treats the root of the problem which is that the field should not resign at all its fr when a token element is selected.
> TODO: determine what the didBlur listener on the input element was supposed to do and if its removal creates any regression.

i think, that @aljungberg only fixed usability of the touch keyboard of the iPad with 26aab29, that is stopping it from reappearing all the time. I believe he did not test this fix with tokenfield. Therefore, i do not worry so much about the didBlur listener.

cacaodev:
I should have mentioned that in my patch I did remove the didBlur listener on the tokenfield input element, otherwise it sets again the fr on the tokenfield. But it may have another functionality that i did not figure yet by reading the code.

daboe01:
-#needs-review
+#needs-improvement

daboe01:
@cacaodev i advocate for fixing such a fundamental app-breaking high-level issue asap. we can try a better fix in the future. any objections against merging?

enquora:
@cacaodev seems to be no longer active.
@daboe01 I have no objection to merging if you believe this is at least a partial improvement and doesn't cause other immediately identifiable problems.

PR #2903, opened 2020-07-02
removed: Flash support

daboe01:
flash does not play anymore by default in current browsers.
this PR removes flash support from cappuccino.


daboe01:
-#new
+AppKit
+#needs-review

daboe01:
@mrcarlberg flash is removed from the major browsers. can we move this forward?

PR #2906, opened 2020-08-22
FIXED: Before this PR, nib2cib will always set buttons image position to left

didierkorthoudt:
This PR fixes mask values used to determine image position.

It seems that Apple changed those values in the past, at least before Xcode 10.1.

Before this PR : 
BeforePR

After this PR : 
AfterPR


mrcarlberg:
Merged! Thanks!

PR #2907, opened 2020-08-24
TEST: Creating a PR on aristo3 branch

didierkorthoudt:
This is only a test


didierkorthoudt:
@mrcarlberg Well, I think this is the way to do. BUT Travis didn't run (results are those from the real PR #2906). But this is maybe the normal way…

mrcarlberg:
Yes, travis is only setup to run on the master branch

didierkorthoudt:
@mrcarlberg Oh... Is this a problem ? I suppose I just have to ```jake test``` before each PR ? What do you think ?
I'm searching the best way to separate things into coherent PR's…

mrcarlberg:
I think that everything you can should be merged into the master branch. If it is not possible it should be merged into this aristo3 branch. The best thing should be if Aristo3 can exist along with Aristo2 and we don't need this branch at all.

didierkorthoudt:
Well, it should be compatible (I've tried to do so). I'll recheck everything before beginning to create PR's.
I close this test right now as I now know how to push a PR on this branch (and hope we won't need it).

PR #2908, opened 2020-08-25
FIXED: Before this PR, the classname was not specified in the correct DOM element

didierkorthoudt:
The classname for CSS theming was not specified in the correct DOM element.


didierkorthoudt:
+FIFO

didierkorthoudt:
+FIFO

mrcarlberg:
Is this not a subset of #2909?

mrcarlberg:
Merged! Thanks!

PR #2909, opened 2020-08-25
NEW: Material icon image

didierkorthoudt:
Introduction of material icon image (_CPMaterialIconImage)


didierkorthoudt:
+FIFO

didierkorthoudt:
I need to complete this one with specific Material Icons font.

didierkorthoudt:
+FIFO

AppKit/CPImage.j
```diff
@@ -85,7 +88,7 @@ function CPImageInBundle()
 
     if (typeof(arguments[1]) === "number")
     {
-        if (arguments[1] != nil)
+        if (arguments[1] !== nil && arguments[1] !== undefined)
```
mrcarlberg, line 91: Please use `arguments[1] != nil` instead of `arguments[1] !== nil && arguments[1] !== undefined` as it is equivalent and much more clean looking

AppKit/CPImage.j
```diff
@@ -161,7 +164,7 @@ function CPAppKitImage(aFilename, aSize)
 - (id)initByReferencingFile:(CPString)aFilename size:(CGSize)aSize
 {
     // Quietly return nil like in Cocoa, rather than crashing later.
-    if (aFilename == nil)
+    if (aFilename === undefined || aFilename === nil)
```
mrcarlberg, line 167: Please use `aFilename == nil` instead of `aFilename === undefined || aFilename === nil` as it is equivalent and much more clean looking

mrcarlberg (changes requested):

AppKit/CPImage.j
mrcarlberg: And I guess we could remove this if statement altogether as `typeof(arguments[1]) === "number")`  will imply that arguments[1]` can't be `nil` or `undefined`.

mrcarlberg (approved):

mrcarlberg:
Merged! Thanks!

PR #2910, opened 2020-08-25
FIXED: Before this PR, theme color values with alpha component weren't correctly compiled

didierkorthoudt:
Fixes a theme compilation problem because of lack of rgba capability of the compiler.


mrcarlberg:
Merged! Thanks!

PR #2913, opened 2020-08-25
FIXED: Before this PR, when using a custom view with subviews, subviews were presented in the reversed order

didierkorthoudt:
Workaround for the ibtool bug which reverses subviews order when using a custom view.
This is valid as of ibtool version 14460.31


mrcarlberg:
As this is a trick to fix a bug in `ibtools` why not just do `_subviews.reverse();` instead of this manual reverse?

Also `setSubviews:` does a lot of things and is that necessary here?

didierkorthoudt:
@mrcarlberg OK, I've implemented your suggestion and it works without needing ```setSubview:``` & Co. as ```nib2cib``` doesn't cope with the DOM. I've also added a manual test to verify that it works as expected.

Before the fix : 
Bad

After the fix : 
Good

mrcarlberg:
Merged! Thanks!

PR #2917, opened 2020-08-28
FIXED: This PR adapts views and custom views decoding to fix "not ready view" bugs.

didierkorthoudt:
This is a (working) attempt to fix those issues.

### DIFF
--- a/AppKit/CPView.j
+++ b/AppKit/CPView.j
@@ -3768,16 +3768,21 @@ var CPViewAutoresizingMaskKey       = @"CPViewAutoresizingMask",
     // Also decode these "early".
     _frame = [aCoder decodeRectForKey:CPViewFrameKey];
     _bounds = [aCoder decodeRectForKey:CPViewBoundsKey];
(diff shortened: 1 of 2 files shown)


didierkorthoudt:
+FIFO

mrcarlberg:
I guess the problem that is being fixed here is the good old problem we have had for many many years? The question is if this is a final solution for this problem or is it as every earlier fixes just a partly fix? That means that someone else will have a nib file that will not work with this solution of the problem?

didierkorthoudt:
Well, indeed this is not a structural solution, this is just a fix on the current system. It works with all my tests but I can't say it resolves all potential problems. This was constructed by observing things at runtime.

A more global approach would be to rewrite the whole CIB loading system and, in fact, I think we could completely bypass the XIB conversion and load directly the XIB and interpret it. This would make Cappuccino more close to Cocoa than today.

I've noted this on my (looooong) list of things to think of after Aristo3…

mrcarlberg:
Merged! Thanks!

PR #2918, opened 2020-08-28
NEW: This PR introduces the ability to follow host OS UX-UI

didierkorthoudt:
Please refer to the wiki page : 
[https://github.com/cappuccino/cappuccino/wiki/Following-host-OS-UX-UI](https://github.com/cappuccino/cappuccino/wiki/Following-host-OS-UX-UI)


mrcarlberg:
This looks really good!

One question, why is the variable `CPApplicationSelectedOSBehavior` and `CPApplicationShouldMimicWindows` global variables? Is it not better if they are something that is managed by set and get class methods on CPApplication?

didierkorthoudt:
Well, in a first implementation, that was the case but as they may be heavily used in low level rendering methods, I thought it would be better to use global variables, like CPApp for example.

But indeed, this could introduce sides effects if the global variables are directly modified without using the ```setOSBehavior``` method.

What do you prefer ? We leave like this or I add getters and setters and keep the variables private ?

mrcarlberg:
It is much better to have getter and setter for a private variable as you say there might be side effects if you set the global variable directly.

didierkorthoudt:
OK, I'll adapt this.

didierkorthoudt:
I've followed @mrcarlberg suggestion to remove usage of global variables.
Aristo2 retro-compatibility is also better handled.
The wiki page will be adapted to reflect new behavior.

didierkorthoudt:
Wiki page adapted.

daboe01:
-#new
+Theme

daboe01:
moved to #3038
+#duplicate

PR #2923, opened 2020-08-29
NEW: This PR adds a new theme parameter "menu-offset" to CPPopUpButton instead of hard coded value

didierkorthoudt:
Adds a new theme parameter ```menu-offset``` instead of hard coded value.

### DIFF
--- a/AppKit/CPPopUpButton.j
+++ b/AppKit/CPPopUpButton.j
@@ -45,6 +45,13 @@ CPPopUpButtonStatePullsDown = CPThemeState("pulls-down");
     return "popup-button";
 }
 
++ (CPDictionary)themeAttributes
+{
+    return @{
+             @"menu-offset": CGSizeMake(0, -1)
+             };
+}
+
 + (CPSet)keyPathsForValuesAffectingSelectedIndex
 {
     return [CPSet setWithObject:@"objectValue"];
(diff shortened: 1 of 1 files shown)


AppKit/CPPopUpButton.j
```diff
@@ -685,14 +692,16 @@ CPPopUpButtonStatePullsDown = CPThemeState("pulls-down");
...
+            location       = CGPointMake(menuOffset.width, CGRectGetMaxY(bounds) + menuOffset.height);
     }
     else
     {
         var contentRect = [self contentRectForBounds:bounds],
             positionedItem = [self selectedItem],
             standardLeftMargin = [_CPMenuWindow _standardLeftMargin] + [_CPMenuItemStandardView _standardLeftMargin],
-            location = CGPointMake(CGRectGetMinX(contentRect) - standardLeftMargin, 0.0);
+            menuOffset = [self currentValueForThemeAttribute:@"menu-offset"],
+            location = CGPointMake(CGRectGetMinX(contentRect) - standardLeftMargin + menuOffset.width, 0.0 + menuOffset.height);
```
mrcarlberg, line 704: If I understand this correctly so will this put the location one pixel to the left compared to the previous behavior? Is this to correct something a position of something? Just so we understand why this is done.

didierkorthoudt: Well, indeed the default value of ```menu-offset``` should be (0, 0).
And also, this ```0.0 + ...``` is hurting my eyes !
;-)

BTW, just to answer your question, with the previous default value, it would put the location one pixel upper (as default value was (0, -1))

mrcarlberg:
Merged! Thanks!

PR #2928, opened 2020-08-29
NEW: This PR adds the ability to control CPPopoverWindow background color.

didierkorthoudt:
This introduces a Cappuccino specific behavior that permits to control the background color of the popover by setting a background color on the content view.

One can, using this, create "post-it" like yellow popovers : 

- Without doing anything (usual situation) : 

before

- After setting content view background to rgb 254, 246, 156 : 

after

- Without this PR (after setting content view background) : 

without


mrcarlberg:
Merged! Thanks!

PR #2930, opened 2020-08-29
FIXED: This PR specifies correct / complete theme states for themed values

didierkorthoudt:
This PR specifies correct / complete theme states for themed values `bezel-color-up-button` and `bezel-color-down-button`.


mrcarlberg:
Merged! Thanks!

PR #2941, opened 2020-08-31
FIXED: This PR supports modern Cocoa behavior regarding radio buttons groups.

### LINKED ISSUE #2901: Interface Builder No Longer Supports Configuration of Radio Button Groups
This has been the case for some time - `NSMatrix` was once provided in the Interface Builder controls palette for configuring groups of radio buttons in which selection of only one should be possible.

https://developer.apple.com/documentation/appkit/nsmatrix?language=objc

The current Interface Builder mechanism relies on buttons having the same containing view and sharing an IBAction. Beyond those two requirements, everything is automatic.  I presume this is accomplished through run-time manipulation, and we would need to do the same if we want to re-enable direct IB support for groups of radio buttons. It is certainly a nuisance having to wire up a `CPRadioGroup` in a controller for views which can be otherwise be completely defined in IB.

Is creating an IBDesignable `CPRadioButtonGroup` for Xcode another possible fix?

Note also that the `NSRadioButton` style is now deprecated in favour of `NSSwitch`, presenting a further future obstacle for our IB support of mutually-exclusive option controls.
https://developer.apple.com/documentation/appkit/nsradiobutton
I suspect this is driven by Catalyst.

didierkorthoudt:
The current Interface Builder mechanism relies on buttons having the same containing view and sharing an IBAction. Beyond those two requirements, everything is automatic.

This PR fixes #2901

### DIFF
--- a/AppKit/CPRadio.j
+++ b/AppKit/CPRadio.j
@@ -163,15 +163,19 @@ CPRadioImageOffset = 4.0;
 - (void)awakeFromCib
 {
     // Implementation of modern Cocoa behavior : automatic radio group
-    // Search for first CPRadio subview of my superview having an action set.
+    // Search for first CPRadio subview of my superview having the same action as me (if any).
     // If this is me, do nothing.
-    // If this is another radio button and if we have the same action, set my radio group to its, so we'll live in the same radio group.
+    // If this is another radio button, set my radio group to its, so we'll live in the same radio group.
+
+    if (![self action])
+        // I have no action set, so there can be no grouping
+        return;
 
     for (var i = 0, superviewSubviews = [[self superview] subviews], count = [superviewSubviews count], firstRadioButton; (i < count) && !firstRadioButton; i++)
-        if ([superviewSubviews[i] isKindOfClass:CPRadio] && [superviewSubviews[i] action])
+        if ([superviewSubviews[i] isKindOfClass:CPRadio] && ([superviewSubviews[i] action] === [self action]))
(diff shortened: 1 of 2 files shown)


didierkorthoudt:
Oops ! This will only work with Xcode IB. I'll push another commit for code created radio buttons.

didierkorthoudt:
OK. This version works now for radio buttons create with Xcode IB and by code. It takes care of radio manipulations (adding, removing, changing action, ...). 

Should be good (all my tests are working as expected).

enquora:
Works as expected with no apparent edge cases.

Inline documentation should be in a neutral third-person voice rather than first-person and conversational.

An explanation of functionality should be included in the inline documentation in a form that the api documentation generator will recognize. This behaviour isn't well documented in the native Cocoa variant and it isn't necessarily easy to determine the correct behaviour using a search engine.

Is it worth adding a manual test? It's how I verified the functionality.

didierkorthoudt:
This commit adds unit test ```testAutomaticRadioGroup``` in ```CPButtonTest``` that checks correct behavior with automatic grouping of radio buttons.

The commit also provide a fix for a running condition where changing a radio button action after grouping will leave the button in the radio group.

Finally, comments are adapted / completed.

enquora:
@mrcarlberg This looks ready to merge into master.

mrcarlberg:
This works for me. Should be a nice with a manual test for this also but I'm going to accept this as we have a nice unit test that is showing how it works.

PR #2920, opened 2020-08-29
Aristo3: CPButton

didierkorthoudt:
Updates to CPButton & NSButton (nib2cib) :
- all styles of buttons are now managed
- buttons can now use material icon images (with automatic color management)
- updated state management (highlight & showState)
- hovered specific image
- Aristo3 CSS theme ready

PR #2943, opened 2020-10-04
FIXED: Before this PR, CPButton's won't display their title

didierkorthoudt:
With #2920, CPButton's now set their font using [self font]. The corresponding adaptation in CPControl was missing in the previous PR.


mrcarlberg:
Merged! Thanks!

PR #2944, opened 2020-10-06
Update NibApplication template's xib and cib to modern format

enquora:
Any recent version of Xcode will complain that the original xib format is unsafe and force and upgrade when opening.
The MainMenu.xib is upgraded from the Xcode 3.2 format to a more modern one.
The corresponding MainMenu.cib is regenerated using nib2cib.

Xcode 9 and macOS 10.10 are used as the new baseline (minimum supported) platform requirements.


mrcarlberg:
This looks good. I want to wait to merge this until the Aristo 3 support for menus are merged in.

mrcarlberg:
There are some changes merged into master now that alters how the cib file is generated by nib2cib. It should be that the cib file will look different now with the new master.
Can you please check if that is true and commit the new cib file if that is the case.

enquora:
This is the argument against committing read-only generated artifacts to a repository.
Is there a scenario where the installation process won't generate the cib (giving assurance the format is current)? The Starter package?

Do you still prefer including the cib?

enquora:
I will need to move this to its own branch to accommodate GitHub's PR workflow.

mrcarlberg:
I agree on committing read-only and generated artifacts to a repository is not optimal but as this is a template that should work out of the box when you generate a new application we need to include it.

mrcarlberg:
@enquora What do you want to do with this? Do I understand you correctly that you want to move some parts to a separate pull request?

enquora:
I'm closing this and have issued a new PR from its own branch.

PR #2947, opened 2020-10-25
FIXED: default bezel style is now set to CPRoundRectBezelStyle

didierkorthoudt:
This PR :
- changes default bezel style to CPRoundRectBezelStyle
- removes unwanted comments

PR #2948, opened 2020-10-29
FIXED: Before this PR, CPStepper wouldn't display correctly

didierkorthoudt:
This bug was introduced by recent CPButton changes (#2920 and #2947).

This PR simply specifies correct CPButton bezel style to be used by steppers.


mrcarlberg:
Merged! Thanks!

PR #2949, opened 2020-10-31
FIXED: Before this PR, placing a CPTableView in a CPBox would break nib2cib

didierkorthoudt:
This PR simply adds a missing initialization method call to ```NS_initWithCoder``` within ```NSTableView.j```.

It is unclear why this didn't occur before…


mrcarlberg (approved):

mrcarlberg:
Merged! Thanks!

PR #2952, opened 2020-11-07
FIXED : This PR fixes a running condition that could occur in CPRadio

didierkorthoudt:
This PR fixes a running condition where if the value is set before the radio button is associated in a radio group, the radio group will miss the currently selected button.


mrcarlberg (approved):

mrcarlberg:
Merged! Thanks!

PR #2954, opened 2020-11-14
Fix version check for XcodeCapp compatibility with Big Sur

enquora:
Fixed - compatibility check for building XcodeCapp depends on outmoded macOS version numbering scheme.

### DIFF
--- a/Tools/XcodeCapp/Jakefile
+++ b/Tools/XcodeCapp/Jakefile
@@ -23,7 +23,7 @@ task ("build", function()
                 majorVersion = parseInt(versions[0], 10),
                 minorVersion = parseInt(versions[1], 10),
                 buildVersion = parseInt(versions[2], 10);
-				
+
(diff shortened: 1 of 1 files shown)


mrcarlberg:
Thanks for this. Can you please remove the extra whitespaces at the end of one of the lines.

Personally I have turned on auto trim whitespace on all my code editors. It is a very convenient feature that most editors have.

mrcarlberg:
Merged! Thanks!

cacaodev:
@mrcarlberg @enquora With this commit, the min version for Xcode changed from 10.7 to 10.12.
Is it intentional ? because it's not in the pr description.

enquora:
This was intentional, but the cutoff point was somewhat arbitrary - 10.12 represents a convenient dividing line between deprecated and mostly modern Cocoa (for both XcodeCapp and APIs).

The project should ideally have a global statement about support cutoff.

PR #2959, opened 2020-12-12
FIXED: Before this PR, outlet connection decoding was done before views are ready

didierkorthoudt:
This PR simply moves the connections decoding at the end of the initWithCoder: process.
It does it also in the nib2cib process.


enquora:
Initial testing indicates this works as expected.

daboe01:
-#new
+AppKit
+#ready-to-commit

mrcarlberg:
Merged! Thanks!

PR #2962, opened 2020-12-16
Fixed: CPTextView now handles more options from Interface Builder

mrcarlberg:
The new options are:
1. Resizable Horizontally
2. Resizable Vertically
3. Max Size

Also two options from the Text Container is recognized:
1. Tracks text view width
2. Tracks text view height

The last two are needed to allow the text view to recalculate
the size of its content if the frame size is changed.

There are many more options we can add but it is out of the scope for this pull request.


daboe01:
excellent work! thank you.

-#new
+TextView
+#ready-to-commit

PR #2963, opened 2020-12-21
New: Support to compile deep structured project in the web browser

mrcarlberg:
A Cappuccino project has always needed to have a flat file structure to
be able to compile from source in the web browser.

A new dictionary can be added to the project's Info.plist with the key
'CPFileTranslationDictionary'. It should contain all the path
translations in the project to allow the Browser to find out where in
the file structure the source files are. Check in the AppKit framework
for an example.

Also, a include list can be added in the Info.plist for the key
'CPCompileIncludeFileArray'. The files in the list will be included
by a '#include' statement for each file that is compiled in this project

A last new feature is to add macros in the 'OBJJ_COMPILER_FLAGS' global
variable. It is usually declared in the index.html file. It is done
with the format '-Dmacroname[=macrodefinition]'.

An updated Info.plist file in the AppKit framework will allow the AppKit
to compile from the source in the web browser. The following things is
needed to get this to work:

1. Replace the AppKit folder in your project's Frameworks folder with a
copy of the AppKit folder from the Cappuccino project.
2. Add a copy of the built Aristo2.blend folder to the AppKit's Resource
folder.
3. Add "-DPLATFORM_DOM", "-DPLATFORM_BROWSER" and optionally "-DDEBUG"
to the global variable 'OBJJ_COMPILER_FLAGS' in the project's index.html
file.

The app load time will increase (about 3 seconds on a 2018 MacBook Pro)
as the AppKit framework is compiled from source.

The Foundation framework is not yet tested but it might be working by
adding info about the file structure in the Info.plist file.


cacaodev:
Hi @mrcarlberg , this is really great.
Is the Info.plist with the file map mandatory in order to test this ? (e.g. no mixed fmks compiled and source). The console also complains about Info.plist not being found in some AppKit subfolders. Are they also needed ? tx.

mrcarlberg:
@cacaodev Yes, a file map is mandatory for the Foundation framework as it has three subdirectories (CPArray, CPPredicate and CPSet) that contains file that need to be found.

Also, the `Platform/Platform.h` file is needed to be copied from the AppKit into the Foundation framework. The `Platform.h` file has to be included in all file by adding it to the `CPCompileIncludeFileArray` key in the `Info.plist`

It might be possible to get rid of the file not found messages in the console but I have not looked into that yet.

mrcarlberg:
I have now added a file translation dictionary for the Foundation framework too. This allows us to compile both the Foundation and the AppKit frameworks when loading an application in the browser. It is also possible to use any combinations with precompiled or non-compiled versions of the frameworks.

It takes around 4 extra seconds on a 2018 MacBook Pro to compile both frameworks in the browser while loading the applications.

mrcarlberg:
@cacaodev I'm now finished with this. Can you take a look at it?

This pull request is also needed for the coming Node version so it would be nice if it can be merged.

daboe01:
-#new
+Tools
+Objective-J

cacaodev:
Sorry for the late response. I will have a look at it this week-end.

cacaodev:
This merged&reverted PR needs to be reopened.

PR #2965, opened 2020-12-28
FIXED: Before this PR, imagesize would provide image size in points (that is, taking care of the image resolution) instead of pixels

didierkorthoudt:
This PR simply uses ```NSImageRep``` ```pixelsWide``` and ```pixelsHigh``` methods to get correct sizes.

Please note that this PR only impacts CIB based applications as for code created images, size is either specified manually, either determined using the HTML IMG object ```width``` and ```height``` attributes, so in pixels size.


daboe01:
-#new
+Tools
+#ready-to-commit

PR #2968, opened 2021-01-17
fixed: sharedApplication was instantiating the wrong class in subclas…

daboe01:
…ses of CPApplication

previously, CPApplication was hard-coded.
this PR makes sharedApplication instantiating the CPApp of the correct class


mrcarlberg (approved):
Good catch! I will merge this.

mrcarlberg:
Thanks! Merged!

PR #2973, opened 2021-03-14
Fix Java version check to cleanly identify both Oracle and OpenJDK va…

enquora:
…riants

Set minimum supported version to 8 to simplify this check.
Java 8 is the last variant from Oracle to be free and unencumbered, and is classified LTS.
It is also available from OpenJDK.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -130,17 +130,14 @@ function check_install_environment () {
         exit 1
     fi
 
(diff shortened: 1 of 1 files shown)


ThrudTheBarbarian:
This worked for me, using Ubuntu 20.04 and `sudo apt install default-jre`

mrcarlberg:
Merged! Thanks!

PR #2975, opened 2021-03-25
fixed: width of text sizing was slightly off in chrome

daboe01:
see bug https://bugs.chromium.org/p/chromium/issues/detail?id=1192834.
This bug induces white gaps after heavy editing in CPTextViews using chrome/canary/chromium.
This PR makes cappuccino test whether canvas text sizing is functional.
If not so, cappuccino falls back to dom-based sizing

### DIFF
--- a/AppKit/CPSegmentedControl.j
+++ b/AppKit/CPSegmentedControl.j
@@ -802,7 +802,8 @@ CPSegmentSwitchTrackingMomentary = 2;
             label = [segment label],
             image = [segment image];
 
-        width = (label ? [label sizeWithFont:[self font]].width : 4.0) + (image ? [image size].width : 0) + contentInsetWidth;
+        // add 1 pixel to account for possible fractional pixels at right edge
+        width = (label ? [label sizeWithFont:[self font]].width + 1 : 4.0) + (image ? [image size].width : 0) + contentInsetWidth;
     }
 
     return CGRectMake(left, top, width, height);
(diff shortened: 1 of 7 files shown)


daboe01:
+Chrome
+TextView

daboe01:
-#new

daboe01:
i tried putting this in CPCompatibility.j but this turned out not feasible because cappuccino is not sufficiently initialized at this time.

daboe01:
+#needs-review
@mrcarlberg can you have a look, please?

daboe01:
unfortunately, this PR comes with a moderate performance penalty in chrome.

daboe01:
in case why you are wondering why segmented control and tooltips need to be adjusted: previously, the dom-sizing function added one pixel to the real size. this PR makes [CPPlatformString sizeOfString:withFont:forWidth:] return the correct size and adjusts downstream for edge cases from fractional pixels

mrcarlberg:
How much of a performance issues is this? As the sizes are cached I guess it will mostly be when editing texts in text fields and textviews? Maybe also when showing large texts that are for example fetched from somewhere? Or am I missing something?

daboe01:
@mrcarlberg the slowdown is barely noticeable in my apps for the users. IMHO, this does not count, however. Correctness is more important than speed. If something is broken, just do not use it.

daboe01:
@mrcarlberg can we merge this in, please?
we use it in production for several months without any issues.
this would help me switch to the current master and test the next branch in production.
the fallback to dom-sizing will go away automatically as soon as the issue gets fixed in the future.

daboe01:
-#needs-review
+#ready-to-commit

mrcarlberg:
Merged! Thanks!

PR #2988, opened 2021-07-08
fixed: CPButton's images were lost when becoming disabled

### LINKED ISSUE #2977: CPButton setImage: does not set the image in disabled state
@didierkorthoudt 
IMHO can be easily fixed by adding this line to setImage: 
[self setValue:spinnerImg forThemeAttribute:@"image" inState:CPThemeStateDisabled];
should i make a PR with this?
am i missing something?

daboe01:
fixes #2977

### DIFF
--- a/AppKit/CPButton.j
+++ b/AppKit/CPButton.j
@@ -380,6 +380,8 @@ CPButtonImageOffset   = 3.0;
         _bezelState = CPThemeStateNormal;
 
     [self setValue:anImage forThemeAttribute:@"image" inState:_bezelState];
+    // if we omit this, images will disappear as soon as the button becomes disabled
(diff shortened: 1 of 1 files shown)


daboe01:
+theme

daboe01:
-#new

didierkorthoudt:
Well, I have to look but your proposed fix disables setting another image when the button is disabled…

didierkorthoudt:
Well, your proposed fix is the best solution.
In Cocoa, there's no specific mean to set a specific image for the disabled state.
This could always been achieved programmatically even with your fix.
So, OK for me.

daboe01:
+#ready-to-commit

mrcarlberg:
I agree that this is the best solution as mentioned, there's no specific mean to set a specific image for the disabled state in Cocoa.

mrcarlberg:
Merged! Thanks!

PR #2990, opened 2021-07-27
FIXED: Before this PR, CPButton would always have the same label text size whatever the control size

### LINKED ISSUE #2984: For CPButton, the font size does not changes according to control size
For CPButton, when the control size is set to small or mini the font of the button title remains bold and font size also remain same as that of regular control.  This is happening after i started using **Cappuccino 1.0.0** version. in the early version 0.9.10 this was working as expected.
Attached is the screenshot of it. The example is taken from [https://cappuccino-testbook.5apps.com/#ThemeKitchenSink](url) - CPControlSize.

Bildschirmfoto-2021-06-01-um-22 50 10

didierkorthoudt:
This PR fixes #2984 

To cope with new control size management, we have to add some ```font``` attribute definitions to Aristo2.


daboe01:
-#new
+Theme

mrcarlberg:
Merged! Thanks!

PR #2993, opened 2021-07-29
FIXED: Before this PR, using setTextColor on a CPTextField would not take visual effect before an event

### LINKED ISSUE #2985: CPTextField doesn't accept -setTextColor unless it already has string content
This will work:
`

		_date	= [[CPTextField alloc] initWithFrame:CGRectMake(w-70,0,50,14)];

		[_date setStringValue:"--/--/--"];

		[_date setTextColor:[CPColor colorWithHexString:"888888"]];

		[self addSubview:_date];
`

... but remove the dummy -setStringValue: and the text is always black.

didierkorthoudt:
This PR simply adds a call to ```layoutSubviews``` after setting the text color.

Fixes #2985

### DIFF
--- a/AppKit/CPTextField.j
+++ b/AppKit/CPTextField.j
@@ -2020,6 +2020,8 @@ CPTextFieldStatePlaceholder = CPThemeState("placeholder");
     }
 
     [self setValue:placeholderColor forThemeAttribute:@"text-color" inState:CPTextFieldStatePlaceholder];
+
+    [self layoutSubviews];
 }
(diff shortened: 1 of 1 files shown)


mrcarlberg (approved):
This looks great!

mrcarlberg:
Thanks! Merged!

daboe01:
we also need a [self _setCSSStyleForInputElement];
without this, the text-color will change when this is called from an action method.

Issue #2971, opened 2021-01-28
nib2cib incompletely converts NSImageView instances


enquora:
Instances of NSImageView in Interface Builder are incompletely converted by nib2cib
Instances of CPImageView are correctly converted but image data is not converted when an image is set statically in Interface Builder.
[ImageTest.zip](https://github.com/cappuccino/cappuccino/files/5889437/ImageTest.zip)

cappbot:
**Milestone:** Someday.  **Label:** #new.  **What's next?** A reviewer should examine this issue.

didierkorthoudt:
I’lol have a look at this one as I’m currently working on image views...

enquora:
Note in the logging statement in the AppDelegate - the filepath is present in the CPImage, but nothing else.
It appears the coder simply doesn't deal with possible image data.

cacaodev:
@enquora
Do you know if IB encodes the image in the Nib file for images views ?
Also, can you describe in what situation it would be better to have the image encoded instead of having a reference to the file, and what problem the current behavior is causing.

There is an issue opened for the CPBox resize masks : #2970

enquora:
@cacaodev 
I do not, in fact, know if IB encodes images for Nibs rather than providing references - more research needed.
I've been thinking about this since running into the problem - encoding a reference seems more efficient and the more likely approach. Encoding the image data each time it is used (when used multiple times) doesn't seem like a good approach.

More research and testing needed. For our specific immediate need, I'm simply loading and setting them manually. A nuisance, but it works.

enquora:
@daboe01 Can you make a tag for Xcode and nib2cib issues? I'd like to start grouping them together. IB is one of our major advantages (rapid app development) and I'd like to see reliability improvements in our next release (presumably with Aristo3). 

I'm in the middle of a major rollout at the moment and can't stop to fix things like this, but will come back to them once settled. wrt Xcode, its editor is now a very good option, especially if we can get an LSP server going for Objective-J to take advantage of refactoring features.

I now have a functional xclangspec for Objective-J and am improving it incrementally.
The installer will be rolled into Xcode-Capp as part of my general cleanup of it.
I've identified a likely culprit for nib2cib's random (false) failures on Big Sur and will soon file a placeholder issue for that.
nib2cib and IDE issues seem to be related, so a single tag seems appropriate.

daboe01:
@enquora good idea.  this can only be done by @mrcarlberg as i am not a core-dev.

[...]

PR #2995, opened 2021-07-31
FIXED: Before this PR, nib2cib would fail if, for some "not ready" condition, no superview is defined for a view

didierkorthoudt:
This PR checks if there's a defined superview in ```resizeWithOldSuperviewSize```.

The problem can be observed when trying to nib2cib the MainMenu.xib file of the reduction in issue #2971 (ImageTest.zip).


daboe01:
-#new
+Tools
+#ready-to-commit

mrcarlberg:
Merged! Thanks!

PR #2998, opened 2021-08-17
Modernise "document.write" in «index.html» to "innerHTML"

michaelbach:
Replace `document.write(loadingHTML);` (which should be avoided for speed) with equivalent `innerHTML` code. If acceptable, the other templates (Application, Framework, ThemeDescriptor) might be upgraded accordingly.

### DIFF
--- a/Tools/capp/Resources/Templates/Application/index-debug.html
+++ b/Tools/capp/Resources/Templates/Application/index-debug.html
@@ -188,9 +188,9 @@
     <body>
         <div id="cappuccino-body">
             <div class="container">
-                <div class="content">
+                <div class="content" id="loadingDiv">
                     <script type="text/javascript">
-                        document.write(loadingHTML);
+                        document.getElementById("loadingDiv").innerHTML = loadingHTML;
                     </script>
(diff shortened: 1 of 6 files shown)


daboe01:
-#new
+Tools

michaelbach:
> What about index-debug.html?

Not important, because that will not be deployed and thus not seen by Google etc. which tag the current approach as outdated & slower.

mrcarlberg:
That might be true but do we want to use two different methods in the two index files? If we should we need to write some comments why we are doing this differently as it will increase the complexity.

michaelbach:
> That might be true but do we want to use two different methods in the two index files? …

Yes, you have convinced me. So should I just also do an additional pull request on `index-debug.html` or rather delete this and do a new one with both?

mrcarlberg:
Just add the changes to your branch and it will be included in this pull request

cacaodev:
Browser support for innerHTML is 95% of global usage, so this seems safe.

mrcarlberg:
Merged! Thanks!

PR #3001, opened 2021-08-30
Fixed: Bezel state ivar _bezelState must remove current state when changing to a new.

mrcarlberg:
If the bezel state was changed and some attribute was set like an "image". The image would not be found as there were two states active and it would not match.


daboe01:
-#new
+Theme
+#ready-to-commit

daboe01:
looks good to me

mrcarlberg:
Merged! Thanks!

PR #3006, opened 2021-10-04
Fix nib2cib under Xcode 13

enquora:
Backwards-compatible changes to accommodate different ibtool output from Xcode 13 onwards.


mrcarlberg (commented):
Great work!

A couple of things to consider:
1. Use `CPLog.info` or similar instead of `print` or `console.log` to allow the `--verbose` option to work.
2. Is it necessary to print out all environment variables?

enquora:
Updated to use CPLog.info.
Logging environment variables was useful during testing for this fix, both by the author and at least one user who encountered an unexpected problem. Reporting them is moved to verbose mode.

mrcarlberg:
What about using `CPLog.error` when printing the error?

Also, on the environment variables, I beleive it is bad habit to just dumping out stuff in `--verbose` mode. It has to be some relevant information. This is just too much information that can easily be obtain in other ways. Please remove it.

PR #3008, opened 2021-10-23
Include commented-out code to hide mainMenu in NibApplication template

enquora:
Add commented-out code to capp gen NibApplication template.
The technique itself is easy to forget and not obvious to newcomers.
It is already included in the standard capp gen application template
and added here for completeness.


mrcarlberg:
Thanks! Merged!

PR #3014, opened 2021-12-15
Fixed: Correct init of CPTextView when running nib2cib

mrcarlberg:
nib2cib can fail and crash if a CPTextView is in the .xib file as the text view is not initialized correctly


daboe01:
-#new
+TextView
+#ready-to-commit

mrcarlberg:
Merged! Thanks!

Issue #3013, opened 2021-12-11
CSP failure


darkoverlordofdata:
When I try to deploy my project, I get this error:
```
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'".
``` 

I'm trying to use cappuccino in a chrome extension. I've compiled using 
```
jake release
```
so I thought that would resolve the run runtime compilation. Is that a wrong assumption? Is there a way to remove the string evaluation from the final output?

cappbot:
**Milestone:** Someday.  **Label:** #new.  **What's next?** A reviewer should examine this issue.

darkoverlordofdata:
I'm analyzing the output  *.sj  files, and I see eval used quite a bit; I suspect removing them would not be feasible. I can sandbox my extension, and that will work in dev, but I don't like eval much in production. What mitigations are in place that prevent this from being an attack surface?

daboe01:
-#new
+#needs-discussion

darkoverlordofdata:
@daboe01, I'm always concerned about injection attack when I see an eval, but in this case I'm also concerned about google's review procedure, and I'm trying to anticipate roadblocks to passing that review. 

Chrome extensions are not sandboxed by default, though it is an option.

Feedback from the extension dev community has pointed out some other issues, mainly the size of the libraries and clarity of the code in them - google is likely to reject it if it requires too much resource on their end to manually inspect all the code.

Do you know if Cappuccino has been submitted in an extension in the past? That could be leveraged to expedite on their end.

mrcarlberg:
We had/have a feature to deploy a Cappuccino app as a single Javascript file. This is not working in the Node version but might still work in the Narwhal version. You create it with `jake deploy`

daboe01:
@darkoverlordofdata i am not aware of any chrome-extension that has been done with cappuccino. IMHO cappuccino is not well suited for this use-case.

darkoverlordofdata:
Deploy sounds like what I'm looking for. Unfortunately, the only blocking issue I have is that Narwhal doesn't support arm. My last intel machine died last year, all I have now is arm. Looking deeper into Narwhal, I don't have any expectation that it ever will support arm. It looks to me like the project died about 10 years ago, along with 280north. That explains alot.

[...]

PR #3016, opened 2022-01-07
fixed: usage of eval instead of JSON.parse for parsing JSON

daboe01:
using eval for parsing JSON is considered bad practice and is also slower.
addresses some of the concerns raised in #3013


daboe01:
-#new
+AppKit

mrcarlberg:
Merged! Thanks!

PR #3020, opened 2022-03-01
fixed: key event issues of handling escape in recent firefox …

daboe01:
…versions
recently FF returns "Escape"  instead of chr(27) in the key property of keyboard events. This caused the esc key to malfunction in cappuccino on FF.

This PR fixes this by treating esc analogous to webkit. Tested on FF 94.0.2 (64-Bit) for mac.


daboe01:
+Firefox
-#new
+AppKit

mrcarlberg:
Merged! Thanks!

Issue #3021, opened 2022-03-10
Cursor is blinking in a non-editable CPTextView


daboe01:
the cursor should not be visible at all.

cappbot:
**Milestone:** Someday.  **Label:** #new.  **What's next?** A reviewer should examine this issue.

daboe01:
+#fixed

cappbot:
**Milestone:** Someday.  **Label:** #fixed.  **What's next?** This issue is considered successfully resolved.

cappbot:
**Milestone:** Someday.  **Label:** #fixed.  **What's next?** This issue is considered successfully resolved.

PR #3022, opened 2022-03-12
fixed: caret was blinking in non-editable text in CPTextView

daboe01:


daboe01:
-#new
+TextView
+#ready-to-commit

daboe01:
fixes #3021

PR #3024, opened 2022-03-17
Revert "New: Support to compile deep structured project in the web browser"

cacaodev:
There is an error at runtime with this commit, let's revert until we find the problem.
Reverts cappuccino/cappuccino#2963


mrcarlberg:
@cacaodev What kind of error do you get?

cacaodev:
@mrcarlberg really sorry for answering so late
After `jake clean && rm -rf $CAPP_BUILD && jake sudo-install`, you can see the error in any Manual test with symlinked frameworks to $CAPP_BUILD, so not using this pr feature.
Without the OBJJ_COMPILER_FLAGS "IncludeDebugSymbols", you can see the error in the console: 
Capture d’écran 2022-03-31 à 10 30 17
"

cacaodev:
@daboe01 thank you, good catch !
Embedded comments like this are causing the error:
`OBJJ_COMPILER_FLAGS = [/*"IncludeDebugSymbols"*/, "IncludeTypeSignatures"/*, "SourceMap"*/, "InlineMsgSend"];`
Now in master I don't see the error with the previous line but there is no reason to think it is related to this PR.

michaelbach:
Then problem is the misplaced comma after */, it should be WITHIN the out-commented part.

PR #3028, opened 2022-03-25
Modernize XcodeCapp jakefile

enquora:
* Properly check for macOS versions which use a major version > 10.
* Clean up minimum/target platform versions - minimum version should now be 10.13 everywhere.
* Implement separate build and install phases.
* Build to Cappuccino source folder
* Install to ~/Applications folder (to avoid problems with elevated permissions in /Applications)
* Ensure Xcode's modern build system is used for all schemes and targets (legacy build system is deprecated and may be removed without further notices)

### DIFF
--- a/Tools/XcodeCapp/Jakefile
+++ b/Tools/XcodeCapp/Jakefile
@@ -1,30 +1,70 @@
 
 require("../../common.jake");
 
-var OS = require("os"),
-    task = require("jake").task,
-    stream = require("narwhal/term").stream,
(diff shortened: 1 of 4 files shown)


daboe01:
-#new
+Tools

enquora:
See comments for #3029 - this is a work-in-progress, especially determining whether narwhal branch be left as-is.

daboe01:
imho can be merged when conflict is resolved.

PR #3029, opened 2022-03-26
Update Python usage for version 3.6+ and make dependency explicit

enquora:
* Python dependency was previously implicit because it was guaranteed to be available on default OS X installations.
* Change `python` references to `python3`.
* Change print statements to print function calls.
* Use f strings for string formatting and interpolation (for clarity).


daboe01:
-#new
+Tools

enquora:
There are multiple things happening with inside this.
This is a work-in-progress, probably requiring either separate patches for node and narwhal, or leaving narwhal as-is.
Xcode integration is being finalized for node branch now - raise questions on Gitter.

enquora:
Conversion of Python 2 code used in XcodeCapp to Python 3 is problematic and is being replaced with native Objective-C code. This will be merged in a separate pull request. This PR kept open for reference until then.

enquora:
Python usage will be replaced by native code in future.
This patch on partially resolves problem - full conversion to Python 3 is too problematic to justify.

PR #3031, opened 2022-03-30
fixed: å and Å did not work well on scandinavian keyboards in CPTextView

### LINKED ISSUE #3030: The key å and Å is lost or misplaced when typing in CPTextView
We have problems with CPTextView and the Scandinavian character Å and å. When typing fast the text view will loose or misplace it. This makes CPTextView unusable for us. There is some special handling with it as browser has some old weird handling for dead keys. ("https://stackoverflow.com/questions/25043934/is-it-ok-to-ignore-keydown-events-with-keycode-229").
I did some quick check in the code and the CPTextView uses some global variables for handling this (´_CPNativeInputFieldKeyDownCalled´ and friends). It does not work when you type fast as multiple keydowns will come in a row and probably the global variables will be overwritten? This can be tested by setting the keyboard layout to a Scandinavian layout and type å or Å (the key directly to the right of P). It does work if you type very slow.

Scandinavian keyboard layout:
image

daboe01:
previously, å and Å had been awkwardly filtered in keyDown: and reinserted via the deadkey handler. this caused a sluggish typing experience for our Skandinavian friends.

this filtering is cleaned up by this PR.
deadkey-diacritics still work.

tested on chrome, safari and FF, although only on mac.
fixes #3030

### DIFF
--- a/AppKit/CPTextView/CPTextView.j
+++ b/AppKit/CPTextView/CPTextView.j
@@ -1051,7 +1051,7 @@ Sets the selection to a range of characters in response to user action.
 
     [[_window platformWindow] _propagateCurrentDOMEvent:YES];  // for the _CPNativeInputManager (necessary at least on FF and chrome)
 
-    if (![_CPNativeInputManager isNativeInputFieldActive])
+    if (![_CPNativeInputManager isNativeInputFieldActive] && ![_CPNativeInputManager isDeadKey:event])
         [self interpretKeyEvents:[event]];
 
     [_caret setPermanentlyVisible:YES];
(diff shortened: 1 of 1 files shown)


daboe01:
sorry, i still see issues. will try to fix asap.

daboe01:
-#new
+TextView

daboe01:
@mrcarlberg works for me now. please test it and merge at your discretion (also into node, please)

daboe01:
+#ready-to-commit

mrcarlberg:
Fixes #3030

mrcarlberg:
Merged! Thanks!

PR #3038, opened 2022-04-01
Aristo3

daboe01:


daboe01:
-#new

daboe01:
👍🏿 I tested nealy every manual test for structural issues + use Aristo3 in production in several non-trivial applications

michaelbach:
Nice work! Found this problem: Comma-separated declaration lists of ivars are not possible. An example, this here: `int i1, i2;` causes the error message: "Unexpected token" and parsing aborts. Placing only one ivar on a line parses fine, but is a restriction compared to Objective-C or JavaScript.

michaelbach:
Modern JavaScript parsing is incomplete: Destructuring assignments with an array pattern are flagged with warnings. Two examples: `const [CI0025, med, CI0975] = quantilesArray;` or `for (let [name, meta] of gNamesAndTypesMap) {…}`. They both result in the warning "**Variable 'undefined' is never read**". Apart from the warnings, the code works correctly.

daboe01:
> Nice work! Found this problem: Comma-separated declaration lists of ivars are not possible. An example, this here: `int i1, i2;` causes the error message: "Unexpected token" and parsing aborts. Placing only one ivar on a line parses fine, but is a restriction compared to Objective-C or JavaScript.

i am pretty sure this is not from aristo3. try with the latest main.

daboe01:
> Modern JavaScript parsing is incomplete: Destructuring assignments with an array pattern are flagged with warnings. Two examples: `const [CI0025, med, CI0975] = quantilesArray;` or `for (let [name, meta] of gNamesAndTypesMap) {…}`. They both result in the warning "**Variable 'undefined' is never read**". Apart from the warnings, the code works correctly.

i am pretty sure this is not from aristo3. try with the latest main.

michaelbach:
> > Nice work! Found this problem: Comma-separated declaration lists of ivars are not possible. An example, this here: `int i1, i2;` causes the error message: "Unexpected token" and parsing aborts. Placing only one ivar on a line parses fine, but is a restriction compared to Objective-C or JavaScript.
> 
> i am pretty sure this is not from aristo3. try with the latest main.

Yes, both my reported issues stem from the changes in `Objective-J.js`. The same problems occur when I switch to Aristo2 in `Info.plist`.

yvess:
👍🏽 works for me, some small layout shifts, but function wise my app works as expected

PR #3045, opened 2022-06-08
fix CPSlider.j

michaelbach:
Fix: slider knob did not follow “setObjectValue” when set to “Only stop at tick marks”. Needed a new method closestTickMarkValueToIndex, called in knobRectForBounds. [Tick marks still are not drawn, but binding to value works now.]

### DIFF
--- a/AppKit/CPSlider.j
+++ b/AppKit/CPSlider.j
@@ -312,7 +312,7 @@ var AFFINITY = 5;
     if (!trackRect || CGRectIsEmpty(trackRect))
         trackRect = bounds;
 
-    _closestTickMarkIndex = [self closestTickMarkValueToIndex: [self doubleValue]];
+    _closestTickMarkIndex = [self closestTickMarkValueToIndex:[self doubleValue]];
     _currentTickMarkSegment = _closestTickMarkIndex;
 
     if (_isCircular)
--- a/AppKit/CPSlider.j
+++ b/AppKit/CPSlider.j
@@ -312,6 +312,9 @@ var AFFINITY = 5;
     if (!trackRect || CGRectIsEmpty(trackRect))
         trackRect = bounds;
 
+    _closestTickMarkIndex = [self closestTickMarkValueToIndex: [self doubleValue]];
+    _currentTickMarkSegment = _closestTickMarkIndex;
+
     if (_isCircular)
     {
         var angle  = 3 * PI_2 - (1.0 - [self doubleValue] - _minValue) / (_maxValue - _minValue) * PI2,
@@ -664,6 +667,32 @@ var AFFINITY = 5;
     }
 
     return _closestTickMarkIndex ? _cachedTickMarksValues[_closestTickMarkIndex] : _minValue;
+}
+
+// Returns the index of the tick mark closest to the specified value,
+// extracted from the function `closestTickMarkValueToValue` (above)
+- (int)closestTickMarkValueToIndex:(double)value
+{
+    if (_numberOfTickMarks === 1)
+        return 0;
+
+    _closestTickMarkIndex = CPNotFound;
+
+    var foundDelta = 2 * _maxValue,
+        delta;
+
+    for (var i = 0; i < _numberOfTickMarks; i++)
+    {
+        delta = ABS(value - _cachedTickMarksValues[i]);
+
+        if (delta < foundDelta)
+        {
+            foundDelta = delta;
+            _closestTickMarkIndex = i;
+        }
+    }
+
+    return (_closestTickMarkIndex != CPNotFound) ? _closestTickMarkIndex : 0;
 }
 
 // Returns the index of the tick mark closest to the location of the receiver represented by the given point.


AppKit/CPSlider.j
```diff
@@ -312,6 +312,9 @@ var AFFINITY = 5;
     if (!trackRect || CGRectIsEmpty(trackRect))
         trackRect = bounds;
 
+    _closestTickMarkIndex = [self closestTickMarkValueToIndex: [self doubleValue]];
```
daboe01, line 315: please remove the space character following closestTickMarkValueToIndex:

daboe01:
+#ready-to-commit
+Bindings
+AppKit

daboe01:
i tend not to merge this one.
closestTickMarkValueToIndex is not part of cocoa.
what is a good use case for this?
at least, the method should be underscored to mark it as a private addition.
-#ready-to-commit

enquora:
I'll take a look at this too. 
We do want to precisely match AppKit APIs (for documentation and reference testing) unless a very good reason exists.

michaelbach:
Dear Daniel:

> i tend not to merge this one.
> closestTickMarkValueToIndex is not part of cocoa.
> what is a good use case for this?
> at least, the method should be underscored to mark it as a private addition.
> -#ready-to-commit

Please do merge. Use case: without it, one cannot make slider values stop at tick marks and communicate via bindings. I use it very often. Example:
https://michaelbach.de/ot/mot-rmi/

`closestTickMarkValueToIndex` was already called in the code, but was massing, so I added it.
Very happy with underscore.


Thanks, Michael
-- 
https://michaelbach.de

enquora:
I'll look at this next week too. If it introduces a non-cocoa API, I would like to consider carefully.
Agree entirely with your rationale, @michaelbach. I also want to introduce more automated and manual tests - we've let this slide a bit over the last two or three years. Restricting ourselves to cocoa-only APIs opens opportunities to automate testing against a native app (not necessarily for this PR, but more complex ones).

daboe01:
> `closestTickMarkValueToIndex` was already called in the code, but was massing, so I added it. Very happy with underscore.

i cannot find any call to closestTickMarkValueToIndex in the current implementation. where did you see this?

daboe01:
if this is only used by your code, it may be better to add this to your your codebase by means of a category of CPSlider.

enquora:
@michaelbach Do you have a reduction for this?
If I understand your description correctly, bindings are not currently working with `CPSlider` when 'Only stop at tick marks' is set - because `setObjectValue` is not fully implemented?

If so, yes, it needs to be fixed. 
We should not implement 'public' methods not present in the canonical API - but if this an implementation problem, we can use any 'private' method(s) necessary. Of course, Objective-C doesn't have 'private' methods - the usual solution is to prefix and underscore.

michaelbach:
My comment re `closestTickMarkValueToIndex` above, namely "was already called in the code" was wrong. But `closestTickMarkValueToIndex` is necessary, because otherwise the slider knob does not follow `setObjectValue` when set to “Only stop at tick marks”. So I am not intending to add something to the API, it's just used internally. So the idea to lead with an underscore is fine. And it DOES NEED to be fixed to achieve standard Cocoa behaviour of `CPSlider`.

enquora:
Any 'normal' method, whether intended for internal usage only or otherwise, will be indexed when running the generator for the API documentation. Internal-only methods need to be implemented so they aren't included in generated documentation. Underscored method names are the standard way to do this, although I'm not familiar enough with the documentation generator to know if there may be other ways to suppress documentation of individual methods.

No argument here about this being a deficiency in the implementation which is needed for Cocoa conformance.

michaelbach:
My added `closestTickMarkValueToValue` is a variation of `closestTickMarkValueToValue`. That is _also_ internal only; I now find that in `CPSlider` there are several internal-only functions w/o a leading underscore. What prevents them from showing up in documentation is their missing "/*! …" structure. Given that so many internal routines are missing the leading underscore, I now think this CPSlider pull request is fine as is. I have been using this version (merging it into the stream before building) for 2 years now in quite a number of projects w/o problems.

enquora:
You're correct, `closestTickMarkValueToValue:` is part of the canonical API, and has been since 10.0.0.
If all indexable methods in use are present in the canonical API, this can move ahead, imo.

I won't be able to perform a sign-off review of the code until the coming week, and would also prefer to see a manual test included with this patch - perhaps I can put something together for you to include.

enquora:
There is an opportunity for improvement here - a script which quickly diffs our API against the canonical one ;-)

enquora:
@daboe01 is correct: closestTickMarkValueToIndex is not part of Cocoa – closestTickMarkValueToValue *is* and has been since OS X 10.0.0

This implementation needs to reflect such.

daboe01:
i just had a second look: your closestTickMarkValueToIndex is exclusively used to update the ivars _closestTickMarkIndex and _currentTickMarkSegment in the method rightTrackRectForBounds:(CGRect)bounds.

is this correct?

if so, why don't we just replace your lines 315-316 with a 'dummy' call to closestTickMarkValueToValue:, just for the side effect to update _closestTickMarkIndex?

would this work?

michaelbach:
@daboe01 This looks perfect to me, congrats & thank you! I am deep in other Cappuccino stuff, so can't test right now, will try to make time.

michaelbach:
Indeed, we only need these two lines, `closestTickMarkValueToIndex` is then superfluous:
`    [self closestTickMarkValueToValue: [self doubleValue]];//only need side effect
    _currentTickMarkSegment = _closestTickMarkIndex;
`
BTW: all this is only necessary when "Only stop on tick marks" is clicked for the slider.

daboe01:
> Indeed, we only need these two lines, making my `closestTickMarkValueToIndex` superfluous: ` [self closestTickMarkValueToValue: [self doubleValue]];//only need side effect` ` _currentTickMarkSegment = _closestTickMarkIndex;` BTW: This code is only reached when "Only stop on tick marks" is clicked for the slider.

it would be perfect, if you could modify this PR accordingly.
please wrap it in if (_allowsTickMarkValuesOnly)... for documentation purposes

daboe01:
i am closing this one as we need a new PR anyway.
(this PR is against the old master.)
@michaelbach would you please be so kind to open a new PR with the suggested fix?

PR #3046, opened 2022-10-17
fixed: dangling link to our website

daboe01:
subject says it all.


daboe01:
the link is not dead anymore.

PR #3053, opened 2023-02-25
fixed: poor user experience when entering dates that are too high for…

daboe01:
… the current month.
E.g. if you enter 30 in Februrary (30/02), "30" will be silently reset to 01/02 after tabbing over to the month field.
This patch makes sure that 30 is retained in the day field.
To achieve this, the month is bumped by one, so 30/02 would automatically get to 30/03 after tabbing over the month field.


daboe01:
-#new
+AppKit

daboe01:
i have used this in production for several months and my users are extremely happy with this.

PR #3056, opened 2023-06-06
Fix nib2cib failing on xibs with deployment target set to recent macOS versions

enquora:
For node branch only.


enquora:
I would like to leave this open for a few more weeks - pending a more complete audit of the toolchain usage under node.
It can be manually applied as a patch (add .patch to the end of the page url) in the interim.
Applicable to both narwhal and node branches.

enquora:
Made redundant by PR 3077 - which handles all current cases, unlike this PR.

Issue #2177, opened 2014-08-20
CPWindow doesn't implement setReleasedWhenClosed: [+1]


Dogild:
When closing a CPWindow with setReleasedWhenClosed: set to YES. We should release all components of the window. It means remove all views form the notification center, deassign every delegate and so on...

daboe01:
+1

cappbot:
**Milestone:** Someday.  **Vote:** 1.  **Labels:** #needs-patch, AppKit, bug.  **What's next?** This issue needs a volunteer to write and submit code to address it.

daboe01:
-#needs-patch

cappbot:
**Milestone:** Someday.  **Vote:** 1.  **Labels:** AppKit, bug.  **What's next?** A reviewer should examine this issue.

PR #3060, opened 2023-12-23
new: support for releasedWhenClosed in CPWindow

daboe01:
fixes: [#2177](https://github.com/cappuccino/cappuccino/issues/2177)

### DIFF
--- a/AppKit/CPView.j
+++ b/AppKit/CPView.j
@@ -44,6 +44,7 @@
 @class CPClipView
 @class CPScrollView
 @class CALayer
+@class CPBinder
 
 @global appkit_tag_dom_elements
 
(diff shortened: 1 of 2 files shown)


daboe01:
@enquora hi david, can you have a look at this one, if time permits?

daboe01:
-#new
+#needs-review
+AppKit

enquora:
I can look at it in detail next week, if you have concerns.
Quick thoughts:

I will need some time to grok if we're covering all possible observers.
You have a more comprehensive understanding of this than I do - if you're comfortable all conditions are covered and my next two comments are not a concern, and want to merge before next week, I'm ok with that.

The view hierarchy is called recursively. Is there risk of a hierarchy so deep it reaches stack limits?
I have a patch adding view controllers to the responder chain, but it walks *up* the view hierarchy rather than down.
Only the most recent Safari supports tail call recursion. It's a work-in-progress for Firefox and Chrome seems uninterested.
I don't know whether this is all moot or of real concern.

I do have an antipathy to `if` statements which use indenting alone and omit braces - prefer seeing explicit block delimiters. Having said that, I know this has seen significant usage throughout the codebase's history. If explicitly part of our style guide, I'll get over it. I do find it an impediment to reliably scanning code though.

PR #3062, opened 2023-12-30
new: blendedColorWithFraction:(CGFloat)fraction ofColor:(CPColor)color

daboe01:
this adds a missing method from the cocoa spec from here: https://developer.apple.com/documentation/appkit/nscolor/1524689-blendedcolorwithfraction


daboe01:
looks like nobody needed this so far.

PR #3064, opened 2024-03-18
Allow building XcodeCapp under any version greater than Sierra

enquora:
Individual new OS releases had support added individually until now.


daboe01:
looks good to me

PR #3072, opened 2024-06-03
Remove narwhal-specific artifacts

enquora:
Remove bootstrap.sh and narwhal.local.conf.
Used by/useful for Narwhal javascript engine - no longer needed.


daboe01:
excellent!
please go ahead :-)

enquora:
A project-wide search shows a surprising number of usages of `require("narwhal/somemodule").
Some are transitional, but considerable testing will be required to determine how they all interact with the current codebase - and to ensure they are not now aliases to something in Node.

We may be better off removing these references in batches, rather than trying to do everything at once.
The two mentioned here are obvious and trivial, for instance.

daboe01:
```bash
find . -type f -exec grep -l 'narwhal' {} +
```

enquora:
I want to confirm these remaining references are, in fact, dead code - rather than being calls to something that was abstracted to Node.

enquora:
I have divided feelings about the proper approach to this: consolidating as many removals as possible in as few commits as possible, vs getting on with a much larger number of smaller, sometimes trivial, commits. I see many more Narwhal references than I at first thought were present.

mrcarlberg:
I agree that we should not remove all references to narwhal as there are still many things that is not converted to Node. To do this in batches is a good approach.

I think removing these two files is a good start so why not merge this in and we can create more pull requests for other changes.

enquora:
> I agree that we should not remove all references to narwhal as there are still many things that is not converted to Node. To do this in batches is a good approach.
> 
> I think removing these two files is a good start so why not merge this in and we can create more pull requests for other changes.

Agree. Addressing all instances in a single squashed commit is not practical.

PR #3074, opened 2024-06-09
README.markdown: remove 'checkout node' as node is now main

daboe01:


enquora:
Is this rendered superfluous with the just-merged general update of the README?

michaelbach:
This is still needed, the `git checkout node` command needs to be removed as Daniel says.

daboe01:
thank you

PR #3077, opened 2024-07-02
Update nib2cib's Converter.j to handle xib files with deployment targ…

enquora:
Update nib2cib to accommodate xibs with deployment versions of Xcode 11 or later.


enquora:
Working, but requires cleanup of style and inline documentation.

PR #3080, opened 2024-07-05
Re-enable installation of capp_lint

enquora:
Requires Python 2, but this is true for XcodeCapp also. Installs to /usr/bin/local, but must be manually installed using install script in source directory.


enquora:
Re-enable installation of capp_lint.
This is an optional installation, which must be performed manually using the install.sh script in the source directory.
Requires Python 2, XcodeCode will not yet use it, even if installed.

PR #3088, opened 2024-07-20
Update README.markdown

michaelbach:
Some headers were missing space after the last #,
Code block in the simple example ran on.


enquora:
Further README refinements in PR 3089, some of which would be overwritten by this one.
It will be the last for awhile.
Let's proof 3089 and commit before reconciling these.

daboe01:
sorry, for not seeing this PR early enough.
i fixed this myself directly.

PR #3089, opened 2024-07-20
Readme refinements

enquora:


enquora:
This needs review - I've been editing using both Xcode and an external Markdown editor - and they each have formatting idiosyncrasies when saving.

PR #3091, opened 2024-08-04
added documentation to the copy paste limitation

### LINKED ISSUE #2983: copy/paste with CPTextView only works with a main menu
you can see this e.g. in the manual test CPTextViewCibTest that does not have a main menu.

this limitation is not present in cocoa and is not obvious.

one solution could be to raise an exception if no main menu is installed.
another solution could be to add this to the documentation.
a third solution could be to always add the main menu on startup but hide it by default.

what do you think?

daboe01:
fixes #2983


enquora (commented):
I'm not current with the documentation generator, but is there a simple tag which can include this comment/caveat in generated API docs?

enquora:
This seems ready to commit

PR #3093, opened 2024-08-27
XcodeCapp: Import toolchain environment as configured by user login settings

enquora:
Initial version, for testing.
Does not provide overrides for loaded settings.
Some edge cases not yet handled, but reliability should still improve on release version.


mrcarlberg:
Is this ready or do you want to work more on this?

enquora:
More testing is required - I have just received a new M4 Mac and will be using setup of it as a testing opportunity.
In particular, we will be accommodating installation of Node (and Python, until that dependency is removed) using [`asdf`](https://github.com/asdf-vm/asdf)

enquora:
Yes, it is still valid. Do not merge anything.
Fixes for larger structural fragility in the toolchain are ongoing - this is needed until those are ready.

daboe01:
@enquora will this obsoleted by your ongoing work on the toolchain?
if so, we should close it

enquora:
It will be.

PR #3094, opened 2024-08-30
Add editorconfig file

enquora:
Enforce/guide compliance with Cappuccino project coding standards. Most modern editors supported directly or via plugins.
https://editorconfig.org


enquora:
Github applies .editorconfig files automatically - it appears the rules in this one are problematic.
This should be pursued nevertheless - to replace tabs with spaces, strip whitespace at end of lines, maintain consistent line endings, and a few other things.

PR #3100, opened 2024-10-15
Fixed: Make the isa property on classes not enumerable.

mrcarlberg:
As we have a tool-free bridges between many classes and JavaScript types by setting the isa property on the type, for example Array and CPArray. Using the for...in enumeration the isa property will be included in the loop if the property is enumerable (this is the default behavior). This commit set the property as not enumerable.


mrcarlberg:
Merged!

PR #3103, opened 2024-12-30
New: give nil values the chance to be transformed to true via CPNegateBoolean

daboe01:
fixes #1986 

@mrcarlberg what do you think?


mrcarlberg:
Excellent. Can you also merge in the test case from #1986 so we make sure it fixes the problem.

daboe01:
@mrcarlberg no i finally made it :-)

mrcarlberg:
Good work! But you have not yet answered why you are using `==` instead of `===`  when comparing the transformer name with `CPNegateBooleanTransformerName`?

mrcarlberg:
Thanks! Merged!

PR #3108, opened 2025-03-25
fixed: coalescing with empty attributes did not work

daboe01:


mrcarlberg:
You need to fix the test cases too.

PR #3111, opened 2025-06-11
new: CPImage method drawAtPoint:fromRect: operation: fraction:

daboe01:


michaelbach:
> i have not found the symbolic constants for the blendmodes in cappuccino so far.…
In CGContext.j there is a list "@group CGBlendMode"; but I'm sure you would have seen this.

daboe01:
@michaelbach thank you!

PR #3120, opened 2025-06-22
fixed: subview's window setting did not happen BEFORE _setSuperview

### LINKED ISSUE #2735: CPWindow is null in CPViewController viewDidAppear
`[[self view] window]` returns null in `CPViewController viewDidAppear`

daboe01:
fixes #2735


daboe01:
@enquora what do you think?

PR #3121, opened 2025-06-22
fixed: CPView was missing rotateByAngle: method

### LINKED ISSUE #2982: CPView is missing rotateByAngle method
Do we have an entirely distinct (from Objective-C) of accomplishing arbitrary rotation of a view or am I correct assuming it isn't currently possible without directly manipulating the underlying DOM?

The only technically feasible way of implementing this is using an underlying CSS transform, afaik. 
Does this risk a collision with Aristo3's CSS theming mechanism @didierkorthoudt?

daboe01:
fixes #2982

### DIFF
--- a/AppKit/CPView.j
+++ b/AppKit/CPView.j
@@ -3291,7 +3291,13 @@ setBoundsOrigin:
 }
 
 /*!
-    Rotates the view's frame by a given angle (in degrees) around its center point.
+    Rotates the view's visual representation by a given angle (in degrees) around its center point.
+
+    This method achieves the rotation by applying a transform directly to the view's backing CALayer.
+    Because this is a direct layer manipulation, the view's own `frame` property is not updated to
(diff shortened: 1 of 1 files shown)


daboe01:
manual test has been added to CGCanvasContext

mrcarlberg:
Nice!

One question. Shouldn't the method `rotateByAngle:` posts a `CPViewBoundsDidChangeNotification` to the default notification center if the view is configured to do so?

daboe01:
> One question. Shouldn't the method `rotateByAngle:` posts a `CPViewBoundsDidChangeNotification` to the default notification center if the view is configured to do so?

You are right in principle that a visual geometry change should trigger a notification. However, due to this method being a leaky abstraction and the significant work required for a clean implementation, the current approach is a reasonable compromise because, strictly speaking, the view's frame property does not change from the rotation of the layer.

PR #3123, opened 2025-06-23
fixed: CPViewController doesn't participate fully in responder chain

### LINKED ISSUE #2725: CPViewController doesn't participate fully in responder chain
Several items:

CPViewController allows overriding `- (BOOL)acceptsFirstResponder` and manually inserting itself into the responder chain.
* Recursively following the responder chain after doing so indicates CPViewController is part of the chain. However, using `[CPApp sendAction:to:from:]` to propagate a message elicits no response from CPViewController or subclass, even if the sent selector is present and confirmed visible using `[AViewController respondsToSelector]`
* `CPViewController` does not seem to be present in the responder chain, although this seems to be the default behaviour in modern Cocoa

Problem seen using `CPViewController` for a `CPOutlineView` subclass.

daboe01:
fixes #2725

### DIFF
--- a/AppKit/CPViewController.j
+++ b/AppKit/CPViewController.j
@@ -174,7 +174,7 @@ var CPViewControllerCachedCibs;
         [cib instantiateCibWithExternalNameTable:_cibExternalNameTable];
     }
     else
-        [self setView:[CPView new]];
+        _view = [CPView new];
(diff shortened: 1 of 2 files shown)


daboe01:
@enquora what do you think?

PR #3124, opened 2025-06-23
fixed: touch scrolling did not support momentum

daboe01:
With these changes, two-finger touch scrolling will now have a pleasant and intuitive momentum effect, significantly enhancing the feel of your application on touch-enabled devices.


daboe01:
@michaelbach confirmed that 2-finger scrolling is working as advertised on the ipad

PR #3125, opened 2025-06-24
update flatten and press for node

daboe01:
alternative to #2577


daboe01:
has some issues that i will fix asap.

daboe01:
sorry, this is impossbile. both tools are obsolete anyway.

PR #3126, opened 2025-06-25
removed: obsolete code from the old compiler

daboe01:
this is for testing at the moment, do not merge yet.


daboe01:
i can run jake install without issues. i opened a few manual tests without seeing issues. and all the tests pass.

mrcarlberg:
I can't see that this is obsolete code? If it is removed there will be no way of building the Objective-J framework for the Browser environment.

daboe01:
i will have to find out, what i am doing wrong with my setup.
maybe jake install is not the right way to build?

PR #3127, opened 2025-06-26
fixed: theming for CPTextView

daboe01:
a more modern version in comparison to #2877

### DIFF
--- a/AppKit/CPColor.j
+++ b/AppKit/CPColor.j
@@ -503,6 +503,16 @@ var cachedBlackColor,
     return [[CPColor alloc] _initWithCSSString: aString];
 }
 
++ (CPColor)textColor
+{
+    return [CPColor blackColor];
+}
+
++ (CPColor)insertionPointColor
+{
+    return [CPColor blackColor];
+}
+
 + (CPColor)selectedTextBackgroundColor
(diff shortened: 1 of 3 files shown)


AppKit/CPColor.j
```diff
@@ -501,16 +504,36 @@ var cachedBlackColor,
     return [[CPColor alloc] _initWithCSSString: aString];
 }
 
++ (CPColor)textColor
+{
+    return [[self _cachedThemeColor] valueForThemeAttribute:@"text-color"] || [CPColor blackColor];
+}
+
++ (CPColor)insertionPointColor
```
daboe01: If a user creates a Dark Mode theme with white text, their cursor will remain black and be invisible against a dark background. This method needs to look up [[self _cachedThemeColor] valueForThemeAttribute:@"insertion-point-color"] (or similar) and fallback to black. Currently, this method lies about being theme-aware.

daboe01 (commented):
Fix the logic inside setEditable: that toggles the Disabled state.
Fix insertionPointColor in CPColor.j to actually perform a theme lookup.

daboe01:
this one is a mess. we need to start from scratch

PR #3133, opened 2025-07-13
fixed: CPCollectionview displayed an erroneous offset before the first row of content

### LINKED ISSUE #2483: CPCollectionView adds an extra separator before index 0
The height of the collection view is calculated using:

`    height = MAX(height, numberOfRows * (_minItemSize.height + _verticalMargin));
`

and the layout is done using:

`    [displayItems enumerateObjectsUsingBlock:function(item, idx, stop)
    {
        var view = [item view];

        if (idx >= displayCount)
        {
            [view setFrameOrigin:CGPointMake(-anItemSize.width, -anItemSize.height)];
            return;
        }

        if (idx % numberOfColumns == 0)
        {
            x = _horizontalMargin;
            y += _verticalMargin + anItemSize.height;
        }

        [view setFrameOrigin:CGPointMake(x, y)];
        [view setFrameSize:anItemSize];

        x += anItemSize.width + _horizontalMargin;
    }];
`

... which means there's always an offset of _verticalMargin before the first item in the scrollbar. I was calculating my required height as (height of row * num rows) + (height of margin * num rows - 1) and it was always slightly off (by 5 pixels, since that's the default separation).

The 'idx % numberOfColumns' part probably needs a check for (idx != 0) for the addition of _verticalMargin and the overall height then needs to take into account the lack of an initial separator offset.

I've only looked at 1-item-per-row collection views, but if you have multiple items per row, the 'x +=' line should probably be suitably modified as well.

daboe01:
fixes #2483

### DIFF
--- a/AppKit/CPCollectionView.j
+++ b/AppKit/CPCollectionView.j
@@ -566,14 +566,14 @@ var HORIZONTAL_MARGIN = 2;
 
 - (void)_computeGridWithSize:(CGSize)aSuperviewSize count:(Function)countRef
 {
-    var width               = aSuperviewSize.width,
-        height              = aSuperviewSize.height,
-        itemSize            = CGSizeMakeCopy(_minItemSize),
-        maxItemSizeWidth    = _maxItemSize.width,
-        maxItemSizeHeight   = _maxItemSize.height,
-        itemsCount          = [_items count],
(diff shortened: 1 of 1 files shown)


daboe01:
just tested this, works like a charm.

PR #3140, opened 2025-08-29
new: CPMapTable.j

### LINKED ISSUE #2440: Implement CPMapTable
In cocoa there is a NSMapTable that looks like a NSMutableDictionary but with the ability to store arbitrary pointers as keys, instead of NSStrings only.

In cappuccino, we could use the ECMASCRIPT 6 Map object. When not available, a javascript library  could be used and ideally lazy loaded.

Some useful docs:
[NSMaptable reference](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/NSMapTable_class/)
[EcmaScript 6 availability](http://kangax.github.io/compat-table/es6/)
[jshashtable a javascript hash table](http://timdown.co.uk/jshashtable/)

daboe01:
fixes #2440

### DIFF
--- a/Tests/Foundation/CPMapTableTest.j
+++ b/Tests/Foundation/CPMapTableTest.j
@@ -1,6 +1,7 @@
 @import <Foundation/CPMapTable.j>
 @import <Foundation/CPDictionary.j>
 @import <Foundation/CPEnumerator.j>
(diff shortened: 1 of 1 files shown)


mrcarlberg:
Thanks! Merged!

PR #3142, opened 2025-09-03
new: CPSplitViewController and CPSplitViewItem

### LINKED ISSUE #2350: Implementation of CPSplitViewController
More informations here : https://developer.apple.com/library/mac/documentation/AppKit/Reference/NSSplitViewController_Class/

daboe01:
fixes #2350

### DIFF
--- a/AppKit/CPSplitViewController.j
+++ b/AppKit/CPSplitViewController.j
@@ -262,6 +262,13 @@ var CPSplitViewControllerAutomaticDimension = -1.0;
     return proposedMaximumPosition;
 }
 
+- (CGRect)splitView:(CPSplitView)splitView effectiveRect:(CGRect)proposedEffectiveRect forDrawnRect:(CGRect)drawnRect ofDividerAtIndex:(CPInteger)dividerIndex
+{
+    // Default behavior: return the proposed rectangle.
+    // This can be overridden to provide a larger or custom hit area for the divider.
+    return proposedEffectiveRect;
+}
+
 - (void)splitViewDidResizeSubviews:(CPNotification)notification
 {
     // Can be used to respond to user-initiated resizing.


enquora:
Happy to see this, it's been on my TODO list for awhile.
Several comments:

1. Needs creation date
2. Needs copyright attribution, use from any existing class implementation
3. LLMs are good, in my experience, at synthesizing inline documentation. Should be included here, imo. Prompt needs to ask for neutral technical tone in passive third party voice.
4. LLMs are also good at ordering method definitions. Here, the class methods are at the end of the implementation. Absent any implications for our documentation generator, class methods should precede instance methods. This follows Apple's own documentation standards, and makes understanding the functionality easier. Most of our implementations do not follow these ordered, and I have always found this problematic. Our documentation generator or established precedents may impose constraints on this, though.
5. I strongly believe that if any reference is made to an LLM which generate all or parts of the code, it belongs in class documentation rather than the preamble. The person directing the LLM should be referenced as per normal copyright/author notice.
6. Needs and is a good candidate for both automated and manual test cases. While LLMs have proven very helpful in code cleanup, and in some Greenfield code generation, they are all too often confidently wrong - in my experience. Manual tests especially provide a valuable guide for people new to Cappuccino or particular parts of the codebase.

enquora:
Note that #pragma mark used in this is for Objective-C.
For Objective-J and Javascript it is:
// MARK: – Creates a named section in the jump bar.
// MARK: - – Creates a section divider with a horizontal rule.
// TODO: and // FIXME: – Appear as tasks in the jump bar’s task list (and in Issue Navigator if enabled).

Apple calls these landmarks. They are only supported by Xcode, but I favour their inclusion.
Xcode 26's Code Intelligence feature under Tahoe is effective enough to justify 'landmarks' for use in that editor.

PR #3159, opened 2025-12-07
fixed: Context Menus, do not update selections while mouse is down

### LINKED ISSUE #1993: Context Menus, do not update selections while mouse is down.
To reproduce:
In a view that supports context menus ie:tableview...
Activate a context menu via right click and keep the mouse down.
The menu appears but hovering the mouse over any of the menu items shows no highlighting. Releasing the mouse over a menu item does trigger the menu action.

I could not find a context menu test, so using  Tests/Manual/TableTest/BorderTableTest.
In its AppController add this to awakeFromCib:

```
    [theTableView setDelegate:self];
```

And add the following 2 methods.

```
- (CPMenu)tableView:(CPTableView)aTableView menuForTableColumn:(CPTableColumn)aTableColumn row:(int)aRow
{
    var menu = [[CPMenu alloc] initWithTitle:@"Right Click Menu"],
        menuItem = [[CPMenuItem alloc] initWithTitle:@"Item 1" action:@selector(menuAction:) keyEquivalent:@""];

    [menuItem setTarget:self];
    [menu addItem:menuItem];

        menuItem = [[CPMenuItem alloc] initWithTitle:@"Item 2" action:@selector(menuAction:) keyEquivalent:@""];
    [menuItem setTarget:self];
    [menu addItem:menuItem];

    menuItem = [[CPMenuItem alloc] initWithTitle:@"Item 3" action:@selector(menuAction:) keyEquivalent:@""];
    [menuItem setTarget:self];
    [menu addItem:menuItem];

    return menu;
}

- (void)menuAction:(id)sender
{
    alert([sender title]);
}
```

Run the test and right click on one of the rows in the tableview and keep the mouse down while hovering over the menu items.

daboe01:
When a context menu is activated via a mouse down event (e.g. right-click and hold), items were not highlighting until the user moved the mouse. This occurred because the menu tracking loop waits for the next event to update the selection.

This commit adds a manual hit-test immediately after the menu window is ordered to the front. If the mouse is already over an item when the menu opens, that item is now highlighted instantly.

Fixes #1993

### DIFF
--- a/Tests/Manual/TableTest/BorderTableTest/AppController.j
+++ b/Tests/Manual/TableTest/BorderTableTest/AppController.j
@@ -35,6 +35,33 @@ CPLogRegister(CPLogConsole);
     [theWindow setBackgroundColor:[CPColor colorWithHexString:@"f3f4f5"]];
 
     [theBorderTypePopup selectItemWithTag:[theScrollView borderType]];
+
+    // for testing context menus
+    [theTableView setDelegate:self];
+}
+
+- (CPMenu)tableView:(CPTableView)aTableView menuForTableColumn:(CPTableColumn)aTableColumn row:(int)aRow
+{
+    var menu = [[CPMenu alloc] initWithTitle:@"Right Click Menu"],
+        menuItem = [[CPMenuItem alloc] initWithTitle:@"Item 1" action:@selector(menuAction:) keyEquivalent:@""];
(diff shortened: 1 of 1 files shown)


michaelbach:
Excellent! Indeed that was sort of subliminally nagging me.

PR #3183, opened 2026-02-05
Fix event handling cp font panel and cp color well

enquora:
Implement event handling using responder chain


daboe01:
@enquora maybe we can close this one and revisit the colorpanel for a general overhaul when the more urgent issues are fixed?

enquora:
Agreed. Merge whatever is required and not yet done from either your or my own fix, although I believe this has already been done.
Any layout issues remaining are out of scope for this.

I'll leave it to yourself to close the issue.

PR #3236, opened 2026-06-17
new: Add XIB serialization support for CPSplitViewController and CPStackView

### LINKED ISSUE #3231: XIB serialization code is in ./Tools/nib2xib, separate from class definitions and making drift between the two all but inevitable
Neither `CPSplitViewController` nor `CPStackView` have a corresponding NS-class handler in ./Tools/nib2cib, despite both having been part of the framework for some time. Any XIB that references either class will fail silently at runtime — the unarchiver receives no mapping and produces no error.

## This is symptomatic of a structural problem

The nib2cib toolchain maintains a per-class handler for every AppKit class it can serialize.
These handlers are developed and maintained independently of the framework classes they correspond to.
There is no mechanism to enforce or verify correspondence between the two.

This means:

- Any class added to AppKit or Foundation without a matching nib2cib handler is silently unserializable from XIBs.
- Any property added, renamed, or corrected in a framework class may silently diverge from its nib2cib handler with no indication of failure at compile time or at runtime.
- The actual correspondence between existing handlers and their framework counterparts, across the full set of classes, is not verifiable without manual audit of every handler against its framework class.

Given the volume of gaps that have been filled and corrected in AppKit and Foundation over the years, drift in existing handlers is likely and its extent is unknown. Failures are silent.

## Solution
The root cause is an encapsulation violation: the compiler holds semantic knowledge that belongs exclusively to the framework classes themselves. This is addressed in the nearly complete toolchain improvements.

The updated toolchain adopts a class-agnostic structural compiler.
The new `nib2cib` transliterates XIB XML into Cappuccino's native property list format without any per-class knowledge.
Serialization responsibility moves entirely into the framework, where each AppKit and Foundation class carries its own XIB decoding implementation alongside its other interface responsibilities.
A class present in the framework is therefore serializable from XIBs by definition, and a change to its properties is reflected in its decoding implementation as a matter of ordinary development practice.

Correspondence between a class's declared properties and its decoding implementation can be verified continuously by a static analysis pass over the framework source, using tree-sitter-objj, run as a mandatory CI check.
This makes the class of silent drift described above statically detectable and permanently prevented.

A one-time cost of moving the existing NS serialization code into associated class definitions must be paid.
Correctness must then be re-established for all classes.

This is a mostly mechanical task which can be automated by traditional means or using an inference engine.
The code to handle legacy cib structures remains intact.
A branch is created on determination of the legacy 280NPLIST or new CAPPUCCINOPLIST declarations.
Retaining support for both the legacy and new formats does incur the penalty of carrying runtime code for both decoders.
The legacy coders are generally very simple, often just an NS => CP class prefix swap.
The new format coders will be somewhat more complex, carrying a larger lines of code penalty.

I will provide extended background documentation on this, if not made sufficiently clear here.
All other options have been considered and deemed impractical or architecturally problematic.
This problem represents technical debt present since the first version.

daboe01:
Implement NSCoding support for CPSplitViewController, 
CPSplitViewItem, and CPStackView. This resolves runtime deserialization 
failures when unarchiving XIB files that utilize these components.

- Create `NSSplitViewController.j` to decode standard `NSSplitViewController` 
  and `NSSplitViewItem` keys, mapping them back to their native CP equivalents.
- Create `NSStackView.j` to decode standard `NSStackView` attributes (such 
  as orientation, alignment, spacing, edge insets, and arranged subviews).
- Separate the decoding logic and the NS-to-CP class mapping from the core 
  framework implementation files, keeping them clean and self-contained.

This aligns with the framework-integrated serialization model of the 
updated nib2cib toolchain.

fixes #3231

### DIFF
--- a/Tools/nib2cib/NSSplitViewController.j
+++ b/Tools/nib2cib/NSSplitViewController.j
@@ -1,0 +1,104 @@
+/*
+ * NSSplitViewController.j
+ * nib2cib
+ *
+ * Created by Daniel Boehringer.
+ * Copyright 2025 The Cappuccino Project.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+@import <AppKit/CPSplitViewController.j>
+
+
+@implementation CPSplitViewController (NSCoding)
+
+- (id)NS_initWithCoder:(CPCoder)aCoder
+{
+    if (self = [super NS_initWithCoder:aCoder])
+    {
+        _splitViewItems = [CPMutableArray array];
+        _minimumThicknessForInlineSidebars = 20.0;
+
+        if ([aCoder containsValueForKey:@"NSSplitView"])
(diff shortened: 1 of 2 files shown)


enquora:
Note that I have not included every missing serializer in the issue, but every control which is usable in IB requires one.
Rulers are an example.

We'll leave the issue alive and make it more generalized.
New tooling will allow auditing for drift.
Moving the location of serialization code into the class definition itself will make oversight less likely.

Issue #2980, opened 2021-04-16
CPTextContainer doesn't support tracking height changes in parent textView


enquora:
While support for widthTracksTextView is present, heightTracksTextView support is missing.

cappbot:
**Milestone:** Someday.  **Label:** #new.  **What's next?** A reviewer should examine this issue.

daboe01:
-#new
+TextView
+#needs-patch

cappbot:
**Milestone:** Someday.  **Labels:** #needs-patch, TextView.  **What's next?** This issue needs a volunteer to write and submit code to address it.

PR #3239, opened 2026-06-19
new: heightTracksTextView support in CPTextContainer + rtf fixes

daboe01:
Introduced a unified _updateFrameObserver helper method to manage registration for CPViewFrameDidChangeNotification. This ensures that the text container listens to the text view's frame changes if either widthTracksTextView or heightTracksTextView is enabled, and correctly removes the observer only when both are disabled.

Updated textViewFrameChanged: to dynamically adjust either width, height, or both depending on the active tracking properties.

Adjusted setTextView: to safely unregister from the old text view and register with the new one without overriding the user-configured tracking properties.


daboe01:
fixes #2980

PR #3240, opened 2026-06-20
Fixed: baseline alignment in _CPLineFragment to prevent overlapping text descenders

daboe01:
### Description
This pull request addresses an issue in the text layout system where descenders from larger text runs (such as "p", "g", or "y") overlap with characters in the line directly below.

#### Cause
In `CPLayoutManager.j`, the `_CPLineFragment` method `setAdvancements:` calculates glyph vertical offsets relative to the line's total height (`_usedRect.size.height`) rather than the line's calculated baseline (`_location.y`). Because of this, text runs are positioned too far down within the line fragment, causing descenders to bleed past the fragment's bottom boundary.

#### Solution
The vertical offset calculation in `setAdvancements:` has been updated to use `_location.y` instead of `height`:

```javascript
_glyphsOffsets[i] = _location.y - someAdvancements[i].height;
```

This change aligns all text runs on a line to the common baseline (`_lineBase`). It ensures that the space below the baseline is allocated for descenders within the line fragment, keeping them within their proper boundary and preventing overlapping.

#### Testing
* Tested with mixed font sizes on adjacent lines.
* Verified that baseline alignment remains consistent across different font heights and descenders no longer overlap with the line below.

### DIFF
--- a/AppKit/CPTextView/CPLayoutManager.j
+++ b/AppKit/CPTextView/CPLayoutManager.j
@@ -965,16 +965,11 @@ _oncontextmenuhandler = function () { return false; };
                         if (frame)
                         {
                             var correctedRect = CGRectCreateCopy(frame);
-                            correctedRect.size.height -= frame._descent;
-                            correctedRect.origin.y -= frame._descent;
 
                             if (!rect)
                                 rect = CGRectCreateCopy(correctedRect);
                             else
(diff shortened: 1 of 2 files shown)


daboe01:
before:
Bildschirmfoto 2026-06-20 um 11 33 54

daboe01:
after:
Bildschirmfoto 2026-06-20 um 12 10 23

PR #3256, opened 2026-07-21
Modernize CGContextCanvas.j: Remove CPP macros and legacy fallbacks

enquora:
As part of the ongoing effort to remove the C-preprocessor (CPP) from the new Lisette compiler pipeline, this commit refactors CGContextCanvas.j to use native JavaScript and HTML5 Canvas APIs directly.

It is valuable on its own:
* unnecessary CPP directives made code flow less obvious and debugging more difficult than otherwise necessary
* fallbacks and polyfills that no browser has executed for at least a decade obscure intent and reduce performance

Key changes:
- Removed all `#define` macros that wrapped standard Canvas API calls (e.g., `_CGContextMoveToPointCanvas`). Direct method invocations are now used, eliminating the preprocessor pass and simplifying AST generation.
- Removed the `CPJavaScriptCanvasTransformFeature` runtime check and its complex mathematical fallbacks (`scale_rotate`, `rotate_scale`, `eigen`). Modern target environments universally support native canvas transformations.
- Removed obsolete vendor-prefixed line dash properties (`webkitLineDash`, `mozDash`) in favor of the standard `setLineDash`.
- Replaced `var` with `const`/`let` for proper block scoping and immutability.
- Replaced the `ROUND` macro with standard `Math.round`.

These changes strip away 2008-era browser compatibility hacks and micro-optimizations, resulting in cleaner, more maintainable, and strictly browser-targeted code.

### DIFF
--- a/AppKit/CoreGraphics/CGContextCanvas.j
+++ b/AppKit/CoreGraphics/CGContextCanvas.j
@@ -3,6 +3,7 @@
  * AppKit
  *
  * Created by Francisco Tolmasky.
+ * Modified by David Richardson.
  * Copyright 2008, 280 North, Inc.
  *
  * This library is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


enquora:
@daboe01 This patch demonstrates the pattern of directive usage completely: legacy browser fallbacks not relevant for at least the last decade, and functions masquerading as macros.

Some of the further cases affect AppKit in ways which will require my thinking about things a bit more, where your own recent focus on theming and AppKit would provide context more quickly.

These are properly viewed as cleanup of technical debt rather than accommodations for a new toolchain.

AppKit/CoreGraphics/CGContextCanvas.j
```diff
@@ -20,616 +21,518 @@
...
-    context.DOMElement = DOMElement;
+	context.DOMElement = DOMElement;
 
-    // canvas gives us no way to query whether the path is empty or not, so we have to track it ourselves
-    context.hasPath = NO;
+	// canvas gives us no way to query whether the path is empty or not, so we have to track it ourselves
+	context.hasPath = NO;
 
-    return context;
+	return context;
```
daboe01, line 537: formatting: we use 4 spaces instead of tab in cappuccino

daboe01:
whitespace formatting so we can see the actual changes only

daboe01:
excellent work otherwise :-)

PR #3258, opened 2026-07-23
Refactor CPTimer: Remove CPP dependencies and mark legacy anti-patterns

enquora:
As part of the ongoing technical debt cleanup, this commit modernizes CPTimer.j by removing its reliance on the C-preprocessor and updating syntax to modern JavaScript standards.

Key changes:
- Replaced `#define CPTimerDefaultTimeInterval` with a native `const`.
- Replaced `var` with `const`/`let` for proper block scoping and immutability.
- Replaced the build-time `PLATFORM(DOM)` guard with a runtime `typeof(window) !== 'undefined'` check, ensuring Foundation does not depend on build-time platform flags.
- Added `FIXME` comments to explicitly document architectural anti-patterns:
  - The invasive global override of `window.setTimeout` and `window.setInterval`.
  - The use of `new Function` for string evaluation (CSP violation).
  - The questionable exposure of `CPTimerDefaultTimeInterval` as a fallback.

### DIFF
--- a/Foundation/CPTimer.j
+++ b/Foundation/CPTimer.j
@@ -25,285 +25,293 @@
 @import "CPObject.j"
 @import "CPRunLoop.j"
 
-#define CPTimerDefaultTimeInterval 0.1
+// FIXME: Expose CPTimerDefaultTimeInterval via public API or eliminate the fallback behaviour.
+const CPTimerDefaultTimeInterval = 0.1;
 
 /*!
-    @class CPTimer
-    @ingroup foundation
+ @class CPTimer
+ @ingroup foundation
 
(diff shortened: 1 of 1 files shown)


Foundation/CPTimer.j
```diff
@@ -25,285 +25,293 @@
...
-
-    BOOL                _repeats;
-    BOOL                _isValid;
-    CPDate              _fireDate;
-    id                  _userInfo;
+	CPTimeInterval      _timeInterval;
+	CPInvocation        _invocation;
+	Function            _callback;
+
+	BOOL                _repeats;
```
daboe01: whitespace formatting

PR #3259, opened 2026-07-23
Refactor CPNumber: Remove CPP macros and modernize syntax

enquora:
Key changes:
- Removed the `CAST_TO_INT` macro and replaced it with native ES6 `Math.trunc()` for float truncation.
- Replaced `var` with `let` for proper block scoping.
- Simplified boolean coercion in `boolValue` using the `!!` operator.
- Replaced the hostile `throw new Error` in `descriptionWithLocale:` with a safe fallback to `self.toString()`.
- Added `FIXME` comments to explicitly document architectural anti-patterns, including the global mutable state in `CPNumberUIDs` and the unimplemented `decimalValue` method.

### DIFF
--- a/Foundation/CPNumber.j
+++ b/Foundation/CPNumber.j
@@ -39,162 +39,162 @@ const CPNumberUIDs = new CFMutableDictionary();
 
 + (id)alloc
 {
-	// MODIFICATION: Replaced 'var' with 'let' for block scoping.
-	let result = new Number();
-	result.isa = [self class];
-	return result;
+    // MODIFICATION: Replaced 'var' with 'let' for block scoping.
+    let result = new Number();
+    result.isa = [self class];
+    return result;
 }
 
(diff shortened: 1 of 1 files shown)


Foundation/CPNumber.j
```diff
@@ -25,309 +25,290 @@
...
+	// MODIFICATION: Replaced 'var' with 'let' for block scoping.
+	let result = new Number();
+	result.isa = [self class];
+	return result;
 }
 
 + (id)numberWithBool:(BOOL)aBoolean
 {
-    return aBoolean ? 1 : 0;
+	return aBoolean ? 1 : 0;
```
daboe01: whitespace formatting

enquora: > whitespace formatting

.editorconfig does not seem to be working with Xcode 27.

PR #3270, opened 2026-08-11
Fix legacy compiler warnings in CPMapTable

enquora:
Using ES6 destructuring in the `for...of` loop declaration (`var [key, value] of _map.entries()`) causes the legacy Objective-J compiler to emit "uninitialized global variable" warnings for `key` and `value`.

This raises concerns about variable scoping and generates unacceptable noise in the CI pipeline. To resolve the warnings, the loop has been refactored to destructure inside the loop body.

A TODO has been added to revert to ES6 destructuring once the legacy compiler is retired.

### DIFF
--- a/Foundation/CPMapTable.j
+++ b/Foundation/CPMapTable.j
@@ -105,17 +105,31 @@
 // MARK: Creating a Dictionary Representation
 
 /*!
-   Returns a dictionary representation of the map table.
-   Note: This will only work correctly if all keys are strings.
+ Returns a dictionary representation of the map table.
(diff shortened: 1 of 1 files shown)


enquora:
@daboe01 This keeps a history of the problem, which is likely to resurface elsewhere - making it preferable to `_map.forEach`.

Do you concur?

daboe01:
> @daboe01 This keeps a history of the problem, which is likely to resurface elsewhere - making it preferable to `_map.forEach`.
> 
> Do you concur?

yes, thank you!

PR #3272, opened 2026-08-12
replace unload event listener with pagehide

daboe01:
- **Replace `unload` event listener with `pagehide`**: Modern browsers (Chrome 115+) deprecate `unload` and enforce a Permissions Policy that logs `[Violation] Permissions policy violation: unload is not allowed in this document`. Updating to `pagehide` satisfies the policy while maintaining proper window cleanup and bfcache compatibility.


enquora:
+1, looks good. This improves performance, beyond simply accommodating modern permissions.
Browser support significantly predates the last ES6 adoptions in 2017 (Firefox and Microsoft Edge) - which seems to have become our defacto minimum.

PR #3281, opened 2026-08-19
Improve and extend CPTreeNode tests

enquora:


enquora:
It is unclear just yet why this fails, while it runs cleanly locally. 
The expanded test coverage is important but not on the critical path.
This problem will be addressed later.

daboe01:
The test failures are caused by two discrepancies in the new test suite in `CPTreeNodeTest.j`:

---

### 1. `testIndexPaths` Failure

#### Error:
```
addFailure test=[CPTreeNodeTest testIndexPaths]: expected:<null> but was:<<CPIndexPath 0x027b52> >
```

#### Cause:
In `CPTreeNode.j`, calling `-indexPath` on a root node (a node without a `_parentNode`) returns an empty index path (`[CPIndexPath indexPathWithIndexes:[]]`), matching Apple AppKit's `NSTreeNode` behavior:
```objective-c
- (CPIndexPath)indexPath
{
    if (!_parentNode)
        return [CPIndexPath indexPathWithIndexes:[]];
    ...
```
However, the test on line 134 asserted that `[root indexPath]` returns `nil`:
```objective-c
[self assert:nil equals:[root indexPath]];
```

#### Fix:
Assert that `[root indexPath]` is an empty index path (or has length 0):
```objective-c
- (void)testIndexPaths
{
    [self assert:[CPIndexPath indexPathWithIndexes:[]] equals:[root indexPath]];

    [root insertObject:child1 inChildNodesAtIndex:0];
    [self assert:[CPIndexPath indexPathWithIndex:0] equals:[child1 indexPath]];

    [child1 insertObject:child2 inChildNodesAtIndex:0];
    [self assert:[CPIndexPath indexPathWithIndexes:[0, 0]] equals:[child2 indexPath]];
}
```

---

### 2. `testMovingExistingChild` Failure

#### Error:
```
addFailure test=[CPTreeNodeTest testMovingExistingChild]: expected:<<CPTreeNode 0x027b43>> but was:<<CPTreeNode 0x027b41>>
```

#### Cause:
Let's trace how `insertObject:inChildNodesAtIndex:` adjusts the index when reparenting/moving within the same parent:
1. Initially, `_childNodes` is `[child1, child2, c3]` (length 3).
2. Calling `[root insertObject:child1 inChildNodesAtIndex:2]`:
   - `child1` is at `originalIndex = 0`.
   - `child1` is removed, leaving `[child2, c3]` (length 2).
   - Because `originalIndex < anIndex` (`0 < 2`), `CPTreeNode` adjusts `anIndex` down by 1 (`anIndex` becomes `1`) to maintain index semantics relative to the original array (inserting before `c3`).
   - `child1` is inserted at index 1 of `[child2, c3]`, resulting in:
     - Index 0: `child2`
     - Index 1: `child1`
     - Index 2: `c3`
3. The test expected `c3` at index 1 and `child1` at index 2.

#### Fix:

Depending on whether you want to test inserting before `c3` (index 2 in original array) or moving to the end of the array (index 3 in original array):

**Option A (Expected order after inserting at index 2 is `[child2, child1, c3]`):**
```objective-c
- (void)testMovingExistingChild
{
    var c3 = [CPTreeNode treeNodeWithRepresentedObject:@"child3"];

    [root insertObject:child1 inChildNodesAtIndex:0];
    [root insertObject:child2 inChildNodesAtIndex:1];
    [root insertObject:c3 inChildNodesAtIndex:2];

    // Array is [child1, child2, c3]. Move child1 (index 0) to index 2 (before c3).
    [root insertObject:child1 inChildNodesAtIndex:2];

    [self assert:child2 equals:[root objectInChildNodesAtIndex:0]];
    [self assert:child1 equals:[root objectInChildNodesAtIndex:1]];
    [self assert:c3 equals:[root objectInChildNodesAtIndex:2]];

    // Move child1 (index 1) back to index 0.
    [root insertObject:child1 inChildNodesAtIndex:0];

    [self assert:child1 equals:[root objectInChildNodesAtIndex:0]];
    [self assert:child2 equals:[root objectInChildNodesAtIndex:1]];
    [self assert:c3 equals:[root objectInChildNodesAtIndex:2]];
}
```

**Option B (Move to the end of the array using index 3):**
```objective-c
    // Move child1 (index 0) to the end of the list (index 3).
    [root insertObject:child1 inChildNodesAtIndex:3];

    [self assert:child2 equals:[root objectInChildNodesAtIndex:0]];
    [self assert:c3 equals:[root objectInChildNodesAtIndex:1]];
    [self assert:child1 equals:[root objectInChildNodesAtIndex:2]];
```

enquora:
This addresses test logic but the root question is missed - the environmental discrepancy between local execution and the CI pipeline. While this test unit fails the GitHub CI, others which are also failing locally do not. And none cause a local run of the tests to halt or throw an error.

This is the root problem which was raised. The specifics with the PR become obvious to anyone looking closely at the code. What is more important is that the root issue be addressed in the large, especially given the multi-year aristo3 branch which must be merged. 

While the examples serve as useful supporting data when provided as attachments, the verbose, didactic, and stilted presentation used obscures the underlying technical argument and falls below the standard for discourse which community engagement requires, especially those coming across this later. A concise summary is necessary to make the point.

daboe01:
On the positive side, having two independent test environments actually provides an extra layer of safety by catching issues from different angles, as this case showed. So we don’t need to worry about stability for the aristo3 merge: all manual tests have been run, my own apps are fully tested, and we already have two independent confirmations backing this up.

enquora:
> So we don’t need to worry about stability for the aristo3 merge

The tolerant and permissive behaviour of the legacy Node-hosted toolchain is precisely why we must be vigilant regarding the aristo3 branch. Broad, foundational changes require strict verification, not relaxed standards.

In the absence of a comprehensive automated integration suite, our unit tests are the sole reference for correctness. They cannot be treated as optional or approximate; they must be completely clean and free of anomalies. An unresolved anomaly in a unit test is rarely isolated - _it almost always signals broader, unexamined side effects elsewhere in the framework_. This is why the fundamental issue is not the specific failure in this unit, but rather why other similar unit failures are not treated as halts by our CI.

While capp-build is explicitly designed to catch the lax patterns tolerated by Node, its efficacy is constrained by two factors:
1. Dependence on a clean reference: A strict toolchain requires a dependable baseline to validate its own parsing logic. A permissive legacy environment that tolerates or emits anomalous behaviour cannot serve as that reference.
2. Sequential error masking: By design, capp-build operates as a multi-pass compiler that halts at the first phase containing errors. A failure in an early pass obscures distinct classes of errors in downstream passes.

Methodically eliminating two decades of accumulated debt requires addressing these masking effects phase by phase. We cannot establish baseline stability - or safely retire continuing reliance on the legacy toolchain - until the unit tests pass without exception or anomaly. 

The road to successful transition to a new and robust toolchain passes directly through the pristine operation of the existing one.

enquora:
Tests now pass, but not ready to merge.
* Failures were present in both the test and class itself, now fixed in each.
* Further coverage wanted for KVO functionality.

Hierarchical data is fundamental to many non-trivial applications and CPTreeNode is the AppKit primitive used to manage it.

The provided functionality, starting with tree navigation but extending to KVO, is amongst the most complex AppKit supports.

While regression-gating of tree traversal is now present, KVO functionality testing is still weak or absent. 

Given the importance of CPTreeNode, working on extending the tests further.
Decision to come on whether to fix any problems extended KVO testing surfaces or mark and defer.

